PR #27, opened 2023-03-21
Close panel on when user clicks on a nav link

### LINKED ISSUE #26: [DOC-187] User menu is closing after a link is select from the nav but should
<sub>[DOC-187](https://linear.app/documenso/issue/DOC-187/user-menu-is-closing-after-a-link-is-select-from-the-nav-but-should)</sub>

dephraiim:
Fixes #26


ElTimuro (changes requested):
Great, but does not work when I click on Profile..

ElTimuro (approved):

PR #34, opened 2023-04-04
Add `prettier`

dephraiim:


ElTimuro (approved):
Looks good. I removed one last trivial merge conflict. Merging. To be sure I created a tag if we find regression, but I dont' expect that.

PR #35, opened 2023-04-04
Assorted updates from code review

Mythie:
We had a session on 05-04-2023 where we reviewed the Documenso codebase, during that review we covered several items including:

- [x]  Update documentation to reflect seeding and migrating the database
- [x]  Include a seed for a document with a signable pdf?
- [x]  Add instructions for SMTP (inbucket)
- [x]  Signed users are duplicated (bug)
- [x]  Changing filter does not update url state
    - [ ]  Tidy up `document.tsx`
- [x]  Prop types are needed in more places
    - [ ]  More prop types are needed in more places
- [x]  Some useEffects can be swapped for event handlers
    - [ ]  There is still the ones in `documents.tsx` but they are load bearing and should be approached more carefully.
- [x]  api toasts to be moved to caller [Discussed reasoning and will not move for now]
    - sendSigningRequests
    - deleteRecipient
    - signDocument
- [x]  Double check getUserFromToken, specifically the next-auth getToken
- [ ]  Check if we still need a local copy of `node-signpdf` in `addDigitalSignature`
    - [ ]  The PR was merged only last month, currently have the latest version installed already and just need to work on swapping it out.
- [x]  More useEffect shenanigans

This PR contains those changes broken up into smaller commits for individual review.


ElTimuro:
Pretty awesome, looking forward to review

ElTimuro:
@Mythie I just merged the new prettier config, including a format of all files. This causes conflict with your branch, since your files are not formatted with the new config yet. Let me know if you need help merging.

PR #38, opened 2023-04-06
fix: improve text insertion accuracy

Mythie:
Previously, inserted text would appear a little off center from where the user had selected which could cause some frustration.

We improve upon this by updating the code responsible for centering the text to behave in a more accurate manner. From what I can tell it looks to be quite solid but could do with more rigorous testing with shorter and longer inputs.

You can see the improved accuracy in action here:

https://www.loom.com/share/1095fee7605c4790b8b30f573a04f0f0


ElTimuro (approved):
Looks good, merging to test for further manual testing by

Issue #41, opened 2023-04-06
[DOC-208] Deleting Documents does not work after Signing


ElTimuro:
It is just how it happend to be built initially. PDF with signatures cannot be deleted (database wise) without deleting those from the DB first to make sure nothing gets every lost. Options would be A to implement a delete taking this into account or B setting a delete Flag. opinions welcome.

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-208](https://linear.app/documenso/issue/DOC-208/deleting-documents-does-not-work-after-signing)</sub>

PR #45, opened 2023-04-07
Fix-#41: Change referential action for Signature_recipientId_fkey

SauravL3010:
An easy way to allow for PDF document deletion after being signed is changing the Signature_recipientId_fkey's onDelete referential action to Cascade, such that signatures of the recipients are deleted along with the PDF document. 

Draft for issue #41

### DIFF
--- a/packages/prisma/schema.prisma
+++ b/packages/prisma/schema.prisma
@@ -130,6 +130,6 @@ model Signature {
   signatureImageAsBase64 String?
   typedSignature         String?
 
-  Recipient Recipient @relation(fields: [recipientId], references: [id], onDelete: Restrict)
+  Recipient Recipient @relation(fields: [recipientId], references: [id], onDelete: Cascade)
   Field     Field     @relation(fields: [fieldId], references: [id], onDelete: Restrict)
 }


ElTimuro:
- Wouldn't the field restriction also block it?

- Technically this would work, the question is if documenso allows deletion at any point signed or not signed. I thought about this a bit and I think blocking Users from deleting their data or just setting a deleted flag and hiding is not right. If I delete st. on a service, I really want it gone. Maybe we should introduce a high-security confirm dialog (with a field to set the document name or "I really want to permanently delete this") if a signed document is about to be deleted.

SauravL3010:
1. So the schema is defined in the following way:
```
model Document {
    ...
}
model Field {
    ...
  Document    Document   @relation(fields: [documentId], references: [id], onDelete: Cascade)
    ...
}
```
- so when a `Document` record is deleted, all corresponding `Fields` are also deleted. 
```
model Signature {
    ...
  Recipient Recipient @relation(fields: [recipientId], references: [id], onDelete: Cascade)
  Field     Field     @relation(fields: [fieldId], references: [id], onDelete: Restrict)
}
```
- `Document` -> `Recipient` have a one-many relation, so in order to delete, a document record, Signature_recipientId_fkey's OnDelete must be `Cascade`. 
- `Field` onDelete restriction prevents users from somehow deleting a field that already has a signature on it, which makes sense. But has nothing to do with the  Document itself, because the model Field, takes care of that. 


2. Yes, I think flagging and hiding is not ideal. But something similar would be having a temporary `Deleted Bin` where documents can live for certain time. The confirm dialog may mention that as well.

ElTimuro (approved):
Looks good

PR #63, opened 2023-04-09
Disable the edit and add signer button for completed documents

### LINKED ISSUE #37: [DOC-205] Completed Documents should not be editable (for now)
* Documents with status completed should disable the edit and add recipient button
* Adding another recipient with another signature can be added later, to reduce complexity for now

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-205](https://linear.app/documenso/issue/DOC-205/completed-documents-should-not-be-editable-for-now)</sub>

dephraiim:
Fixes #37

### DIFF
--- a/apps/web/pages/documents/[id]/recipients.tsx
+++ b/apps/web/pages/documents/[id]/recipients.tsx
@@ -88,30 +88,36 @@ const RecipientsPage: NextPageWithLayout = (props: any) => {
               href={"/api/documents/" + props.document.id}>
               Download
             </Button>
-            <Button
-              icon={PencilSquareIcon}
-              disabled={props.document.status === DocumentStatus.COMPLETED}
-              color={props.document.status === DocumentStatus.COMPLETED ? "primary" : "secondary"}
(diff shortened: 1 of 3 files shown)


apps/web/pages/documents/[id]/recipients.tsx
```diff
@@ -216,7 +227,7 @@ const RecipientsPage: NextPageWithLayout = (props: any) => {
                                 Not Sent
                               </span>
                             ) : (
-                              ""
+                              <></>
```
Comment (Mythie) [inline apps/web/pages/documents/[id]/recipients.tsx:230]: This can just be `null`

Comment (Mythie) [reply apps/web/pages/documents/[id]/recipients.tsx:230]: Same for those below

Mythie (approved):

ElTimuro (changes requested):
Looks good, just 1 thing: Non availible options should be disabled, not vanished, so the user can learn when sth. is available and when it isn't without being "surprised" something is gone.

ElTimuro (approved):
Looks good

Issue #69, opened 2023-04-11
bug : Can add signature space for empty recipients


SauravL3010:
Preview:
<img width="690" alt="image" src="https://user-images.githubusercontent.com/65268462/231239420-0ade038d-40c1-4a49-bbd3-3c3b45cdc936.png">

Steps to reproduce the bug:
 - add a empty recipient
 - got to edit documents
 - see the list of recipients

SauravL3010:
p.s this doesn't effect any functionality, don't know if this was intended, just wanted to point it out. can close if not required.

ElTimuro:
No, not by design. The "draft recipient" wasn't fully thought through, as elaborated here: https://github.com/documenso/documenso/pull/70

Issue #67, opened 2023-04-11
[DOC-224] Bad Gateway Page and just a tiny quick error message (when sending to two recipients isnt possible) → not possibe to get back to the account via back button automatically, you need to use the back button and refresh the page


ElTimuro:
<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-224](https://linear.app/documenso/issue/DOC-224/bad-gateway-page-and-just-a-tiny-quick-error-message-when-sending-to)</sub>

ElTimuro:
-Here is a loom of the error: https://www.loom.com/share/b66ffe679916408e87e27b27a620bd9a
- The invalid recipients should be ignored in backend
- Exceptions need to be caught at sending level and endpoint level, no endpoint should escalate to complete 502
- This escalates to the top level and is only healed by render restarting automatically
- This works suprisignly well, but is not sound
![image](https://user-images.githubusercontent.com/1309312/233611374-780601bd-c4ba-42b6-8965-ea422e4510ab.png)
![image](https://user-images.githubusercontent.com/1309312/233611525-ea2fc44e-ab1f-47b4-883e-b47055a0e186.png)
![image](https://user-images.githubusercontent.com/1309312/233611637-a7a5e73a-5d53-4129-b1fa-50803a60b78a.png)

@Mythie

ElTimuro:
- @dephraiim True for the concrete error, but such an error maybe never bubble this high
- There is some architecture level Error Handling missing here

Mythie:
Sounds great, first thing to do is handle exceptions in the method that sends the signing request to all members. Try to handle empty values or non email-like values gracefully by skipping them.

Use a simple regex like `.+@.+` for checking email validity since the standard for emails is quite funky.

If we encounter an error with sending it'd be nice to get either a 400 or 500 response with details that could be displayed within the toast so the user can action it.

e.g.

"We were unable to send the signing request to <email>, please ensure their email is correct and try again"

Musacoli:
Hi @Mythie and @dephraiim ,
Is any one of you actively working on this? If not I'd like to take it up as my first issue.

PR #70, opened 2023-04-11
Delete empty recipient fields

### LINKED ISSUE #39: [DOC-206] When adding more than 1 Signer and the Document is sent while some Recipient email fiels stay empty, the document status be stuck in "waiting for others"
<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-206](https://linear.app/documenso/issue/DOC-206/when-adding-more-than-1-signer-and-the-document-is-sent-while-some)</sub>

SauravL3010:
||comments|
|--|--|
|`issue`| - Application allows for sending signing requests to empty recipients <br> - An empty recipient record is added to the `Recipient` in db. (at first I thought this was unnecessary,  but the implementation of  [useFieldArray](https://react-hook-form.com/api/usefieldarray/) forces us to so) <br> - This is why the status is `pending` even after completion|
|`fix`| - A possible workaround is deleting the empty records (from db and fieldArray) when user clicks `Send`. |
|preview| https://www.loom.com/share/1eadbf82efcb4de2ac0b0e00b62cf01f| 

I have opened another related issue here #69

Fixes #39

### DIFF
--- a/apps/web/pages/documents/[id]/recipients.tsx
+++ b/apps/web/pages/documents/[id]/recipients.tsx
@@ -68,6 +68,20 @@ const RecipientsPage: NextPageWithLayout = (props: any) => {
     return !!errors?.signers?.[index]?.email;
   };
 
+  const deleteEmptyFields = () => {
+    let counter = 0;
+    formValues.forEach((item: any, index) => {
+      if (!item.email && item.sendStatus === "NOT_SENT") {
+        remove(index - counter);
+        counter += 1;
+        deleteRecipient(item)?.catch((err) => {
+          append(item);
+          counter -= 1;
+        });
+      }
+    });
+  };
+
   return (
     <>
       <Head>
@@ -109,6 +123,7 @@ const RecipientsPage: NextPageWithLayout = (props: any) => {
                   icon={PaperAirplaneIcon}
                   onClick={() => {
                     setOpen(true);
+                    deleteEmptyFields();
                   }}
                   disabled={
                     (formValues.length || 0) === 0 ||


SauravL3010:
- Yes, I get what you are saying. imo, "Draft recipients" is a good thing, because then users don't have to care about, deleting the empty ones. More importantly, as you mentioned, user might leave it at filing the `Name` tag, for later use maybe ? (which is a nice to have). 
 - One drawback is having lots of empty recipients in db, which imo, not necessarily a bad thing. 
 - An alternative way to address #39.
 - is to have simple, `if (!recipient.email)` then don't bother adding that as a valid recipient to the document. thoughts on that ?

ElTimuro:
- By don't bother adding as a valid recipient you mean valid for sending? Or creating serverside at all?
- Serverside would mean reworking the architecture a bit
---
- The longer I think about it, I think the easier way would be to block sending until every recipient is valid or deleted
- That allows "draft recipients" until sending, but no zombies after sending
- There would need to be a hint, informing the user
- This would also be very fast, which is a huge plus since we are looking toward launch

SauravL3010:
I think you have it right !
Allowing "draft recipients" but giving some `invalid email` feedback before sending it, could work. (So, i guess the user is either forced to remove it, or enter a valid email)

I will work on this !

ElTimuro:
@SauravL3010 let us know, if you need help merging, but I think this is just your stuff?

SauravL3010:
I am closing this PR. This implementation  is not quite right. I think handling exceptions (eg. empty emails) server side would be a better approach. see @Mythie comment [here](https://github.com/documenso/documenso/issues/67#issuecomment-1517867557)

SauravL3010:
I am closing this PR. This implementation  is not quite right. I think handling exceptions (eg. empty emails) server side would be a better approach. see Mythie comment [here](https://github.com/documenso/documenso/issues/67#issuecomment-1517867557)

PR #71, opened 2023-04-11
fixes - bug in disabling send button when email field is invalid

SauravL3010:
This is already implemented, its' just a minor boolean error.

preview: https://www.loom.com/share/4b6492e49bcc46dd9fd5d150ae0f2847

### DIFF
--- a/apps/web/pages/documents/[id]/recipients.tsx
+++ b/apps/web/pages/documents/[id]/recipients.tsx
@@ -112,8 +112,8 @@ const RecipientsPage: NextPageWithLayout = (props: any) => {
                   }}
                   disabled={
                     (formValues.length || 0) === 0 ||
-                    !formValues.some(
-                      (r: any) => r.email && !hasEmailError(r) && r.sendStatus === "NOT_SENT"
+                    formValues.some(
+                      (r: any) => r.email && hasEmailError(r) && r.sendStatus === "NOT_SENT"
                     ) ||
                     loading
                   }>


ElTimuro:
1. Changing it as this allows sending without any valid email
image

2. Is blocking sending with an invalid mail so bad? I would assume, the user would want to fix that before sending (or leave it blank, consciously)

SauravL3010:
1. Oh yes, i may have missed that  effect. 
2. So my thought process was, that the user must enter a valid email, before sending it. (or must remove it)
there is one side effect to allowing users to send, when recipient emails is invalid. Lets say, the user entered the correct email at first (which is saved to db), but then somehow manages to add a letter or something, which will make the email invalid (not saved in db), upon sending, it sends it to the original correct email, which may or may not be intended. I might be overthinking this, so sorry about that 😅

ElTimuro:
> * Oh yes, i may have missed that  effect.
> * So my thought process was, that the user must enter a valid email, before sending it. (or must remove it)
>   there is one side effect to allowing users to send, when recipient emails is invalid. Lets say, the user entered the correct email at first (which is saved to db), but then somehow manages to add a letter or something, which will make the email invalid (not saved in db), upon sending, it sends it to the original correct email, which may or may not be intended. I might be overthinking this, so sorry about that 😅

No no, it's good to think these flows really through 👌 If we choose my suggestion here (blocking send with hint): This would be already as it should be. Shoudl we do that?

PR #73, opened 2023-04-12
Allow copying of signer's link

### LINKED ISSUE #36: [DOC-70] Share Signing Link instead of Email
Allow to copy a Link for each Signer allow using Documenso without an Email Provider.

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-70](https://linear.app/documenso/issue/DOC-70/share-signing-link-instead-of-email)</sub>

dephraiim:
- [x] Copy link on icon click
- [x] Show toast on link copied
- [x] Hide the `Not Sent` when `readStatus` is seen
- [x] Add `COPIED` status and when the link is copied, the status is shown as "Public Link Copied" of `Not Sent`

### Screenshot
Screenshot 2023-04-12 at 09 13 08

Closes #36


ElTimuro:
1. Since the link is kind of already there, just the delivery changed one could keep it like this and just hide the not send when readStatus is seen => no extra field. Is the information that a link was copied but not seen yet relevant? Would be easier to ignore for now since the doc could also be sent via mail additionally.
2. Yes, there should be something nicer here. One could force email or name for link copying or fill it with sth. linke "Recipient 1"

dephraiim:
> Is the information that a link was copied but not seen yet relevant? Would be easier to ignore for now since the doc could also be sent via mail additionally.

I agree. It may not be relevant for now but for audit and compliance, it should definitely be added in the future. 

Knowing if a link was copied can help in establishing a clear audit trail and demonstrating that proper signing procedures were followed. It can also provide evidence in case of any disputes or legal issues related to the signing process.

For teams, yet to be added, knowing if a link was copied can help you monitor and track any potential unauthorized access or misuse of the signing link.

dephraiim:
@Mythie @ElTimuro I have updated this. Can you take a look?

dephraiim:
@ElTimuro @Mythie Can you take a look? Fixed the conflicts

apps/web/pages/api/documents/[id]/copied.ts
```diff
@@ -0,0 +1,37 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { setCopiedField } from "@documenso/lib/copied";
+import { defaultHandler, defaultResponder, getUserFromToken } from "@documenso/lib/server";
+import prisma from "@documenso/prisma";
+
+async function getHandler(req: NextApiRequest, res: NextApiResponse) {
+  const user = await getUserFromToken(req, res);
+  const { id: recipientId } = req.query;
+
+  if (!user) return;
```
Comment (Mythie) [inline apps/web/pages/api/documents/[id]/copied.ts:10]: We should return a response here

apps/web/pages/api/documents/[id]/copied.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  if (!recipientId) {
```
Comment (Mythie) [inline apps/web/pages/api/documents/[id]/copied.ts:12]: Which could be done as a compound check here: `if (!recipientId || !user)`

apps/web/pages/api/documents/[id]/copied.ts
```diff
@@ -0,0 +1,37 @@
...
+    return;
+  }
+
+  const recipients = await prisma.recipient.findMany({
+    where: {
+      id: +recipientId,
+    },
+  });
+
+  console.log("api/documenso/id/copied", recipients);
```
Comment (Mythie) [inline apps/web/pages/api/documents/[id]/copied.ts:23]: Won't be needing this any more

apps/web/pages/api/documents/[id]/copied.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  if (!recipients.length) return res.status(200).send(recipients.length);
```
Comment (Mythie) [inline apps/web/pages/api/documents/[id]/copied.ts:25]: No single line if's please 😄

apps/web/pages/api/documents/[id]/copied.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  recipients.forEach(async (recipient) => {
```
Comment (Mythie) [inline apps/web/pages/api/documents/[id]/copied.ts:27]: This won't work the way you're expecting, you could do a `for of` loop but you'd still need to consider how you would rollback when a later item errors (think transactions)

packages/lib/api/setCopiedField.ts
```diff
@@ -0,0 +1,21 @@
...
+      method: "GET",
+    })
+      .then((res: any) => {
+        console.log("documenso/lib/api/setCopiedField", res);
+
+        if (!res.ok) {
+          throw new Error(res.status.toString());
+        }
+      })
+      .finally(() => {
```
Mythie, line 15: Why are we reloading here?

dephraiim:
Closing this PR in favor of the updates in the [feat/refresh](https://github.com/documenso/documenso/tree/feat/refresh) branch.

PR #74, opened 2023-04-13
chore: optimise depedency tree

Mythie:
Tidies up the dependency tree moving packages into `devDeps` where applicable, doing so means that docker container builds become about 60mb smaller.

image


Mythie:
Force merging as well since the blast radius is pretty small.

boom

PR #77, opened 2023-04-15
Support Next.js standalone builds.

Mythie:
**Letting this hang for until we deal with general docker support.**

Supporting standalone builds lets us shrink our image size dramatically much like #74 making for leaner and faster image in terms of startup time.

One thing to note here is that our usage of `NEXT_PUBLIC_WEBAPP_URL` means that each individual would need to build the image before they can deploy it to their domain.

image


ElTimuro:
Since there will be even more stuff, loaded from the ENV Config, we need to think of a way to let self-hosters easily configure this.
- Building the image after creating a custom config or just requiring to change it in the image is the most basic way IMO
- A web UI based setup like, the creates the local ENV could be the upgrade version

leerob:
Was going to comment and suggest this! We strongly recommend using this for anyone using Docker.

PR #127, opened 2023-04-16
replace neon with zinc-800

PeerRich:
neon has not enough contrast and is damaging a11y.

before a full-time designer is on board, I would prefer to use default colors that have a good color palette 

before:

CleanShot 2023-04-16 at 13 22 37@2x

after:

CleanShot 2023-04-16 at 13 22 26@2x

CleanShot 2023-04-16 at 13 26 55@2x

### DIFF
--- a/apps/web/components/editor/field-type-selector.tsx
+++ b/apps/web/components/editor/field-type-selector.tsx
@@ -36,7 +36,7 @@ export default function FieldTypeSelector(props: any) {
             value={fieldType.id}
             className={({ checked, active }) =>
               classNames(
-                checked ? "border-neon border-2" : "border-transparent",
+                checked ? "border-2 border-zinc-800" : "border-transparent",
                 "relative block cursor-pointer select-none rounded-lg border bg-white px-3 py-2 hover:bg-slate-100 focus:outline-none sm:flex sm:justify-between"
               )
             }>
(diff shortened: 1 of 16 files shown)


ElTimuro:
- True and the black and white looks really clean.
- Would be nice to keep neon somewhere as a splash of color
- If I can't come up with something better, this could be the Launch Version for now

PeerRich:
found some issues with background. fixing

PeerRich:
keeping neon here:
CleanShot 2023-04-16 at 13 48 34@2x

PeerRich:
i think neon as BG is fine with dark font on it.

but neon should not be used as text on white

ElTimuro:
> i think neon as BG is fine with dark font on it.
> 
> but neon should not be used as text on white

That's true. Maybe Buttons with dark text?

Mythie:
But I like inflicting blindness upon myself 😄 

I think we could look at reducing the intensity of the neon so it can continue being used in a more general fashion on dark text.

I'll pop open the old WGAG colorpicker shortly.

ElTimuro:
@Mythie I would like to keep the neon for now, since we are new and edgy. But the buttons should have dark font and here should be no texts in neon on white like "signup here". Also Neon icons should have a dark background.

Like Penta, just more minimal an way less messy:

image

Mythie:
Closing for now since we love blinding lights! Avert thy eyes @PeerRich 
blinding-lights

PR #137, opened 2023-04-19
Redirect to /dashboard if auth user tries to access /login or /signup

### LINKED ISSUE #131: New user not redirected to `/dashboard`
- When a user creates a new account on the `/signup`, the user should be redirected to the `/dashboard` on completion
- A logged-in user is able to access `/signup` and `/login`. Logged In users should be redirected to the `/dashboard`
- When a user is logged in, they can create another user account, whilst logged in.

zahid47:
fixes #131

I think we should be using [next middleware](https://nextjs.org/docs/advanced-features/middleware) to do all the redirecting. What do you think? Do you want me to try to do that instead?

### DIFF
--- a/apps/web/pages/login.tsx
+++ b/apps/web/pages/login.tsx
@@ -1,4 +1,5 @@
 import Head from "next/head";
+import { getUserFromToken } from "@documenso/lib/server";
 import Login from "../components/login";
 
 export default function LoginPage(props: any) {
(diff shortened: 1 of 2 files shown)


dephraiim (approved):
LGTM 👍🏽

Mythie (approved):

Mythie:
Looks great! While a middleware would be nice this is small in scope and means we won't be checking the token for every request.

giphy (2)

PR #138, opened 2023-04-19
feat: add name field

### LINKED ISSUE #53: [DOC-170] Name Field
* Field for Signer Name (in addition to the existing signature field and date field)
* Prefilled with Signer Name from Recipient (The name set for the recipient, the field is associated to should be prefilled)
* Freely editable (The content of the field should be freely editable as text field)

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-170](https://linear.app/documenso/issue/DOC-170/name-field)</sub>

Mythie:
Adds support for a name field which will be pre-filled with the recipients name if they haven't signed a form on Documenso before.

Video of the functionality can be found below, do note that some styling has been done after the fact to reduce font sizes a little:

https://www.loom.com/share/e99ec698fd4242dc901c77b8f5986741

Closes #53


dephraiim (dismissed):

ElTimuro:
- Video looks good, but for some reason, the preview env does not yet show the feature?
- Updated Branch, maybe rebuild will help
-  We could merge IMO, but I would like to take the opportunity to get previews working

image

Mythie:
I think that’s a wise idea!

ElTimuro:
- @Mythie I figured it out: The ENV config for the preview ENV, still has the web app URL of test (API root, navigation root).
- I think this would be fixable via YAML
- Will test and merge this locally first tho
- @dephraiim thx for the 1st review

ElTimuro (changes requested):
One Remark:
- The name should be prefilled in the field, not just the dialog
- Best Case, It's what you want and you do not need to click
- Otherwise great

Mythie:
Updated base on the comments provided above, render is still not using the latest code from this PR though.

ElTimuro:
@Mythie i did not get around to make the env via yaml

ElTimuro:
Redeploying PR env with url support

ElTimuro (approved):

ElTimuro (changes requested):
- @Mythie PR Preview is working
- Name is getting profiled correctly

- One More Change: Just adding and filling in a name FIeld allows you to complete the signing.
- If there are signature fields they all need to be filled
- If there are none (this case) at least one free signature need to be set
- Until these conditions are met, the DONE button has to stay disabled

Mythie:
> * @Mythie PR Preview is working
> * Name is getting profiled correctly
> * One More Change: Just adding and filling in a name FIeld allows you to complete the signing.
> * If there are signature fields they all need to be filled
> * If there are none (this case) at least one free signature need to be set
> * Until these conditions are met, the DONE button has to stay disabled

Updated the code to require one free signature if no signature field is found.

ElTimuro:
- Error when placing free signature and Filling out name Field (not occurring with empty name Field)
- Signing does not fail cleanly, see image. The state implies more signatures, but there is only one, that did not work.
image

Mythie:
> * Error when placing free signature and Filling out name Field (not occurring with empty name Field)
> * Signing does not fail cleanly, see image. The state implies more signatures, but there is only one, that did not work.
> * No migration execution  is missing
>   image

I'm not encountering this locally, could it be due to the migrations being required for this PR? (to add the new enum value)

ElTimuro:
- As far as I saw, migrations where up to date, but smells like sth. in that direction
- I'll check again, I want some PR Preview env management experience anyway

apps/web/components/editor/pdf-signer.tsx
```diff
@@ -16,21 +17,62 @@ const PDFViewer = dynamic(() => import("./pdf-viewer"), {
...
+  const [signatureDialogOpen, setSignatureDialogOpen] = useState(false);
+  const [nameDialogOpen, setNameDialogOpen] = useState(false);
   const [signingDone, setSigningDone] = useState(false);
   const [localSignatures, setLocalSignatures] = useState<any[]>([]);
   const [fields, setFields] = useState<any[]>(props.fields);
-  const signatureFields = fields.filter((field) => field.type === FieldType.SIGNATURE);
+  const signatureFields = useMemo(
+    () => fields.filter((field) => [FieldType.SIGNATURE].includes(field.type)),
+    [fields]
+  );
```
kulgavy, line 28: Did you mean to extent this array? 🤔 If not, not sure why this change was made

Mythie: This is made to support other signature like fields later on like `name` which was in a different PR 😄

PR #141, opened 2023-04-21
fix: date field appears for all recipients

### LINKED ISSUE #48: [DOC-214] Date Field is already fulfilled for the second recipient after the first recipient has signed
<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-214](https://linear.app/documenso/issue/DOC-214/date-field-is-already-fulfilled-for-the-second-recipient-after-the)</sub>

Mythie:
Updates the signing endpoint to only apply changes to the Date field for the current signer. This is made possible through the addition of the `signedAt` column within the database.

Resolves the issue with one signer filling the date for all recipients and also ensures that the date of signing on a document won't always be todays date after each recipient has signed.

Closes #48

### DIFF
--- a/apps/web/pages/api/documents/[id]/sign.ts
+++ b/apps/web/pages/api/documents/[id]/sign.ts
@@ -63,6 +63,7 @@ async function postHandler(req: NextApiRequest, res: NextApiResponse) {
     },
     data: {
       signingStatus: SigningStatus.SIGNED,
+      signedAt: new Date(),
     },
   });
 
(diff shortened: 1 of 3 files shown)


PeerRich (approved):

dephraiim (approved):

ElTimuro (approved):
- Tested and looks good
- I'm merging this, so it's finally done
- We should think about Moving the Sign-Date to the field. Since it's kinda the value inserted into the field. Seems more natural. Thought?
- But fine for now.

Mythie:
I would have to agree, I initially went to place it on the field but found it wasn't quite natural in the current state.

ElTimuro:
@Mythie great Job 
image

PR #146, opened 2023-05-06
Implement Google OAuth for user authentication

### LINKED ISSUE #106: [DOC-31] Register/ Login with Google
* Allowing different Auth-Providers reduces friction in all instances of Documenso
* Google should be among the first to be implement for it's widespread use 

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-31](https://linear.app/documenso/issue/DOC-31/register-login-with-google)</sub>

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-31](https://linear.app/documenso/issue/DOC-31/register-login-with-google)</sub>

dephraiim:
Closes #106


apps/web/components/login.tsx
```diff
@@ -123,40 +124,95 @@ export default function Login(props: any) {
...
               </div>
               <div>
-                <div className="relative">
+                <div className="relative my-2">
                   <div className="absolute inset-0 flex items-center" aria-hidden="true">
                     <div className="w-full border-t border-gray-300" />
                   </div>
-                  <div className="relative flex justify-center"></div>
+                  <div className="relative flex justify-center text-sm">
+                    <span className="bg-gray-100 px-2 text-gray-500">Or continue with</span>
```
Mythie, line 140: This can probably just be `or` and we can apply an `uppercase` class since the below provider says "Sign in with Google"

package.json
```diff
@@ -29,7 +29,7 @@
     "avatar-from-initials": "^1.0.3",
     "bcryptjs": "^2.4.3",
     "next": "13.2.4",
-    "next-auth": ">=4.20.1",
+    "next-auth": "^4.22.1",
```
Mythie, line 32: Don't think this should be in the root `package.json`

packages/lib/server/getUserFromToken.ts
```diff
@@ -1,13 +1,20 @@
...
 
 export async function getUserFromToken(
   req: NextApiRequest,
   res: NextApiResponse
 ): Promise<PrismaUser | null> {
+  const oauthUser = await getServerSession(req, res, authOptions);
+
+  if (oauthUser) {
+    return oauthUser.user as PrismaUser;
+  }
```
Mythie, line 16: Can we still not use `getToken` here? It should work now with the JWT strategy?

apps/web/package.json
```diff
@@ -15,13 +15,14 @@
...
     "@heroicons/react": "^2.0.13",
+    "@next-auth/prisma-adapter": "^1.0.6",
     "@pdf-lib/fontkit": "^1.1.1",
+    "@prisma/client": "^4.13.0",
     "avatar-from-initials": "^1.0.3",
     "base64-arraybuffer": "^1.0.2",
     "bcryptjs": "^2.4.3",
     "formidable": "^3.2.5",
     "next": "13.2.4",
-    "next-auth": "^4.22.0",
```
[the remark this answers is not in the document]
dephraiim: Nope, let me fix it

packages/lib/server/getUserFromToken.ts
dephraiim: Yep. I'll fix it

Mythie:
Does it all work as expected after those changes? 😅

dephraiim:
> Does it all work as expected after those changes? 

Yes, I tested it right after. 
The most important thing is to fetch the user, which then fits into the existing flow of creating and sending documents.

The user is not null for both OAuth and Credentials Login but more tests can still be done.

ElTimuro:
@dephraiim looks like there are some conflicts. can you resolve before i test?

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -69,23 +75,49 @@ export default NextAuth({
...
+              identityProvider: IdentityProvider.GOOGLE,
+            },
+          })
+          .then(() => {
+            return true;
+          });
+      }
+
+      return true;
+    },
```
Comment (kulgavy) [inline apps/web/pages/api/auth/[...nextauth].ts:118]: Should work just fine, no need to mix .then/.catch with async/await

```suggestion
    async signIn({ user, account }) {
      if (account?.provider === "google") {
        await prisma.user
          .update({
            where: {
              email: user.email!,
            },
            data: {
              identityProvider: IdentityProvider.GOOGLE,
            },
          })
      }

      return true;
    },
```

apps/web/pages/api/auth/[...nextauth].ts
```diff
@@ -69,23 +75,49 @@ export default NextAuth({
...
         ...token,
       };
     },
-    async session({ session, token }) {
+    async session({ session, token, user }) {
       const documensoSession: Session = {
         ...session,
         user: {
           ...session.user,
+          id: Number(token.sub),
```
Comment (kulgavy) [inline apps/web/pages/api/auth/[...nextauth].ts:94]: Any specific reason to have id as a number in `session` on the client?

apps/web/components/login.tsx
```diff
@@ -123,40 +124,95 @@ export default function Login(props: any) {
...
+                          fill="#fbbc05"
+                        />
+                        <path
+                          d="M-1089.333-727.244c14.028 0 26.497 4.849 36.455 14.201l27.276-27.276c-16.539-15.413-38.013-24.852-63.731-24.852-37.234 0-69.359 21.388-85.032 52.561l31.692 24.592c7.533-22.514 28.575-39.226 53.34-39.226z"
+                          fill="#ea4335"
+                          clipPath="none"
+                          mask="none"
+                        />
+                      </g>
+                    </svg>
```
kulgavy, line 182: Would be cool if it's possible to move inline svg icons to "components/icons/<IconName>.tsx" or something similar just to not keep it inline (and do not repeat if it needed in the other place)

apps/web/pages/api/auth/[...nextauth].ts
Comment (dephraiim) [reply apps/web/pages/api/auth/[...nextauth].ts]: The auth was failing before because the user is being updated before it is created in the `signIn` callback.

Initially, it may have appeared to work because I had a user with the same email already present in the local db. This situation might have been the case for @Mythie as well. 

However, when starting with a fresh database that only includes the seed user, this callback causes the authentication process to fail.

apps/web/pages/api/auth/[...nextauth].ts
Comment (dephraiim) [reply apps/web/pages/api/auth/[...nextauth].ts:102]: From the schema, the id is stored as a number, hence we are maintaining the same type across the application.

apps/web/components/login.tsx
dephraiim: The Google Logo will hopefully be used once so it is safe to keep it here

dephraiim:
@Mythie I have fixed all the conflicts

Mythie:
Closing as we've dealt with this on `feat/refresh` 🎉

dephraiim:
Captain America Salute

Issue #76, opened 2023-04-14
Files with long names is breaking UI


dephraiim:
Files with long file names are breaking the UI. 

This is a sample screenshot

<img width="546" alt="Screenshot 2023-04-14 at 21 47 54" src="https://user-images.githubusercontent.com/55143799/232160903-4a51ebab-9a16-4f3e-b53b-f1a3d907fac3.png">
<img width="1440" alt="Screenshot 2023-04-14 at 21 48 52" src="https://user-images.githubusercontent.com/55143799/232160911-e6dd4427-2b93-4584-a4b7-46f2c26bdbbc.png">

ElTimuro:
- Low priority.
- We should think about abbreviating everything longer than a context-specific length to "myverylongstri...".
- Should be well fixable with word wrap

TheoOliveira:
Hey guys, can I be assign to this?

dephraiim:
Heyyy @TheoOliviera, it is being worked on at #150.

But you can take a look at this #169

Mythie:
Closed as part of #150 

Thanks everyone!

PR #150, opened 2023-05-10
Long filename fix 🗄

The-Robin-Hood:
**PR** : The pull request addressed an issue with long file names in the UI 
**Issue** : #76

#### Before: 
image

#### After:
image


dephraiim (commented):
The Dialog component is currently in the process of being reusable. Modifying the title in the component could limit the component's reusability.

The issue also affects the signed page

The-Robin-Hood:
How about we having extra param - boolean in Dialog Component and to shorten the title if that param is only true ?

Mythie:
Yep I'm for a `truncateTitle` prop so we can opt in and out as desired. Obviously make it opt-out by default 😄

The-Robin-Hood:
@dephraiim Hope this works. 
@Mythie Thanks for the suggestion

ElTimuro (dismissed):
- Checked and works
- The signed page still needs to be fixed (and all other places, where those properties are used)
- @The-Robin-Hood Do you want to fix those too or should I create a new ticket?

The-Robin-Hood:
I suggest creating a new ticket to address those specific issues. This will allow us to track and prioritize the necessary fixes separately from the current pull request. Thank you.

packages/ui/components/dialog/Dialog.tsx
```diff
@@ -29,11 +31,14 @@ export function Dialog({
   formValues,
   setLoading,
   icon,
+  truncateTitle = true,
```
kulgavy, line 34: Is it used? 🤔

Mythie: Good catch! Updated to fix that oversight on my end 😅

Mythie (approved):

Issue #152, opened 2023-05-21
[DOC-215] Wrong document and sender in an email


ElTimuro:
* Wrong document and sender in an email [Sign | Documenso](https://app.documenso.com/documents/6/signed), after a couple of minutes the right Link appeared [Sign | Documenso](https://app.documenso.com/documents/15/sign?token=dzU6Jv6YJFbUy2dqS1qtNA) 

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-215](https://linear.app/documenso/issue/DOC-215/wrong-document-and-sender-in-an-email)</sub>

ElTimuro:
Cause: token query param from link in email is not passed to /signed page

ElTimuro:
[Sign | Documenso](https://test.documenso.com/documents/14/signed)

ElTimuro:
still bugy

Link in E-Mail follows to another document which wasn't event sent to the mail owner

![image.png](https://uploads.linear.app/884f43d5-69da-459d-bc6b-864f16914e0f/a85399d6-44d5-413d-9c8f-b6a39737bddd/bc00d46e-c97c-48ec-9c6f-322abaec43ce)

ElTimuro:
Retest please

ElTimuro:
@g.barzel  fixed, ready for retest

PR #153, opened 2023-05-21
fix: pass recipient token to signed page

ElTimuro:
fixes https://github.com/documenso/documenso/issues/152


ElTimuro:
Merging for manual retest

PR #157, opened 2023-05-25
feat: introduce Input component

nikospapcom:
I create an input component in order to be have more reusable inputs

### DIFF
--- a/packages/ui/components/input/Input.tsx
+++ b/packages/ui/components/input/Input.tsx
@@ -1,0 +1,36 @@
+import React, { InputHTMLAttributes, ReactNode, forwardRef, useId } from "react";
+import { classNames } from "@documenso/lib";
+
+interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
+  label?: string;
+  labelSrOnly?: boolean;
+}
+
+export const Input = forwardRef<HTMLInputElement, InputProps>((props, ref) => {
+  const { label, labelSrOnly, className, ...rest } = props;
+  const id = useId();
+
+  return (
+    <div>
+      {label && (
+        <label
+          htmlFor={id}
+          className={classNames(
+            "mb-1 block text-sm font-medium text-gray-700",
+            labelSrOnly && "sr-only"
+          )}>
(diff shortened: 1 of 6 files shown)


packages/ui/components/input/Input.tsx
```diff
@@ -0,0 +1,36 @@
...
+          )}>
+          {label}
+        </label>
+      )}
+      <input
+        id={id}
+        ref={ref}
+        {...rest}
+        className={classNames(
+          "focus:border-neon focus:ring-neon mt-1 block w-full rounded-md border border-gray-300 py-2 px-3 shadow-sm focus:outline-none sm:text-sm",
```
kulgavy, line 30: I think it's better to have `mb-1` class on label rather then `mt-1` on input just to avoid unexpected margins if `label` is not provided

packages/ui/components/input/Input.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { InputHTMLAttributes, ReactNode, forwardRef, useId } from "react";
+import { classNames } from "@documenso/lib";
+
+interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
+  label?: string;
+  labelSrOnly?: boolean;
```
kulgavy, line 6: Maybe it will be a bit more flexible to have `labelClassName` instead of this prop? WDYT?

packages/ui/components/input/Input.tsx
nikospapcom: Nice suggestion

packages/ui/components/input/Input.tsx
nikospapcom: No I don't think it's a good idea :)

nikospapcom:
> Hey Nikos, welcome to Documenso! I'll look at your changes very soon 😄

hey @Mythie, any update on this?
I saw this [pr](https://github.com/documenso/documenso/pull/196/files) and we use hard code inputs instead of reusing the component I built :)

Mythie:

PR #159, opened 2023-05-26
feat: support leading cert from custom path

Mythie:
Support loading certificate files from custom paths, useful for those who keep their certificates outside of the project in places like a keychain or similar.


ElTimuro:
- Looks good
- We should introduce a npm command to generate a locally signed cert as part of the dev quickstart/ self-host setup

ElTimuro (approved):

PR #165, opened 2023-05-29
typo: /ressources to /resources

doug-andrade:


Mythie:
Thanks! I was about to update this yesterday before launch as I noticed it during some buffer debugging.

I'll run some tests shortly otherwise it'll be good to merge 😄

Mythie (approved):

Issue #175, opened 2023-05-30
docker-compose command not found


The-Robin-Hood:
Issue Statement:

The package json script contains a docker-compose command that is v1 docker compose command which in v2 it is switched to "docker compose". This inconsistency can cause errors and delays in the installation process, especially for novice users who are unfamiliar with Docker.

Recommendation:

To resolve this issue, it is suggested that the package.json script should be updated to include both versions of the command to accommodate all systems.

Mythie:
Yep this is going to be because I'm a tech boomer 😂 

Appreciate the issue and the PR to resolve

nikmartin:
the patch doesn't work on linux. I'll see if I can find a workaround

The-Robin-Hood:
@nikmartin whats the prob you are facing ? Coz it did worked for me

nikmartin:
running this in bash: 
` docker compose -f ./docker/compose-without-app.yml || docker-compose -f ./docker/compose-without-app.yml up -d`

returns:

```Usage:  docker compose [OPTIONS] COMMAND

Docker Compose
...
```

BUT adding the full command to each option: 
 `docker compose -f ./docker/compose-without-app.yml up -d || docker-compose -f ./docker/compose-without-app.yml up -d`

works for me

The-Robin-Hood:
One way to fix is 

```
"docker:compose-up": "docker compose -f ./docker/compose-without-app.yml up -d || docker-compose -f ./docker/compose-without-app.yml up -d",
"docker:compose-down": "docker compose -f ./docker/compose-without-app.yml down || docker-compose -f ./docker/compose-without-app.yml down"
 ```
 
 and remove docker-compose

Mythie:
Sounds like the way to go! Sorry for the delay, you can either raise a PR or I can do so shortly :)

Mythie:
Resolved with #184

PR #184, opened 2023-06-01
docker script updated 🐳

The-Robin-Hood:
Issue #175 

Package json script for docker updated, to work on various platforms


Mythie (approved):

Mythie:
Thanks for contributing! It was a bit dicey but we got there in the end 🎉

PR #188, opened 2023-06-02
Fixed alignment of div containing selectboxes in mobile and tablet screens

Ashutosh-Bhadauriya:
In your current situation, the `flex-row-reverse` property is causing the alignment to be to the right even on smaller screens (2nd screenshot). It should be on the left imo.

Screenshot (0)

Screenshot (1)

Recommended fix - We could use `justify-start` for smaller screens and `justify-end` for bigger screens. So on smaller screens, the items will start from the left side, and on medium and larger screens, they will start from the right side, while maintaining the same order.
Screenshot (2)

### DIFF
--- a/apps/web/pages/documents.tsx
+++ b/apps/web/pages/documents.tsx
@@ -145,16 +145,16 @@ const DocumentsPage: NextPageWithLayout = (props: any) => {
             </Button>
           </div>
         </div>
-        <div className="mt-3 mb-12 flex items-center justify-start gap-x-4 md:justify-end">
+        <div className="mt-3 mb-12 flex flex-wrap items-center justify-start gap-x-4 md:justify-end gap-y-4">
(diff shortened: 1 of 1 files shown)


Ashutosh-Bhadauriya:
For some reason, my formatter has changed the order of tailwind classes and imports on this page. Please let me know if that works, otherwise will change it. Thanks!

Mythie:
Hey Ashutosh, welcome to Documenso! Nice to see you here after being on Twitter 😄 

I agree that the styling on mobiles devices there is pretty grim, I'll test the PR locally and maybe also make the wrapping element full width as well.

Mythie (approved):

Mythie:
Thanks for your contribution! Sorry I took so long 😅 

thanks-adventure-time

PR #199, opened 2023-06-05
fix: expired slack link

dephraiim:


Mythie:
We should use the shortlink for this: https://documen.so/slack 😄

dephraiim:
Does the short link redirect to the new invite link?

Mythie:
> Does the short link redirect to the new invite link?

Sure does :smile:

Mythie (approved):

Mythie:
Thanks Ephraim! 😄

PR #209, opened 2023-06-15
fix: don't double send error reponses when sending fails

Mythie:


dephraiim (approved):

Mythie:
Too many formatting changes hol' up

PR #213, opened 2023-06-21
feat: dropzone transition animation

dephraiim:
feat: dark mode & theme switching


Mythie:
Was this not working? I feel like it did when I briefly swapped to dark theme 🤔

dephraiim:
It was working but the issue was the order. The animation did not happen in parallel, it was in sequece, the right, left then middle.

Setting a duration will start the animation for all three at the same time

Mythie:
That's odd, it should be left, center, then middle. The animation being staggered is intentional though since it feels more "alive" 😸

PR #217, opened 2023-06-25
feat: stack avatars

dephraiim:
feat: dark mode & theme switching


apps/web/src/components/(dashboard)/avatar/stack-avatar.tsx
```diff
@@ -0,0 +1,36 @@
...
+      classes = 'bg-documenso-200 text-documenso-800';
+      break;
+    default:
+      break;
+  }
+
+  return (
+    <Avatar
+      className={`
+        ${zIndex && `z-${zIndex}`}
```
Mythie, line 29: This won't work the way you think it might, it'll likely get purged in a production build.

packages/lib/client-only/recipient-type.ts
```diff
@@ -0,0 +1,8 @@
+import { Recipient } from '@documenso/prisma/client';
+
+export const type = (recipient: Recipient) =>
+  recipient.sendStatus === 'SENT' && recipient.signingStatus === 'SIGNED'
+    ? 'completed'
+    : recipient.sendStatus === 'SENT' && recipient.signingStatus === 'NOT_SIGNED'
+    ? 'waiting'
+    : 'unsigned';
```
Mythie, line 8: Prefer this:

```typescript
const getRecipientType = (recipient: Recipient) => {
  if (recipient.sendStatus === 'SENT' && recipient.signingStatus === 'SIGNED') {
    return 'completed';
  }

  if (recipient.sendStatus === 'SENT' && recipient.signingStatus === 'NOT_SIGNED') {
    return 'waiting';
  }

  return 'unsigned';
};
```

packages/lib/types/document-with-recipient.ts
```diff
@@ -0,0 +1,5 @@
+import { Document, Recipient } from '@documenso/prisma/client';
```
Mythie, line 1: This could probably sit somewhere in the prisma package since it relates directly to database types

apps/web/src/components/(dashboard)/avatar/stack-avatar.tsx
```diff
@@ -9,6 +9,9 @@ export type StackAvatarProps = {
 
 export const StackAvatar = ({ first, zIndex, fallbackText, type }: StackAvatarProps) => {
   let classes = '';
+  let zIndexClass = '';
+  const firstClass = first ? '' : '-ml-3';
```
Mythie, line 13: We can probably do this with `not:first-child:-ml-3` or something like that.

apps/web/src/components/(dashboard)/avatar/stack-avatar.tsx
```diff
@@ -23,12 +26,32 @@ export const StackAvatar = ({ first, zIndex, fallbackText, type }: StackAvatarPr
       break;
   }
 
+  switch (zIndex) {
```
Mythie, line 29: Let's just use an object outside of the component for this? so it can be `ZIndexes = { ... }; ZIndexes[key] ?? ''`

dephraiim:
finally

PR #256, opened 2023-08-14
Add husky & lint staged

nsylke:
I didn't noticed an issue for this yet, but I did read the blog post about adding it - so this PR implements husky & lint-staged to ensure all future commits are formatted. There are a few ways of configuring lint-staged, I just took the straightforward approach, happy to change it - if preferred.

Additionally, I modified the Dockerfile to disable husky from being installed. You'll see something along the lines of this in the docker build logs when installing dependencies, indicating its been disabled.
```bash
. prepare$ husky install
. prepare: husky - HUSKY env variable is set to 0, skipping install
. prepare: Done
```

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,6 +22,9 @@ RUN apk add --no-cache libc6-compat
 # Copy our current monorepo
 COPY . .
 
+# Disable husky from installing hooks
+ENV HUSKY 0
+
 RUN npm ci
 
 RUN npm run build --workspaces
--- a/lint-staged.config.cjs
+++ b/lint-staged.config.cjs
@@ -1,0 +1,5 @@
+module.exports = {
+  "**/*.{ts,tsx,md}": [
+    "prettier --write"
+  ]
+}
--- a/lint-staged.config.cjs
+++ b/lint-staged.config.cjs
@@ -1,5 +1,3 @@
 module.exports = {
-  "**/*.{ts,tsx,md}": [
-    "prettier --write"
-  ]
-}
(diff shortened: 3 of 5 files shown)


Mythie:
Hey Nicholas, great work! You beat me to creating the issues surrounding those items 😅 

As for the configuration it's preferred to use `lint-staged.config.cjs` so we can do some fancy things later on with other file types 🚀

nsylke:
@Mythie Certainly! Would you like that to be in its own package similar to prettier and then require it?

Mythie:
It can just sit in the root of the monorepo for now 😁

nsylke:
Updated to use `lint-staged.config.cjs`. I noticed prettier (`npm run format`) isn't formatting any of the `cjs` files, is this intentional?

Mythie:
> Updated to use `lint-staged.config.cjs`. I noticed prettier (`npm run format`) isn't formatting any of the `cjs` files, is this intentional?

It's an oversight on my behalf, definitely feel free to update it as part of this 😄

nsylke:
@Mythie Just adding `cjs`, or would you also like to have `js`, `mjs`, `cts`, `mts` and `mdx` as well to future proof it? Also, I just noticed we are formatting md where I think that was meant to be `mdx`.

Mythie:
> @Mythie Just adding `cjs`, or would you also like to have `js`, `mjs`, `cts`, `mts` and `mdx` as well to future proof it? Also, I just noticed we are formatting `md` where I think that was meant to be `mdx`.

Future proofing sounds great! I pray we never step into ESM hell but if we do it’ll be a lifesaver.

nsylke:
Hah, I definitely agree with that. Anyways, should be all good now!

Mythie (approved):

nsylke:
Resolved conflicts

nsylke:
I noticed when I pulled the parent to resolve the conflicts I used lockfile version 2 instead of 3. I just fixed that. I'm used to using pnpm, so didn't even realize a new version of the lockfile was released. 😓

Mythie (approved):

Mythie:
Thanks for your contribution Nicholas! We will now have some level of standard formatting on commits before they go up saving us some time 🎉 

thanks-pickle

PR #273, opened 2023-08-18
feat: feature flags

dguyen:
Added the following items:
- Feature flags
- Analytics and session recording

These are tools are provided by [PostHog](posthog.com/) and are disabled by default.


apps/web/src/components/forms/edit-document/add-fields.tsx
```diff
@@ -25,7 +25,7 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '@documenso/ui/primitive
 import { useToast } from '@documenso/ui/primitives/use-toast';
 
 import { PDF_VIEWER_PAGE_SELECTOR } from '~/components/(dashboard)/pdf-viewer/types';
-import { getBoundingClientRect } from '~/helpers/getBoundingClientRect';
+import { getBoundingClientRect } from '~/helpers/get-bounding-client-rect';
```
Mythie, line 28: thx

apps/web/src/pages/api/feature-flag/get.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const email = jwt?.email;
+  const userId = jwt?.id.toString();
+
+  let fallbackDistinctId = nanoid();
+
+  if (config) {
+    try {
+      const postHogCookie = JSON.parse(
+        request.cookies.get(`ph_${config.key}_posthog`)?.value ?? '',
```
Mythie, line 108: Should we use Zod here?

apps/web/src/providers/feature-flag.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import {
+  FEATURE_FLAG_POLL_INTERVAL,
+  LOCAL_FEATURE_FLAGS,
+  isFeatureFlagEnabled,
+} from '@documenso/lib/constants/feature-flags';
+
+import { getAllFlags } from '~/helpers/get-feature-flag';
+
+export type FeatureFlagValue = boolean | string | number | undefined;
```
Mythie, line 13: I noticed we have Zod for this elsewhere without being explicitly named, seems like we can tidy this into a `ZFeatureFlagValue` and `TFeatureFlagValue` combo for usage in the api route.

apps/web/src/providers/posthog.tsx
```diff
@@ -0,0 +1,52 @@
...
+    let url = window.origin + pathname;
+    if (searchParams && searchParams.toString()) {
+      url = url + `?${searchParams.toString()}`;
+    }
+    posthog.capture('$pageview', {
+      $current_url: url,
+    });
+  }, [pathname, searchParams, postHogConfig]);
+
+  return <></>;
```
Mythie, line 51: We can return null here instead of a fragment, not required bit a nitpick.

apps/web/src/pages/api/feature-flag/get.ts
dguyen: To handle parsing the cookie value? 

```
z.string().transform((value) => JSON.parse(value)).catch(() => {});
```

apps/web/src/providers/feature-flag.tsx
dguyen: Updated to use Zod, but I couldn't find other instances where it was being duplicated.

apps/web/src/providers/posthog.tsx
dguyen: Thanks, updated

apps/web/src/pages/api/feature-flag/get.ts
dguyen: I think it's a bit overkill

apps/web/src/helpers/get-feature-flag.ts
```diff
@@ -0,0 +1,79 @@
...
+  const response = await fetch(url, {
+    headers: {
+      ...requestHeaders,
+    },
+    next: {
+      revalidate: 60,
+    },
+  })
+    .then((res) => res.json())
+    .then((res) => z.union([z.boolean(), z.string()]).parse(res))
```
Mythie, line 36: This is where it was being duplicated, we can now use the dedicated Zod type you created

apps/web/src/helpers/get-feature-flag.ts
```diff
@@ -0,0 +1,79 @@
...
+  return fetch(url, {
+    headers: {
+      ...requestHeaders,
+    },
+    next: {
+      revalidate: 60,
+    },
+  })
+    .then((res) => res.json())
+    .then((res) => z.record(z.string(), z.union([z.boolean(), z.string()])).parse(res))
```
Mythie, line 68: And here

apps/web/src/pages/api/feature-flag/all.ts
```diff
@@ -0,0 +1,44 @@
...
+    return NextResponse.json(LOCAL_FEATURE_FLAGS);
+  }
+
+  const distinctId = extractDistinctUserId(token, nextReq);
+
+  const featureFlags = await postHog.getAllFlags(distinctId, mapJwtToFlagProperties(token));
+
+  const res = NextResponse.json(featureFlags);
+
+  res.headers.set('Cache-Control', 'public, s-maxage=120, stale-while-revalidate=60');
```
Mythie, line 41: I know we did this as a demo, for production use I think we should do `s-maxage=60, stale-while-revalidate=300` for a 1 minute cache and 5 minute revalidation.

apps/web/src/helpers/get-feature-flag.ts
dguyen: :face-palm:

thanks

apps/web/src/pages/api/feature-flag/all.ts
dguyen: thanks

Mythie (approved):

Issue #78, opened 2023-04-15
Cannot sign pdf with `’` in filename.


dephraiim:
PDF files containing `’` in their filenames throws an error when the user is about to sign.

### To Reproduce
- Take any pdf file and rename it but include an `’` in the name.
- Send the pdf to a user and try signing 
- The `/documents/[id]/sign` page will be stuck on the loading skeleton with the error displayed on the terminal

<details>
<summary><h3>Error</h3></summary>

```
TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["Content-Disposition"]
    at ServerResponse.setHeader (node:_http_outgoing:666:3)
    at _res.setHeader (/Users/duncan/documenso/node_modules/next/dist/server/base-server.js:129:24)
    at getHandler (webpack-internal:///(api)/./pages/api/documents/[id].ts:79:7)
    at async eval (webpack-internal:///(api)/../../packages/lib/server/defaultResponder.ts:12:22)
    at async eval (webpack-internal:///(api)/../../packages/lib/server/defaultHandler.ts:16:5)
    at async Object.apiResolver (/Users/duncan/documenso/node_modules/next/dist/server/api-utils/node.js:367:9)
    at async DevServer.runApi (/Users/duncan/documenso/node_modules/next/dist/server/next-server.js:474:9)
    at async Object.fn (/Users/duncan/documenso/node_modules/next/dist/server/next-server.js:736:37)
    at async Router.execute (/Users/duncan/documenso/node_modules/next/dist/server/router.js:252:36)
    at async DevServer.run (/Users/duncan/documenso/node_modules/next/dist/server/base-server.js:384:29) {
  code: 'ERR_INVALID_CHAR'
}                                                 
```
</details>

ElTimuro:
Noted, low priority for now. The easiest fix would be to disallow the upload of files with that name.

Jared87:
I also noticed this error with umlauts like ä, ö and ü.

mittalsam98:
Hey is this issue resolved?

ditorodev:
I would like to provide a fix for this @ElTimuro if you can point me where in the code this is happening I would add a dictionary that converts special ascii like `ä, ö, ü` and remove/replace special ascii characters from the name, im sure it is likely something on the parsing which is making it impossible to use these files if their name gets pre-processed before being uploaded/parsed by the backend then this issue can be fixed

Mythie:
[...]

PR #275, opened 2023-08-18
Implements URI encoding for file names

yashug:
Fixes https://github.com/documenso/documenso/issues/78

This commit adds encoding/decoding logic to ensure that file names are properly encoded before insertion into the database via prisma middleware


yashug:
@Mythie please have a look

Mythie:

PR #284, opened 2023-08-21
feat: added codeql and dependabot workflows

adithyaakrishna:
**Description:**

- This PR adds dependabot and codeql workflows
- Also, updates the readme as well :)
- Enabled `engines` in `package.json` to enforce Node v18.0.0 and above

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        language: [ 'javascript', 'typescript' ]
+        language: [ 'javascript' ]
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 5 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: true
+      matrix:
+        language: [ 'javascript', 'typescript' ]
```
nsylke, line 22:
```suggestion
        language: [ 'javascript' ]
```

The `typescript` is not a valid option in codeql. The `javascript` option will automatically run on both javascript and typescript files.

nsylke (changes requested):

.github/workflows/codeql-analysis.yml
adithyaakrishna: Yeah, makes sense. I added it by mistake. Thanks for pointing it out :)

Mythie (approved):

Mythie:
Thanks Adithya 🥳 

thanks-rainbow

PR #289, opened 2023-08-22
fix: fixed build command in codeql workflow

adithyaakrishna:
**Description:**

- The build fails for `codeql workflow` as there was a small mistake (Apologies for the same) in the build command


Mythie:
No stress, these things happen and I should have caught it 😅

Mythie (approved):

PR #290, opened 2023-08-22
feat: create robots.txt and sitemap.xml

nsylke:
With the latest versions of Next.js we can automatically generate robots & sitemap files. 

In the `robots.txt` we are allowing all paths except for `_next` since those files don't need to be crawled. 

In the `sitemap.xml` we are setting all the pages that have been explicitly created and then automatically pulling all generic and blog posts from contentlayer. The sitemap entries are alphabetized based on the directories inside of `app/`.


Mythie (approved):

Mythie:
Thanks Nicholas, loving your contributions 💚

PR #295, opened 2023-08-23
fix: use -p cli option for next dev

nsylke:
Use the built-in `-p` cli option in next for handling port. This is to support Windows, another option would be to use [`cross-env`](https://www.npmjs.com/package/cross-env).

### DIFF
--- a/apps/marketing/package.json
+++ b/apps/marketing/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "license": "AGPL-3.0",
   "scripts": {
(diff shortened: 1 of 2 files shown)


Mythie:
`cross-env` is probably nice to introduce, although I fear for windows users that aren't using WSL in general 😨

Mythie:
Thanks for the work Nicholas, let's add cross-env later on once this gets more complicated 🙌🏻

PR #296, opened 2023-08-24
feat: add inbox

dguyen:
Added an inbox page to display all incoming documents that require signing by the current user.


dguyen:
Notes:

- Emails are currently not hooked up to the signing token yet
- The date of the emails are not correct since the date the document was sent is currently unavailable

Mythie (approved):

Mythie:
Good work Lucas and David, especially Lucas you deserve a raise!

PR #300, opened 2023-08-24
feat: update document flow

dguyen:
Changes:

- Updated drag and drop fields
- Updated how height is calculated for PDFs and document flow
- Updated claim plan dialog
- Added button spinner
- Added animation for document flow slider
- Fixed z-index when dragging pre-existing fields
- Refactored document flow

Notes:

- Stacking issue still remains when dragging a new field
- Since the document flow now adjusts to the height of the PDF we may want to consider animating the height to reduce the jarring height change

### DIFF
--- a/packages/ui/primitives/document-flow/field-item.tsx
+++ b/packages/ui/primitives/document-flow/field-item.tsx
@@ -91,7 +91,7 @@ export const FieldItem = ({
   return createPortal(
     <Rnd
       key={coords.pageX + coords.pageY + coords.pageHeight + coords.pageWidth}
-      className={cn(' z-20', {
+      className={cn('z-20', {
         'pointer-events-none': passive,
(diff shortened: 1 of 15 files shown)


packages/ui/primitives/document-flow/field-item.tsx
```diff
@@ -91,9 +91,10 @@ export const FieldItem = ({
   return createPortal(
     <Rnd
       key={coords.pageX + coords.pageY + coords.pageHeight + coords.pageWidth}
-      className={cn('absolute z-20', {
+      className={cn(' z-20', {
```
Mythie, line 94: whitespace

dguyen: cheerio

Mythie:
blinding-lights

PR #307, opened 2023-08-26
chore: optimize images

adithyaakrishna:
**Description:**

The images under assets seem to occupy ~25 MB and these images can be compressed (in a lossless way) to reduce the size

- This PR saves ~8MB


Mythie:
This’ll come in handy, thanks! 🚀

Mythie (approved):

PR #309, opened 2023-08-26
feat: set min/max lengths and autocomplete for password

nsylke:
I adjusted the `minLength` to 6 characters since the password change already had it at 6 characters but the signup/signin had it at 1 character. Lets make it consistent across the platform! 

Additionally, I added the `maxLength` to 72 characters since bcrypt supports up 72 bytes. We're currently doing a silent truncation which isn't ideal. In the future, if people request longer passwords, we could look into prehashing the password using `bcrypt(hmac-sha256(salt, password))` to avoid hash shucking and allow passwords longer than 72 characters. Alternatively, we could look at other algorithms like `scrypt`, `pbkdf2` or even `argon2id`.

**Edit:** Added `autoComplete` based on @adithyaakrishna suggestion. This configures password managers to either autofill with the current password or generate a new password.

### DIFF
--- a/apps/web/src/components/forms/password.tsx
+++ b/apps/web/src/components/forms/password.tsx
@@ -18,8 +18,8 @@ import { FormErrorMessage } from '../form/form-error-message';
 
 export const ZPasswordFormSchema = z
   .object({
-    password: z.string().min(6),
-    repeatedPassword: z.string().min(6),
+    password: z.string().min(6).max(72),
+    repeatedPassword: z.string().min(6).max(72),
   })
   .refine((data) => data.password === data.repeatedPassword, {
     message: 'Passwords do not match',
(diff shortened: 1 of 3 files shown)


adithyaakrishna:
Min and Max lengths are good ideas, I also feel we should add a feature to know or find the strength of the passwords. Along with this, how abt we enable auto-complete?

nsylke:
@adithyaakrishna We'd likely want to use [zxcvbn](https://github.com/dropbox/zxcvbn) library to handle password strengths. I'll see what Lucas says before doing it though. I'll likely do it in a separate pull request.

adithyaakrishna:
Yes, that's a p good library. I guess, if I am not wrong, we use a version of that at Red Hat for some projects. They also have a REST API library

Mythie:
G'day, password strength is fun but it can cause users to pick passwords that are less secure since they're trying to match an arbitrary schema. The best passwords are long and contain decent bits of entropy:

image

Also TIL about the specific autocomplete values, thanks Nicholas! 🥇

Mythie (approved):

Mythie:
Thanks again Nicholas, I also always love an excuse to post an XKCD 😆

PR #312, opened 2023-08-27
fix: authentication allowing any password

G3root:
currently the login process allows any random password as a legit password, this PR aims to fix that.

### DIFF
--- a/packages/lib/next-auth/auth-options.ts
+++ b/packages/lib/next-auth/auth-options.ts
@@ -7,6 +7,7 @@ import GoogleProvider, { GoogleProfile } from 'next-auth/providers/google';
 import { prisma } from '@documenso/prisma';
 
 import { getUserByEmail } from '../server-only/user/get-user-by-email';
+import { ErrorCodes } from './error-codes';
 
 export const NEXT_AUTH_OPTIONS: AuthOptions = {
   adapter: PrismaAdapter(prisma),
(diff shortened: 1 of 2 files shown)


packages/lib/next-auth/error-codes.ts
```diff
@@ -0,0 +1,5 @@
+export const ErrorCodes = {
```
G3root, line 1: Error code to show an banner on the login form if an error occurs

packages/lib/next-auth/auth-options.ts
```diff
@@ -23,21 +24,23 @@ export const NEXT_AUTH_OPTIONS: AuthOptions = {
...
+        });
 
-        if (!user || !user.password) {
-          return null;
+        if (!user.password) {
+          throw new Error(ErrorCodes.UserMissingPassword);
         }
 
-        const isPasswordsSame = compare(password, user.password);
+        const isPasswordsSame = await compare(password, user.password);
```
G3root, line 40: this seems to be the issue

Mythie: I need to add an eslint rule for this 😓

adithyaakrishna: If we have a `no-floating-promises`rule

adithyaakrishna: We can combine it with `require-await` rule as well

Mythie (approved):

Mythie:
Thanks Nafees! We weren't able to get your first PR in but this one is great 🚀 

thanks-adventure-time

PR #314, opened 2023-08-27
fix: removed unnecessary whitespace before className

adithyaakrishna:
**Description:**

This PR removes unnecessary white space before or after the class names

Thought: How about we add an eslint rule, specifically this, https://eslint.org/docs/latest/rules/no-multi-spaces

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,9 +1,28 @@
 /** @type {import('eslint').Linter.Config} */
 module.exports = {
   root: true,
-  extends: ['@documenso/eslint-config'],
+  extends: [
+    '@documenso/eslint-config',
+    'plugin:package-json/recommended',
+    'plugin:import/recommended',
+  ],
+  plugins: ['package-json', 'import', 'simple-import-sort'],
   rules: {
     '@next/next/no-img-element': 'off',
+    'no-multi-spaces': [
+      'error',
+      {
+        ignoreEOLComments: false,
+        exceptions: {
+          BinaryExpression: false,
+          VariableDeclarator: false,
+          ImportDeclaration: false,
+          Property: false,
+        },
+      },
+    ],
+    'import/named': 'off',
+    'no-duplicate-imports': 'error',
   },
   settings: {
     next: {
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,28 +1,9 @@
 /** @type {import('eslint').Linter.Config} */
 module.exports = {
   root: true,
-  extends: [
-    '@documenso/eslint-config',
-    'plugin:package-json/recommended',
-    'plugin:import/recommended',
-  ],
-  plugins: ['package-json', 'import', 'simple-import-sort'],
+  extends: ['@documenso/eslint-config'],
   rules: {
     '@next/next/no-img-element': 'off',
-    'no-multi-spaces': [
-      'error',
-      {
-        ignoreEOLComments: false,
-        exceptions: {
-          BinaryExpression: false,
-          VariableDeclarator: false,
-          ImportDeclaration: false,
-          Property: false,
-        },
-      },
-    ],
-    'import/named': 'off',
-    'no-duplicate-imports': 'error',
   },
   settings: {
     next: {
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -15,7 +15,6 @@
     "typescript": "^5.1.6"
   },
   "devDependencies": {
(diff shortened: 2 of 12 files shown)


Mythie:
Should add the eslint rule as well so we never have to deal with this again 😎

adithyaakrishna:
@Mythie Done 🚀

adithyaakrishna:
Oops! Had made a mistake. Fixing it rn

.eslintrc.cjs
```diff
@@ -1,9 +1,28 @@
...
     '@next/next/no-img-element': 'off',
+    'no-multi-spaces': [
+      'error',
+      {
+        ignoreEOLComments: false,
+        exceptions: {
+          BinaryExpression: false,
+          VariableDeclarator: false,
+          ImportDeclaration: false,
+          Property: false,
```
Mythie, line 20: Are these not defaults?

.eslintrc.cjs
```diff
@@ -1,9 +1,28 @@
...
+        },
+      },
+    ],
+    'import/named': 'off',
+    'no-duplicate-imports': 'error',
```
Mythie, line 25: What's up with these two stragglers?

packages/eslint-config/package.json
```diff
@@ -9,9 +9,14 @@
...
     "eslint-config-next": "13.4.12",
     "eslint-config-prettier": "^8.8.0",
+    "eslint-config-turbo": "^1.9.3",
     "eslint-plugin-prettier": "^4.2.1",
     "eslint-plugin-react": "^7.32.2",
-    "eslint-config-turbo": "^1.9.3",
     "typescript": "^5.1.6"
+  },
+  "devDependencies": {
+    "eslint-plugin-import": "^2.28.1",
```
Mythie, line 18: This one is my enemy

packages/eslint-config/package.json
```diff
@@ -9,9 +9,14 @@
...
+    "eslint-plugin-package-json": "^0.1.4",
+    "eslint-plugin-simple-import-sort": "^10.0.0"
```
Mythie, line 20: We do this via prettier plugins now :D

.eslintrc.cjs
```diff
@@ -1,9 +1,28 @@
 /** @type {import('eslint').Linter.Config} */
 module.exports = {
   root: true,
-  extends: ['@documenso/eslint-config'],
+  extends: [
+    '@documenso/eslint-config',
+    'plugin:package-json/recommended',
```
Mythie, line 6: I like this plugin, we should add it to the core eslint-config package 🙌

.eslintrc.cjs
adithyaakrishna: Only `Property` is by default

packages/eslint-config/package.json
adithyaakrishna: IMHO it's a pretty good rule to have I feel, but like please lmk if I need to remove it 👀

.eslintrc.cjs
adithyaakrishna: That's the main mistake I made, I was adding it to the `eslint` file file by mistake 😂

.eslintrc.cjs
adithyaakrishna: 1. It is to verify if all the named imports are part of the set of named exports in the referenced module. More Info: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/named.md
2. This is to prevent duplicate imports, there was one under 'create.ts` file for formidable package

packages/eslint-config/package.json
adithyaakrishna: The `import-sort` or the `package-json` one? 👀

packages/eslint-config/package.json
Mythie: I'd remove it, I've spent years battling it and making it work across everything that I attempt to import and have come to the conclusion that it's not worth it. 

The things that the prettier plugin missed, typescript catches and so it's not worth the overhead.

I do appreciate the initiative though!

adithyaakrishna: Yeah, I can understand. That plugin is a real pain though xD

Removed it 👀

packages/eslint-config/package.json
adithyaakrishna: @Mythie Could you please let me know about this? So that I can wrap this PR up

Mythie (approved):

Mythie:
Thank you Adithya! 🔥 

thanks-a-bunch

PR #316, opened 2023-08-28
feat: added eslint rules to handle promises

adithyaakrishna:
**Description:**

- This PR adds `eslint` rules to handle promises
- This PR also fixes issues related to promises and their handling

/claim #315

PR #323, opened 2023-08-29
feat: promise safety

Mythie:
Updated based on #316 

This takes a more lax approach while still ensuring promise safety for most use cases.

### DIFF
--- a/apps/marketing/src/components/(marketing)/widget.tsx
+++ b/apps/marketing/src/components/(marketing)/widget.tsx
@@ -124,7 +124,7 @@ export const Widget = ({ className, children, ...props }: WidgetProps) => {
     setValue('signatureDataUrl', draftSignatureDataUrl);
     setValue('signatureText', '');
 
-    trigger('signatureDataUrl');
+    void trigger('signatureDataUrl');
     setShowSigningDialog(false);
   };
 
@@ -135,7 +135,9 @@ export const Widget = ({ className, children, ...props }: WidgetProps) => {
     signatureText,
   }: TWidgetFormSchema) => {
     try {
-      const delay = new Promise<void>((resolve) => setTimeout(resolve, 1000));
+      const delay = new Promise<void>((resolve) => {
+        setTimeout(resolve, 1000);
+      });
 
       // eslint-disable-next-line turbo/no-undeclared-env-vars
       const planId = process.env.NEXT_PUBLIC_STRIPE_COMMUNITY_PLAN_MONTHLY_PRICE_ID;
--- a/apps/web/src/components/(dashboard)/layout/profile-dropdown.tsx
+++ b/apps/web/src/components/(dashboard)/layout/profile-dropdown.tsx
@@ -118,7 +118,7 @@ export const ProfileDropdown = ({ user }: ProfileDropdownProps) => {
 
         <DropdownMenuItem
           onSelect={() =>
-            signOut({
+            void signOut({
               callbackUrl: '/',
             })
           }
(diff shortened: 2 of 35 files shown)


Mythie:
@adithyaakrishna To have a peep :)

apps/marketing/src/components/(marketing)/widget.tsx
```diff
@@ -124,7 +124,7 @@ export const Widget = ({ className, children, ...props }: WidgetProps) => {
     setValue('signatureDataUrl', draftSignatureDataUrl);
     setValue('signatureText', '');
 
-    trigger('signatureDataUrl');
+    void trigger('signatureDataUrl');
```
adithyaakrishna, line 127: If I am not wrong, this resolves to `undefined` right?

apps/web/src/components/(dashboard)/layout/profile-dropdown.tsx
```diff
@@ -118,7 +118,7 @@ export const ProfileDropdown = ({ user }: ProfileDropdownProps) => {
 
         <DropdownMenuItem
           onSelect={() =>
-            signOut({
+            void signOut({
               callbackUrl: '/',
             })
```
adithyaakrishna, line 123: I was curious when I made some changes to these files, Is there any specific reason we are handling the onSelect like this instead of the more traditional way of handleSubmit and creating a new function for it? or is it to prevent the infinite loop?

apps/marketing/src/components/(marketing)/widget.tsx
Mythie: Yep we’re void’ing promises that we’re happy to leave floating

apps/web/src/components/(dashboard)/layout/profile-dropdown.tsx
Mythie: For onselect specifically it’s to have commands on the item rather than normal handlers that have to route it

packages/tailwind-config/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "@documenso/tsconfig/base.json",
+  "compilerOptions": {
+    "allowJs": true,
+    "noEmit": true,
+  },
+  "include": ["**/*.cjs", "**/*.js"],
+  "exclude": ["dist", "build", "node_modules"]
```
adithyaakrishna, line 8: If we are extending from the tsconfig package, do we need to add the same excluded directories? 😶

Mythie: Not strictly required but preferred since we’re changing up the include we don’t want to confuse devs as to what’s being excluded

adithyaakrishna: This file and this, https://github.com/documenso/documenso/blob/ca40e983e3f35e8232477cb8e85061fcc35eff98/packages/eslint-config/tsconfig.json seems to be the exact same, so do we need to add the same complierOptions, include and exclude  keys?

adithyaakrishna:
Also a small nitpick, I saw that we are using regular functions to create some components under `.tsx` files, Is there any chance we can make them arrow functions, as from what I saw, most of the components use arrow functions itself

Mythie:
> Also a small nitpick, I saw that we are using regular functions to create some components under `.tsx` files, Is there any chance we can make them arrow functions, as from what I saw, most of the components use arrow functions itself

Where might we be doing that? I can’t think of any cases with it but maybe I missed something. The email templates and pages need to be functions for compiler reasons.

adithyaakrishna:
My bad, these were mostly on `page.tsx` files themselves 🤦🏻‍♂️

packages/tailwind-config/tsconfig.json
Mythie: I believe that's what I did, since I copy pasted the file 🙈

Mythie:
I'll let this sit overnight before I smack that merge button 

cat-button-explosion

PR #324, opened 2023-08-29
fix: prevent GitHub apis from being cached

adithyaakrishna:
**Description:**

The metrics on the `/open` page seem to be a bit off and I guess the GH API responses are being cached, if I am not wrong

**Current Stats:**

Screenshot 2023-08-29 at 11 39 04 AM

Screenshot 2023-08-29 at 11 39 09 AM

**Actual Stats:**

Screenshot 2023-08-29 at 11 36 58 AM

Screenshot 2023-08-29 at 11 37 07 AM

Edit: Had Pasted the same Screenshots :/

### DIFF
--- a/apps/marketing/src/app/(marketing)/open/page.tsx
+++ b/apps/marketing/src/app/(marketing)/open/page.tsx
@@ -31,14 +31,13 @@ const ZStargazersLiveResponse = z.record(
 
 export type StargazersType = z.infer<typeof ZStargazersLiveResponse>;
 
-// const ZOpenPullRequestsResponse = ZMergedPullRequestsResponse;
-
 export default async function OpenPage() {
+  const randomQueryParam = `?${Math.random()}`;
   const {
     forks_count: forksCount,
     open_issues: openIssues,
     stargazers_count: stargazersCount,
-  } = await fetch('https://api.github.com/repos/documenso/documenso', {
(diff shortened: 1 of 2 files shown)


apps/marketing/src/app/(marketing)/open/page.tsx
```diff
@@ -31,14 +31,13 @@ const ZStargazersLiveResponse = z.record(
...
-// const ZOpenPullRequestsResponse = ZMergedPullRequestsResponse;
-
 export default async function OpenPage() {
+  const randomQueryParam = `?${Math.random()}`;
   const {
     forks_count: forksCount,
     open_issues: openIssues,
     stargazers_count: stargazersCount,
-  } = await fetch('https://api.github.com/repos/documenso/documenso', {
+  } = await fetch(`https://api.github.com/repos/documenso/documenso${randomQueryParam}`, {
```
Mythie, line 40: You will notice above that we revalidate every hour meaning there will be a one hour delay between metrics. This is used for the card metrics at the top.

For the charts we use a separate API that polls GitHub every now and then to capture changes. The lag is fine for the most part since it's not super critical to have correct information at the second of the request.

adithyaakrishna: Ohhh, that makes sense, I had missed the revalidate param. I will go ahead and close this :)

PR #328, opened 2023-08-29
chore: change root package.json name

nsylke:
Maybe this was an oversight, or just opinionated, but with rest of apps and packages having `@documenso/[name]`, it makes sense to keep the root package.json consistent and name it `@documenso/root`.


Mythie:
lgtm

cat-button-explosion

Mythie (approved):

PR #331, opened 2023-08-30
fix: change signature color to white in dark mode

dephraiim:
Before
CleanShot 2023-08-30 at 03 53 02@2x

After

CleanShot 2023-08-30 at 03 52 39@2x


Mythie:
Closing as I accidentally redid this during some other theming work 😓

PR #333, opened 2023-08-30
fix: redirect signin page to dashboard when logged in

dephraiim:


apps/web/src/middleware.ts
```diff
@@ -1,25 +1,24 @@
...
-  //   console.log('token', token);
+  if (req.nextUrl.pathname.startsWith('/signin')) {
+    const token = await getToken({ req });
 
-  //   if (!token) {
-  //     console.log('has no token', req.url);
-  //     const redirectUrl = new URL('/signin', req.url);
-  //     redirectUrl.searchParams.set('callbackUrl', req.url);
+    if (token) {
+      console.log('has has token', req.url);
```
adithyaakrishna, line 16: Same here as well, do we need the `console.log`? 👀

Mythie: He learns from the best (me)

Mythie (approved):

PR #334, opened 2023-08-30
feat: redirect to dashboard when document is sent

dephraiim:
- Redirect to dashboard after document is sent
- Show a toast when document is sent

### DIFF
--- a/packages/trpc/server/document-router/router.ts
+++ b/packages/trpc/server/document-router/router.ts
@@ -18,6 +18,9 @@ export const documentRouter = router({
       try {
         const { documentId, recipients } = input;
 
+        console.log('documentId', documentId);
+        console.log('recipients', recipients);
+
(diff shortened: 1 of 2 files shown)


packages/trpc/server/document-router/router.ts
```diff
@@ -18,6 +18,9 @@ export const documentRouter = router({
       try {
         const { documentId, recipients } = input;
 
+        console.log('documentId', documentId);
+        console.log('recipients', recipients);
+
         return await setRecipientsForDocument({
```
adithyaakrishna, line 24: Do we need the `console.log` here? 👀

dephraiim: nope 👀

i'll take care of it

Mythie (approved):

Mythie:
Thanks Duncan!

PR #336, opened 2023-08-31
feat: add design system page

Mythie:
Adds a page for the design system that we have created in Figma. Can be accessed on the marketing site by navigating to `/design-system`

### DIFF
--- a/apps/marketing/content/design-system.mdx
+++ b/apps/marketing/content/design-system.mdx
@@ -1,0 +1,17 @@
+---
+title: Design System
+---
+
+# Design System
+
+At Documenso we aim to be a design driven company. We believe that design is not just about how things look, but also how they work. We want to make sure that our products are easy to use and intuitive. We also want to make sure that our products are consistent and that they look and feel like they belong together.
+
+To achieve this we have created our design system which contains our primitives, token and screens alongside branding assets.
+
+We're open-sourcing our design system so that you can see how we build our products and think about design as a whole.
+
+<iframe
+  src="https://documen.so/design-system-embed"
+  className="aspect-square w-full border-none"
+  frameBorder="0"
+/>
--- a/apps/marketing/content/design-system.mdx
+++ b/apps/marketing/content/design-system.mdx
@@ -2,13 +2,15 @@
 title: Design System
 ---
 
-# Design System
+# We're building a beautiful, open-source alternative to DocuSign
 
-At Documenso we aim to be a design driven company. We believe that design is not just about how things look, but also how they work. We want to make sure that our products are easy to use and intuitive. We also want to make sure that our products are consistent and that they look and feel like they belong together.
+At Documenso, we aim to be a design-driven company. 
 
-To achieve this we have created our design system which contains our primitives, token and screens alongside branding assets.
+We believe that design isn't just about how things look, but also how they work. We want to make sure that the product is easy to use and intuitive. We also want to ensure that the website, desktop and mobile apps are consistent and look and feel like they belong together.
 
-We're open-sourcing our design system so that you can see how we build our products and think about design as a whole.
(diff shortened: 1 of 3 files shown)


apps/marketing/content/design-system.mdx
```diff
@@ -0,0 +1,17 @@
...
+title: Design System
+---
+
+# Design System
+
+At Documenso we aim to be a design driven company. We believe that design is not just about how things look, but also how they work. We want to make sure that our products are easy to use and intuitive. We also want to make sure that our products are consistent and that they look and feel like they belong together.
+
+To achieve this we have created our design system which contains our primitives, token and screens alongside branding assets.
+
+We're open-sourcing our design system so that you can see how we build our products and think about design as a whole.
```
Mythie, line 11: @fmerian @ElTimuro Do your thang 💅🏻

adithyaakrishna:
These three orgs have been my source for inspiration wrt to design systems, 

- Airbnb, https://airbnb.design/
- Spotify, https://spotify.design/
- Atlassian, https://atlassian.design/

apps/marketing/content/design-system.mdx
dephraiim: Looks like its above your paygrade

Mythie: no doubt no doubt

apps/marketing/content/design-system.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+<iframe
+  src="https://documen.so/design-system-embed"
```
fmerian, line 14: what do we want to embed here? 

we plan to publish the design system to the Figma community so anyone can remix it. the current embed links to a read-only page. it'd be best to link to the Figma community file.

[See example](https://www.figma.com/community/file/1182079839084966604/ADS%3A-Design-Tokens)

apps/marketing/content/design-system.mdx
fmerian: `/design` may be a better (simpler) URI?

adithyaakrishna: Yes, I love the idea! 

Just imagine, https://design.documen.so, sounds powerful 😮‍💨

fmerian: for SEO, it's best to keep it under the main `documenso.com` domain, e.g. `documenso.com/design` (although [design.documenso.com](https://design.documenso.com/)) 🤯

apps/marketing/content/design-system.mdx
Mythie: Honestly an embed vs the figma community file is of very little difference. Jumping into a read-only will still let you create a local copy which is all the same.

This provides some interactivity on the page which is quite low-touch right now.

fmerian: > Honestly an embed vs the figma community file is of very little difference. Jumping into a read-only will still let you create a local copy which is all the same.

nope, there are *not so little* differences. 
1. when in read-only mode, you can just view the file. you don't access the data you need -- text style, spacings, measurements, and so on. 
2. you can't duplicate a Figma file or save it to your drafts. you have to invite people manually. plus, 
3. it doesn't leverage Figma as a distribution channel -- no likes, no comments, no search, no discovery.

these are significant differences. it's best to publish it to the Figma community and add a link here :)

ElTimuro: "when in read-only mode, you can just view the file. you don't access the data you need -- text style, spacings, measurements, and so on." => you can access those, just need to be logged in, I checkd

ElTimuro: - nur sure about 2, 3 is a thing
- Why wouldn't we publish after embed? Leverage both

fmerian: > Why wouldn't we publish after embed?
compound effect

PR #346, opened 2023-09-04
chore: removed console logs and warn

catalinpit:
Removed unneeded console.logs and warn.


adithyaakrishna:
I feel we should add a eslint rule to warn about `console.log()` and wherever its needed, we just use eslint-ignore

Mythie (approved):

Mythie:
> I feel we should add a eslint rule to warn about `console.log()` and wherever its needed, we just use eslint-ignore

It'll never work because I exist 😆

Mythie:
Thanks for removing my tactical logs, most of the things I was watching have been resolved a while back so it's nice to remove this 🙌

PR #347, opened 2023-09-05
feat: limit document upload size

G3root:


Mythie:
We certainly should have limits, although I'll need to consider how big is too big. Obviously we don't want gigabyte sized documents but we should certainly entertain something reasonable.

@ElTimuro you have any thoughts on the matter?

ElTimuro:
There should a limited stopping attackes or damaging actions but not much more at this point IMO. 100MB? 200? I've seen 50MB PDF Scans and Im sure there are bigger ones..

Mythie:
> There should a limited stopping attackes or damaging actions but not much more at this point IMO. 100MB? 200? I've seen 50MB PDF Scans and Im sure there are bigger ones..

Yeah that's my line of thinking, high dpi scans or figma exports easily enter the 100's of MB's

G3root:
From what I found online, DocuSign has an upload limit of either 25MB or 50MB. I think we should be okay with those numbers.

Mythie (approved):

PR #350, opened 2023-09-05
feat: redirect signed document to completed page

dephraiim:


apps/web/src/app/(signing)/sign/[token]/page.tsx
```diff
@@ -44,6 +44,10 @@ export default async function SigningPage({ params: { token } }: SigningPageProp
     return notFound();
   }
 
+  if (document?.status === 'COMPLETED') {
```
Comment (Mythie) [inline apps/web/src/app/(signing)/sign/[token]/page.tsx:47]: You should be able to do `document.status` rather than `document?.status` since we perform our undefined check above 😄

Comment (dephraiim) [reply apps/web/src/app/(signing)/sign/[token]/page.tsx]: I fixed it

apps/web/src/app/(signing)/sign/[token]/page.tsx
```diff
@@ -44,6 +44,10 @@ export default async function SigningPage({ params: { token } }: SigningPageProp
     return notFound();
   }
 
+  if (document.status === 'COMPLETED') {
```
Comment (Mythie) [inline apps/web/src/app/(signing)/sign/[token]/page.tsx:47]: `if (document.status === DocumentStatus.COMPLETED || recipient.signStatus === SignStatus.SIGNED)`

apps/web/src/app/(signing)/sign/[token]/page.tsx
```diff
@@ -53,6 +53,13 @@ export default async function SigningPage({ params: { token } }: SigningPageProp
 
   const user = await getServerComponentSession();
 
+  if (
+    document.status === DocumentStatus.COMPLETED ||
+    recipient.signingStatus === SigningStatus.SIGNED
+  ) {
+    redirect(`/sign/${token}/complete`);
+  }
+
```
Comment (dguyen) [inline apps/web/src/app/(signing)/sign/[token]/page.tsx:62]: I think this should go directly after 

```
  if (!document || !document.documentData || !recipient) {
    return notFound();
  }
```

Since it doesn't need to rely on the two sequential awaits that aren't relevant to the statement

PR #359, opened 2023-09-07
feat: send email when recipient is done signing and every recipient is done signing

dephraiim:
- Send email when a recipient has signed
- Send email once document has been sealed

### Summary by CodeRabbit

- New Feature: Added email notifications for pending and completed documents. When a document is sealed, an email notification is sent to all recipients. If there are recipients who have not yet signed the document, a pending email notification is sent to them.
- Refactor: Enhanced the `sealDocument` and `completeDocumentWithToken` functions to include the new email notification feature.


packages/lib/server-only/document/complete-document-with-token.ts
```diff
@@ -69,6 +70,16 @@ export const completeDocumentWithToken = async ({
     },
   });
 
+  const numberOfRecipients = await prisma.recipient.count({
+    where: {
+      documentId: document.id,
+    },
+  });
+
```
Mythie, line 78: I think this should filter on those that haven't signed and make sure it's `> 1` otherwise the final recipient will get a double email of pending + completed.

PR #360, opened 2023-09-08
feat: add sorting for documents table

G3root:
[Screencast 2023-09-08 07:19:16.webm](https://github.com/documenso/documenso/assets/84864519/d00cb22d-2ea6-4bd5-948c-aeb1c9b3061e)


Mythie:
Appreciate the effort but I'd like to take a step back and consider sorting with our designer! Likewise we will probably avoid in memory sorting and do it server side 🫶🏻

PR #361, opened 2023-09-08
feat: admin ui for metrics

catalinpit:


Mythie:

https://documen.so/cla

apps/web/src/components/(dashboard)/layout/profile-dropdown.tsx
```diff
@@ -35,8 +37,8 @@ export type ProfileDropdownProps = {
 
 export const ProfileDropdown = ({ user }: ProfileDropdownProps) => {
   const { theme, setTheme } = useTheme();
-
   const { getFlag } = useFeatureFlags();
+  const userIsAdmin = isAdmin(user);
```
Mythie, line 41: Silly rule, but I'd be consistent with the below flag using the `is<resource>` naming scheme so `userIsAdmin => isUserAdmin` 😄

packages/lib/index.ts
```diff
@@ -1 +1,5 @@
-export {};
+import { Role, User } from '@documenso/prisma/client';
```
Mythie, line 1: I'd consider moving this to something like `universal/guards/is-admin.ts` I've been avoiding barrel files since they can have some gnarly outcomes during bundling 😸

apps/web/src/app/(dashboard)/admin/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+    status: InternalReadStatus.NOT_OPENED,
+  },
+  {
+    icon: Send,
+    title: 'Total recipients with sent count',
+    status: InternalSendStatus.SENT,
+  },
+  {
+    icon: Archive,
+    title: 'Total recipients with unsent count',
```
catalinpit, line 65: these titles need a better name

apps/web/src/app/(dashboard)/admin/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+    | 'TOTAL_RECIPIENTS'
+    | 'OPENED'
+    | 'NOT_OPENED'
+    | 'SIGNED'
+    | 'NOT_SIGNED'
+    | 'SENT'
+    | 'NOT_SENT';
+}[];
+
+const CARD_DATA: TCardData = [
```
Mythie, line 42: For this I'd consider dropping `[]` from `TCardData` declaration and instead adding it to the constant declaration.

e.g. `CARD_DATA: TCardData[] = [...]`

apps/web/src/app/(dashboard)/admin/page.tsx
Mythie: @ElTimuro to the rescue perhaps 🤔

packages/lib/server-only/admin/get-recipients.ts
```diff
@@ -0,0 +1,20 @@
...
+  return {
+    TOTAL_RECIPIENTS: results.length,
+    [ReadStatus.OPENED]: results.filter((r) => r.readStatus === 'OPENED')?.[0]?._count ?? 0,
+    [ReadStatus.NOT_OPENED]: results.filter((r) => r.readStatus === 'NOT_OPENED')?.[0]?._count ?? 0,
+    [SigningStatus.SIGNED]: results.filter((r) => r.signingStatus === 'SIGNED')?.[0]?._count ?? 0,
+    [SigningStatus.NOT_SIGNED]:
+      results.filter((r) => r.signingStatus === 'NOT_SIGNED')?.[0]?._count ?? 0,
+    [SendStatus.SENT]: results.filter((r) => r.sendStatus === 'SENT')?.[0]?._count ?? 0,
+    [SendStatus.NOT_SENT]: results.filter((r) => r.sendStatus === 'NOT_SENT')?.[0]?._count ?? 0,
+  };
```
Mythie, line 19: I needed some time to look at this one, you can simplify this with the following 😸

```typescript
export const getRecipientsStats = async () => {
  const results = await prisma.recipient.groupBy({
    by: ['readStatus', 'signingStatus', 'sendStatus'],
    _count: true,
  });

  const stats = {
    TOTAL_RECIPIENTS: 0,
    [ReadStatus.OPENED]: 0,
    [ReadStatus.NOT_OPENED]: 0,
    [SigningStatus.SIGNED]: 0,
    [SigningStatus.NOT_SIGNED]: 0,
    [SendStatus.SENT]: 0,
    [SendStatus.NOT_SENT]: 0,
  };

  results.forEach((result) => {
    const { readStatus, signingStatus, sendStatus, _count } = result;

    stats[readStatus] += _count;
    stats[signingStatus] += _count;
    stats[sendStatus] += _count;

    stats.TOTAL_RECIPIENTS += _count;
  });

  return stats;
};
```

packages/lib/server-only/admin/get-recipients.ts
```diff
@@ -0,0 +1,20 @@
+import { prisma } from '@documenso/prisma';
```
Mythie, line 1: I'd also change the filename to `get-recipient-stats` and align the method name to that 🙌

packages/prisma/schema.prisma
```diff
@@ -21,6 +26,7 @@ model User {
   password         String?
   source           String?
   signature        String?
+  roles            Role[]           @default([USER])
```
Mythie, line 29: Excellent 🤩

packages/lib/server-only/admin/get-recipients.ts
catalinpit: Much better, thanks!

packages/lib/server-only/admin/get-users-stats.ts
dephraiim, line 1: This file is only exporting the number of users with subscription.

Isn't `get-users-stats` too generic?

packages/prisma/schema.prisma
dephraiim, line 29: Isn't `role` better than `roles`?

dephraiim: I have a question, @catalinpit  @Mythie , why is it  `@default([USER])` with the brackets and not ` @default(USER)`

packages/lib/server-only/admin/get-users-stats.ts
catalinpit: it also exports the total number of users in the db. In the future, it might export more stuff. but feel free to suggest another name.

packages/prisma/schema.prisma
catalinpit: since a user can have multiple roles, I thought `roles` is more appropriate.

catalinpit: @dephraiim Prisma schema expects array field to have array values. So, for the Role[] field, the default must be an array, not a scalar, afaik.

Mythie: Yeah I think the idea of allowing for multiple role memberships in the future might be worthwhile. Admittedly we could make the change later but this saves headache 😄

turbo.json
```diff
@@ -2,23 +2,16 @@
...
   "pipeline": {
     "build": {
-      "dependsOn": [
-        "^build"
-      ],
-      "outputs": [
-        ".next/**",
-        "!.next/cache/**"
-      ]
+      "dependsOn": ["^build"],
```
catalinpit, line 5: I didn't change the style myself, I don't know why this happened.

packages/prisma/schema.prisma
dephraiim: Yeahh

dephraiim: For multiple roles, I understand

apps/web/src/components/(dashboard)/layout/profile-dropdown.tsx
```diff
@@ -67,6 +69,15 @@ export const ProfileDropdown = ({ user }: ProfileDropdownProps) => {
           </Link>
         </DropdownMenuItem>
 
+        {isUserAdmin && (
+          <DropdownMenuItem asChild>
+            <Link href="/admin" className="cursor-pointer">
```
dguyen, line 74: Aren't links by default already applied with the "cursor-pointer" style?

turbo.json
```diff
@@ -48,6 +41,7 @@
     "NEXT_PRIVATE_SMTP_SECURE",
     "NEXT_PRIVATE_SMTP_FROM_NAME",
     "NEXT_PRIVATE_SMTP_FROM_ADDRESS",
-    "NEXT_PRIVATE_STRIPE_API_KEY"
+    "NEXT_PRIVATE_STRIPE_API_KEY",
+    "APP_VERSION"
```
dguyen, line 45: I think placing this at the top would align with the somewhat ordered list

apps/web/src/app/(dashboard)/admin/page.tsx
```diff
@@ -0,0 +1,114 @@
...
+} from '@documenso/lib/server-only/admin/get-users-stats';
+import {
+  ReadStatus as InternalReadStatus,
+  SendStatus as InternalSendStatus,
+  SigningStatus as InternalSigningStatus,
+} from '@documenso/prisma/client';
+
+import { CardMetric } from '~/components/(dashboard)/metric-card/metric-card';
+
+type TCardData = {
```
dguyen, line 29: I think the intention of having T prefixed on a type is for inferred ZOD schemas, normal types would normally go without them

@Mythie pls correct me if wrong

apps/web/src/app/(dashboard)/admin/page.tsx
```diff
@@ -0,0 +1,114 @@
+import {
+  Archive,
+  File,
+  FileX2,
+  LucideIcon,
+  User as LucideUser,
```
dguyen, line 6: I think it would be better named as `User as UserIcon`

iirc lucide should be exporting `UserIcon` anyway but I assume we're doing something weird so it doesn't work or something

dguyen (changes requested):

apps/web/src/components/(dashboard)/layout/profile-dropdown.tsx
catalinpit: followed the style from the existing code. all the links have that className

Mythie: Dropdown items lack cursor styling, this is intentional :)

Mythie (approved):

Mythie:
Happy first PR @catalinpit we're glad to have you on the team 🙌 

thanks-flower-dog

apps/web/src/app/(dashboard)/admin/page.tsx
ElTimuro: Until we know more about how the metric is used, I would keep it technical, "close to the metal"

PR #365, opened 2023-09-11
feat: add avatar email fallback

dguyen:
Added a fallback to avatars so that when no name is provided, it will use the first letter of the email if possible.

Examples:
- Recipient John Doe (foo@bar.com) will have the fallback text be 'JD'
- Recipient with no name (foo@bar.com) will have the fallback text be 'F'

### DIFF
--- a/packages/lib/client-only/recipient-avatar-fallback.ts
+++ b/packages/lib/client-only/recipient-avatar-fallback.ts
@@ -1,0 +1,7 @@
+import { Recipient } from '@documenso/prisma/client';
+
+import { initials } from './recipient-initials';
+
+export const recipientAvatarFallback = (recipient: Recipient) => {
+  return initials(recipient.name) || recipient.email.slice(0, 1).toUpperCase();
+};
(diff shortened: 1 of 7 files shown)


packages/lib/client-only/recipient-initials.ts
```diff
@@ -3,4 +3,4 @@ export const initials = (text: string) =>
     ?.split(' ')
```
Mythie, line 3: this is more of a string util, do we have a file for that?

packages/lib/client-only/recipient-avatar-fallback.ts
```diff
@@ -0,0 +1,7 @@
+import { Recipient } from '@documenso/prisma/client';
+
+import { initials } from './recipient-initials';
+
+export const recipientAvatarFallback = (recipient: Recipient) => {
```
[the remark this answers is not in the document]
dguyen: Will probably refactor as 'recipientAbbreviation' since it's not actually their initials

packages/lib/client-only/recipient-initials.ts
dguyen: Doesn't look like it, what do you think about having a file `packages/lib/universal/recipient-utils.ts` where we dump in any logic related to recipients in?

dguyen: Decided on using `packages/lib/utils/recipient-formatter.ts` lmk if you think there's a better place

PR #371, opened 2023-09-12
feat: update document table layout

dguyen:
Changes:

- Removed dashboard page
- Removed ID column (in my opinion it is redundant from the user perspective)
- Added additional display options for pagination where the "rows selected" is located
- Moved date to first column
- Use the estimated locale from the user headers to format dates on SSR


apps/web/src/components/forms/signin.tsx
```diff
@@ -54,6 +55,8 @@ export const SignInForm = ({ className }: SignInFormProps) => {
 
   const errorCode = searchParams?.get('error');
 
+  const loginRedirectPath = '/documents';
```
Mythie, line 58: What's the point of this one?

Mythie: Oh I see, move it to the top and make it shouty. This should probably be a real constant elsewhere in hindsight.

dguyen: I was thinking to have it here since in the near future we could check for redirects (eg https://localhost:3000/signin?redirectTo="/something") in the query param and attach it here

Mythie: Fair fair, I'd still say top of the file and shouty since it doesn't need to be reinstantiated when the component rerenders 😄

Mythie (approved):

Mythie:
lgtm

cat-button-explosion

PR #376, opened 2023-09-14
feat: add-playwright

catalinpit:


Mythie:
Love it! Should we have the Playwright config at the root of the repo or per "app" 🤔 

I was gonna ask if we could add a GitHub action for it but you beat me to the punch as well 💪🏻

catalinpit:
> Love it! Should we have the Playwright config at the root of the repo or per "app" 🤔
> 
> I was gonna ask if we could add a GitHub action for it but you beat me to the punch as well 💪🏻

Good question @Mythie! If we are going to have multiple apps (except the marketing app), it will make more sense to have it per "app".

apps/web/src/tests/e2e/auth.setup.ts
```diff
@@ -0,0 +1,34 @@
...
+  await page.getByRole('button', { name: 'Sign Up' }).click();
+
+  await page.goto('/');
+  await page.getByLabel('Email').fill(email);
+  await page.getByLabel('Password').fill(password);
+  await page.getByRole('button', { name: 'Sign In' }).click();
+  await page.waitForURL('/documents');
+  await expect(page).toHaveURL('/documents');
+  await page.context().storageState({ path: STORAGE_STATE });
+});
```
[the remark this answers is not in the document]
catalinpit: This is the setup for the tests that require authentication (e.g. upload and send a document for signing). So, before running those tests, Playwright will register a new user and log in with the newly created user.

Playwright enables you to create a global authentication state that can be shared by all the tests.

https://playwright.dev/docs/auth#basic-shared-account-in-all-tests

.env.example
```diff
@@ -15,6 +15,11 @@ NEXT_PRIVATE_DATABASE_URL="postgres://documenso:password@127.0.0.1:54320/documen
 # Defines the URL to use for the database when running migrations and other commands that won't work with a connection pool.
 NEXT_PRIVATE_DIRECT_DATABASE_URL="postgres://documenso:password@127.0.0.1:54320/documenso"
 
+# [[E2E Tests]]
+E2E_TEST_USERNAME=""
+E2E_TEST_USER_EMAIL=""
+E2E_TEST_USER_PASSWORD=""
```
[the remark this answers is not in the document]
catalinpit: Yeah, I'll add them.

apps/web/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
```
Mythie, line 27: Assume this is broken now since we're doing things in `apps/web`, I'd recommend making this use turbo to run it in any applicable dirs.

That way we can have a package script like `test:e2e`.

Additionally I think you'll need to bootstrap some devservers, consider using [act](https://github.com/nektos/act) to test locally 😄

turbo.json
```diff
@@ -17,13 +12,26 @@
     "dev": {
       "cache": false,
       "persistent": true
+    },
+    "e2e:test": {
+      "dependsOn": ["^build"]
```
Mythie, line 17: Smart

apps/web/package.json
```diff
@@ -41,9 +43,11 @@
     "sharp": "0.32.5",
     "ts-pattern": "^5.0.5",
     "typescript": "5.1.6",
-    "zod": "^3.21.4"
+    "zod": "^3.21.4",
+    "start-server-and-test": "^2.0.1"
```
Mythie, line 47: This one should be below `sharp`

apps/web/package.json
```diff
@@ -8,6 +8,8 @@
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
+    "dev:test": "playwright test",
+    "e2e:test": "start-server-and-test \"npm run start\" http://localhost:3000 \"playwright test\"",
```
Mythie, line 12: I don't mind `dev:test` but for the other I think we'd want `test:e2e` so you could then have `test:unit` and `test:integration` while `test` would run all three 😄

apps/web/.env.example
```diff
@@ -0,0 +1,45 @@
+# [[E2E Tests]]
```
Mythie, line 1: Not sure if we want this here, didn't notice the mail stuff there either. If it already existed it was likely a mistake from myself 😓

package.json
```diff
@@ -4,10 +4,12 @@
     "build": "turbo run build",
     "dev": "turbo run dev --filter=@documenso/web --filter=@documenso/marketing",
     "start": "cd apps && cd web && next start",
+    "test": "cd apps && cd web && npm run test:e2e",
```
dephraiim, line 7: Why not `turbo run test` to run the test script in `apps/web`?

packages/lib/server-only/user/delete-user-and-data.ts
```diff
@@ -0,0 +1,41 @@
...
+    select: {
+      documentData: {
+        select: {
+          data: true,
+        },
+      },
+    },
+  });
+
+  return Promise.all([
```
dephraiim, line 29: Prisma Transactions will work best here, what do you think?

apps/web/src/tests/e2e/test-documents-page.authenticated.spec.ts
```diff
@@ -0,0 +1,38 @@
+import path from 'path';
+
+import { expect, test } from '../test-fixtures/documents-page/documents-page';
+
+const signer_name = process.env.E2E_TEST_SIGNER_NAME;
```
dephraiim, line 5: For consistency, would you consider using camelCase for the variable names?

PR #387, opened 2023-09-19
feat: security headers

nsylke:
Implement security headers. As far as I can tell, we not using any APIs that require us to specify it in the headers.


Mythie (approved):
😍

Mythie:
Thanks Nick!

Always nice to see your work 🫶🏻

PR #396, opened 2023-09-21
feat: admin UI for managing Documenso instance

catalinpit:
## Description

I created the first version of the admin interface that allows you to manage your Documenso instance. The interface is quite simple currently, but new functionalities will be added.

## Changes made

At the moment, it enables you to:
1) See your instance stats

A screenshot of the stats page from the admin dashboard

2) Manage the users 

A screenshot of the users page from the admin dashboard

3) Edit an individual user

A screenshot of the individual user page from the admin dashboard

4) List all subscriptions

A screenshot of the subscriptions page from the admin dashboard

## Tests performed

- I checked all the pages against the database to make sure the data matches
- I edited the users through the admin interface and made sure the changes are reflected in the database


apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx
```diff
@@ -0,0 +1,8 @@
+export default function UserPage() {
+  return (
+    <div>
+      <h1>Hey</h1>
+      <h2>Ho</h2>
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx:5]: Hey ho to you too!

apps/web/src/app/(dashboard)/admin/users/data-table-users.tsx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+interface Subscription {
+  id: number;
+  status: string;
+  planId: string | null;
+  priceId: string | null;
+  createdAt: Date | null;
+  periodEnd: Date | null;
+}
```
Mythie, line 30: Any reason to not nab these from Prisma?

apps/web/src/app/(dashboard)/admin/users/data-table-users.tsx
```diff
@@ -0,0 +1,135 @@
...
+          },
+          {
+            header: 'Subscription status',
+            accessorKey: 'subscription',
+            cell: ({ row }) => {
+              return (
+                <>
+                  {row.original.Subscription.map((subscription: Subscription, idx: number) => {
+                    return <span key={idx}>{subscription.status}</span>;
+                  })}
```
Mythie, line 97: Typically opt for the long form `index` or `i` 😄

packages/lib/server-only/user/get-all-users.ts
```diff
@@ -0,0 +1,37 @@
...
+        roles: true,
+        Subscription: {
+          select: {
+            id: true,
+            status: true,
+            planId: true,
+            priceId: true,
+            createdAt: true,
+            periodEnd: true,
+          },
```
Mythie, line 24: Looks like an `include: { Subscription: true } }` is more suitable here 😸 

Great idea with limiting the size of the query though by omitting columns we don't care about.

Mythie (approved):

packages/lib/server-only/user/get-all-users.ts
catalinpit: If I do that, wouldn't it pull all the fields? In this case, I omit `customerId`, `userId`, and `updatedAt`.

apps/web/src/app/(dashboard)/admin/documents/data-table-action-button.tsx
```diff
@@ -0,0 +1,30 @@
...
+export type DataTableActionButtonProps = {
+  row: Pick<Document, 'id'>;
+};
+
+export const DataTableActionButton = ({ row }: DataTableActionButtonProps) => {
+  const { data: session } = useSession();
+
+  if (!session) {
+    return null;
+  }
```
Mythie, line 20: This seems redundant no?

apps/web/src/app/(dashboard)/admin/documents/data-table.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    </AvatarFallback>
+                  </Avatar>
+                </Link>
+              );
+            },
+          },
+          {
+            header: 'Status',
+            accessorKey: 'status',
+            cell: ({ row }) => <DocumentStatus status={row.getValue('status')} />,
```
Mythie, line 93: We typically prefer `row.original` since it maintains type safety

apps/web/src/app/(dashboard)/admin/documents/data-table-action-button.tsx
catalinpit: 🤦‍♂️ I thought I removed the component altogether since it only renders the button. Thanks for the catch!

apps/web/src/app/(dashboard)/admin/documents/data-table-action-dropdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    const link = window.document.createElement('a');
+
+    link.href = window.URL.createObjectURL(blob);
+    link.download = row.title || 'document.pdf';
+
+    link.click();
+
+    window.URL.revokeObjectURL(link.href);
+  };
```
Mythie, line 57: @dguyen Have we extracted this somewhere already

apps/web/src/app/(dashboard)/admin/documents/data-table.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const updateSearchParams = useUpdateSearchParams();
+
+  const onPaginationChange = (page: number, perPage: number) => {
+    startTransition(() => {
+      updateSearchParams({
+        page,
+        perPage,
+      });
+    });
+  };
```
Mythie, line 42: Perfect!

apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { Combobox } from '@documenso/ui/primitives/combobox';
+import { Input } from '@documenso/ui/primitives/input';
+import { Label } from '@documenso/ui/primitives/label';
+import { useToast } from '@documenso/ui/primitives/use-toast';
+
+import { FormErrorMessage } from '../../../../../components/form/form-error-message';
+import {
+  TUserFormSchema,
+  ZUserFormSchema,
+} from '../../../../../providers/admin-user-profile-update.types';
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx:20]: Dang should this be under a provider?

apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx
```diff
@@ -0,0 +1,126 @@
...
+  );
+
+  const user = result.data;
+
+  const roles = user?.roles;
+  let rolesArr: string[] = [];
+
+  if (roles) {
+    rolesArr = Object.values(roles);
+  }
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx:42]: Isn't roles always going to be an array of enums for this? Theoretically we can just have it be 
`const roles = user?.roles ?? []`

apps/web/src/app/(dashboard)/admin/users/data-table-users.tsx
```diff
@@ -0,0 +1,193 @@
...
+            updateSearchParams({
+              page: 1,
+              perPage,
+            });
+          });
+        }
+      } catch (err) {
+        throw new Error(err);
+      }
+    };
```
Mythie, line 78: Typically this would be handled in the server component one level above e.g. `admin/users/page.tsx` so within the data table we're just manipulating the search params which will cause an update.

packages/lib/server-only/admin/get-all-documents.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const findDocuments = async ({ term, page = 1, perPage = 10 }: FindDocumentsOptions) => {
+  const termFilters = !term
+    ? undefined
+    : ({
+        title: {
+          contains: term,
+          mode: 'insensitive',
+        },
+      } as const);
```
Mythie, line 17: I'm so sorry

packages/lib/server-only/user/get-all-users.ts
Mythie: Yeah while beneficial those fields aren't _large_ enough to justify worrying about it. Happy to keep it for now but a rule of thumb is that unless it's adding a bunch of data to be sent across the wire we shouldn't worry about including all columns.

Simplicity => Perf

packages/lib/server-only/user/get-all-users.ts
```diff
@@ -0,0 +1,71 @@
...
+    }),
+    await prisma.user.count({
+      where: whereClause,
+    }),
+  ]);
+
+  return {
+    users,
+    totalPages: Math.ceil(count / perPage),
+  };
```
Mythie, line 70: Should this have the whole page, perPage, totalPages etc going on? I should write a util function for returning that type tbh

packages/trpc/server/admin-router/router.ts
```diff
@@ -0,0 +1,33 @@
+import { TRPCError } from '@trpc/server';
+
+import { isAdmin } from '@documenso/lib/next-auth/guards/is-admin';
+import { updateUser } from '@documenso/lib/server-only/admin/update-user';
+
+import { authenticatedProcedure, router } from '../trpc';
+import { ZUpdateProfileMutationByAdminSchema } from './schema';
+
+export const adminRouter = router({
+  updateUser: authenticatedProcedure
```
Mythie, line 10: We can keep this for now but should consider add an `adminProcedure` in the future to capture these issues.

apps/web/src/app/(dashboard)/admin/documents/data-table-action-dropdown.tsx
dguyen: Kind of, but it's a component `document-download-button.tsx` 

We can extract the logic to a hook but not sure if it's worth it.

apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <FormErrorMessage className="mt-1.5" error={errors.roles} />
+        </div>
+
+        <div className="mt-4">
+          <Button type="submit" disabled={isSubmitting}>
+            {isSubmitting && <Loader className="mr-2 h-5 w-5 animate-spin" />}
+            Update user
+          </Button>
+        </div>
+      </form>
```
Comment (dguyen) [inline apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx:123]: You may want to consider trying a newer form implementation we introduced a while ago:

https://github.com/documenso/documenso/blob/feat/refresh/packages/ui/primitives/document-flow/add-signature.tsx#L209-L294

Handles styling, errors, disabling form, etc

Just a suggestion

apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx
```diff
@@ -0,0 +1,126 @@
...
+            name="roles"
+            render={({ field: { onChange } }) => (
+              <Combobox listValues={rolesArr} onChange={(values: string[]) => onChange(values)} />
+            )}
+          />
+          <FormErrorMessage className="mt-1.5" error={errors.roles} />
+        </div>
+
+        <div className="mt-4">
+          <Button type="submit" disabled={isSubmitting}>
```
Comment (dguyen) [inline apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx:118]: Button has a `loading` prop that shows a spinner

packages/lib/server-only/admin/get-all-documents.ts
dguyen: Should be able to do this instead

```
  const termFilters: Prisma.DocumentWhereInput | undefined = !term
    ? undefined
    : {
        title: {
          contains: term,
          mode: 'insensitive',
        },
      };
```

packages/lib/server-only/admin/update-user.ts
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export const updateUser = async ({ id, name, email, roles }: UpdateUserOptions) => {
+  await prisma.user.findFirstOrThrow({
+    where: {
+      id,
+    },
+  });
+
+  const updatedUser = await prisma.user.update({
```
dguyen, line 18: Can just return this directly

packages/lib/server-only/user/get-all-users.ts
```diff
@@ -0,0 +1,71 @@
+import { Prisma } from '@prisma/client';
+
+import { prisma } from '@documenso/prisma';
+
+type getAllUsersProps = {
```
dguyen, line 5: Is there a reason this is camel case instead of pascal case?

packages/trpc/server/admin-router/router.ts
dguyen: I think now would be better than later since it looks like we're already doing this check more than once.

packages/ui/primitives/combobox.tsx
```diff
@@ -0,0 +1,84 @@
...
+      newSelectedValues = [...selectedValues, currentValue];
+    }
+
+    setSelectedValues(newSelectedValues);
+    onChange(newSelectedValues);
+    setOpen(false);
+  };
+
+  return (
+    <>
```
dguyen, line 47: This outer fragment seems redundant?

apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx
Comment (catalinpit) [reply apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx:42]: True, I just gave it a try. Thanks for the catch.

apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx
Comment (catalinpit) [reply apps/web/src/app/(dashboard)/admin/users/[id]/page.tsx:118]: thanks!

packages/lib/server-only/user/get-all-users.ts
catalinpit: good catch. thanks!

packages/lib/server-only/user/get-all-users.ts
```diff
@@ -0,0 +1,71 @@
+import { Prisma } from '@prisma/client';
```
Mythie, line 1: Sorry I just noticed this, replace these imports with `@documenso/prisma/client` 😄 ❤️

catalinpit: Yeah, but it doesn't have the `validator` and `UserWhereInput`. I removed it and made some changes though, so it should be fine now. 🙏

packages/lib/server-only/user/get-all-users.ts
catalinpit: Ok, implemented the feedback. Thanks!

packages/lib/server-only/user/get-all-users.ts
Mythie: Not really at the moment but later as we add a real API we can defer to the response from this. Happy to ignore for now since it'll be additive to the function contract.

PR #400, opened 2023-09-22
feat: add early adopters graph

dephraiim:


ElTimuro (commented):
The scraper is called StargRazer, did you intentionally call it <BarMetric<StargazersType>?

dephraiim:
@ElTimuro, I was using the same chart for both the github metrics and the early adopters and it was the same code through out so I refactored it into a reusable component as BarMetric to serve both the github graph and the early adopters

ElTimuro:
> @ElTimuro, I was using the same chart for both the github metrics and the early adopters and it was the same code through out so I refactored it into a reusable component as BarMetric to serve both the github graph and the early adopters

Makes total sense, I was just talking about the "r" 😀

dephraiim:
I'll take care of it. I made it Stargazer all round, I'll update it

ElTimuro:
lgtm

ElTimuro:
Syncing to documenso.next via 🫱

PR #429, opened 2023-09-26
chore: next 13.5.3

Mythie:
Fails due to trying to bundle `canvas` which is explicitly ignored using a dynamic/suspense component with `ssr: false`.


Mythie:
Not today javascript

PR #430, opened 2023-09-26
feat: update email templates

dguyen:
Changes:

- Improve consistency across email templates
- Improve button alignments
- Add SPM email attachment
- Add "Sign In" button for forgot password email template
- Update SPM email template to promote new accounts


apps/marketing/src/components/(marketing)/single-player-mode/create-single-player-document.action.ts
```diff
@@ -177,9 +177,7 @@ export const createSinglePlayerDocument = async (
     },
   );
 
-  // Todo: Handle `downloadLink`
   const template = createElement(DocumentSelfSignedEmailTemplate, {
-    downloadLink: `${process.env.NEXT_PUBLIC_MARKETING_URL}/single-player-mode/${documentToken}`,
```
Mythie, line 182: We still want the download link I think 🤔

dguyen: How come? The "download" link just takes them back to the SPM success page

The PDF is now directly attached to the email, so they can directly view/download it

Mythie: Yep you're right I'm the silly one

packages/email/template-components/template-document-self-signed.tsx
```diff
@@ -49,7 +49,7 @@ export const TemplateDocumentSelfSigned = ({
         <Text className="mx-auto mb-6 mt-1 max-w-[80%] text-center text-base text-slate-400">
           Create a{' '}
           <Link
-            href={`${process.env.NEXT_PUBLIC_WEBAPP_URL}/signup`}
+            href={`${process.env.NEXT_PUBLIC_WEBAPP_URL ?? 'http://localhost:3000'}/signup`}
```
Dun-sin, line 52: don't you think this will be neater doing it in a variable?

Dun-sin: of cuz it has nothing to do with your PR, but could be a good first issue

dguyen: Hi Dun-sin!

Thanks for pointing that out, it made me realise that this path is also required somewhere else in this file.

I'll update it as part of this PR, thanks!

Dun-sin: I noticed it's a universal thing in some other files

PR #440, opened 2023-09-27
chore: add empty message

### LINKED ISSUE #437: Missing Empty Recipients List Message in Staging
So the Recipients List is missing an empty message when a particular recipient can't be found.

This opens up a feature allowing the user to add a recipient if it can't be found in the list.
I think for version 1.0 We can show the empty message.
Let me know what you think.

Screenshot (63)

PeterKwesiAnsah:
This PR adds an empty message to the recipient list.
Fixes #437

### DIFF
--- a/packages/ui/primitives/document-flow/add-fields.tsx
+++ b/packages/ui/primitives/document-flow/add-fields.tsx
@@ -341,7 +341,9 @@ export const AddFieldsFormPartial = ({
                 <Command>
                   <CommandInput />
                   <CommandEmpty>
-                    <span>No Recipient matching this description was found.</span>
+                    <span className="text-muted-foreground inline-block px-4">
+                      No recipient matching this description was found.
+                    </span>
                   </CommandEmpty>
 
                   <CommandGroup>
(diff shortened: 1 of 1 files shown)


packages/ui/primitives/document-flow/add-fields.tsx
```diff
@@ -340,7 +340,9 @@ export const AddFieldsFormPartial = ({
               <PopoverContent className="p-0" align="start">
                 <Command>
                   <CommandInput />
-                  <CommandEmpty />
+                  <CommandEmpty>
+                    <span>No Recipient matching this description was found.</span>
```
dguyen, line 344: Hi Peter!

Could you please add the following className to the span:
- `text-muted-foreground inline-block px-4`

And also lowercase Recipient?

Thank you!

dguyen (changes requested):

packages/ui/primitives/document-flow/add-fields.tsx
PeterKwesiAnsah: alright @dguyen 
Thanks

PeterKwesiAnsah: @dguyen  updated

Mythie (approved):

Mythie:
Thanks for your work!

giphy

PR #460, opened 2023-09-28
fix: hiding of action buttons

### LINKED ISSUE #441: [Bug] - Action Buttons Gets Hidden
**Description:**

On the **AddSigners** page, if I add N number of signers, and scroll down, the action buttons for `Go back` and `Continue` gets hidden 😶 

Added a video below for reference :)

[NoActionButtons.webm](https://github.com/documenso/documenso/assets/23498248/74757634-e139-49d0-9876-441e80f004b9)

Also, the area seems to be not properly sticky which was mentioned [here](https://github.com/documenso/documenso/issues/432#issuecomment-1735274536)  as well

adithyaakrishna:
**Description:**

This PR fixes #441


packages/ui/styles/theme.css
```diff
@@ -93,3 +93,14 @@
...
+.hide-scrollbar::-webkit-scrollbar {
+      display: none;
+  }
+
+.hide-scrollbar {
+    /* IE and Edge */
+    -ms-overflow-style: none;
+    /* Firefox */
+    scrollbar-width: none;
+}
```
adithyaakrishna, line 106: @Mythie I wasn't sure if webkit css rules are allowed for inline in css, so added them here. If you give an ack regarding inline ones, will add it through tailwind classes :)

and if so, I will just add `[&::-webkit-scrollbar]:hidden` to the classes

dguyen:
I don't think it's appropriate to hide the scrollbar when overflowing

I think the following changes would be optimal:

- Custom vertical scrollbar 
- When the user click add signer and the content is overflowing, smooth scroll to ensure the 'add signer' button is still visible

Custom vertical scrollbar would look something like this (rounded & if possible have the background transparent)

image

adithyaakrishna:
@dguyen Sure, that makes sense, I will send the changes in a while :)

adithyaakrishna:
@dguyen Would love to know your opinion on how does this look? Used the documenso gray and also made the background transparent :)

Screenshot 2023-10-05 at 3 34 45 PM

packages/ui/primitives/document-flow/document-flow-root.tsx
```diff
@@ -61,8 +61,14 @@ export const DocumentFlowFormContainerContent = ({
...
-      <div className="-mx-2 flex flex-1 flex-col overflow-y-auto px-2">{children}</div>
+    <div
+      className={cn(
+        `flex flex-1 flex-col overflow-y-auto overflow-x-hidden [&::-webkit-scrollbar]:hidden`,
+        className,
+      )}
+      {...props}
+    >
+      <div className="-mx-2 flex flex-1 flex-col px-2">{children}</div>
     </div>
```
dguyen, line 72: > I'm facing one small issue tho, kinda stuck on it. Could you please lmk what I can do here so that the input field gets borders on all sides or perhaps how to prevent the scollbar from pushing the elements to the left?

You can try this:

```
    <div className={cn('-mx-2 flex flex-1 flex-col overflow-y-auto px-2', className)} {...props}>
      {children}
    </div>
```

dguyen:
> Would love to know your opinion on how does this look? Used the documenso gray and also made the background transparent :)

It looks nice in the image, but for MacOS it looks a bit strange since it just cuts off the top and bottom (might just be a MacOS thing though)

image

packages/ui/primitives/document-flow/document-flow-root.tsx
adithyaakrishna: Oki Doki, will try that one :)

adithyaakrishna:
@dguyen Updated the CSS rules for the scrollbar, we won't get the background for the scrollbar and hence preventing the cut-off as well :)

dguyen:
@adithyaakrishna the scrollbar looks nice!

Although now it seems that the input ring is now being cut off for me (MacOS Chrome)

image

adithyaakrishna:
Its again the macOS, smh! I will fix that one up too :)

adithyaakrishna:
Fixed it :)

macOS: 0 | Adi: 1 xP

dguyen:
Sorry for the delay, just tested on MacOS/Windows and it looks good to me, thanks!

PR #469, opened 2023-09-29
feat: added password validation

### LINKED ISSUE #464: Secure Password Parameters
**Currently**:
- Passwords can be easily guessed or brute-forced.
- Takes a minimum password length of 8 characters.

**Desired State**
- Must include at least:
	- One uppercase letter.
	- One lowercase letter.
	- One digit.
- Restriction against using easily guessable passwords (e.g., "password123", "12345678")
- Passwords should not be based on the user's personal information (e.g., name, email).
- Errors and alerts for password change if it detects weak or commonly used passwords.

Password Strength

adithyaakrishna:
**Description:**

This PR Fixes #464


adithyaakrishna:
@dephraiim @Mythie apparently while I was working on this, I pulled the latest changes and my local dev env broke, lol. So keeping this in WIP 

Duncan, could you please help me run it again? Waiting for discord to come back up :(

dguyen:
@adithyaakrishna are you still having trouble with the local dev environment?

adithyaakrishna:
Hey @dguyen 
Nope, not anymore.  I will finish this up by today and send a commit for this :)

adithyaakrishna:
@dguyen Could you please review this PR and lmk if any changes are necessary?

apps/web/src/components/forms/password.tsx
```diff
@@ -105,7 +133,7 @@ export const PasswordForm = ({ className }: PasswordFormProps) => {
           <Input
             id="current-password"
             type={showCurrentPassword ? 'text' : 'password'}
-            minLength={6}
+            minLength={8}
```
dguyen, line 136: You should leave it as `6` because this may prevent old passwords from being reset.

apps/web/src/components/forms/password.tsx
```diff
@@ -18,12 +18,40 @@ import { useToast } from '@documenso/ui/primitives/use-toast';
...
+const commonPasswords = [
+  'password',
+  '12345678',
+  'password123',
+  'admin',
+  'qwerty',
+  'poiuyt',
+  'qwerty123',
+  'qwerty@123',
+];
```
dguyen, line 30: The majority of these passwords would fail the regex anyway. I think just remove common passwords for now.

apps/web/src/components/forms/password.tsx
```diff
@@ -18,12 +18,40 @@ import { useToast } from '@documenso/ui/primitives/use-toast';
...
+    password: z
+      .string()
+      .regex(new RegExp('.*[A-Z].*'), { message: 'One uppercase character' })
+      .regex(new RegExp('.*[a-z].*'), { message: 'One lowercase character' })
+      .regex(new RegExp('.*\\d.*'), { message: 'One number' })
+      .regex(new RegExp('.*[`~<>?,./!@#$%^&*()\\-_+="\'|{}\\[\\];:\\\\].*'), {
+        message: 'One special character is required',
+      })
+      .min(8, { message: 'Must be at least 8 characters in length' })
+      .max(72, { message: 'Cannot be more than 72 characters in length' }),
```
dguyen, line 44: Extract this out to something like `ZPasswordSchema` which then can be reused between `ZPasswordFormSchema` and `ZSignUpFormSchema`

dguyen (changes requested):

apps/web/src/components/forms/password.tsx
adithyaakrishna: Gotcha, will update it :)

apps/web/src/components/forms/password.tsx
adithyaakrishna: Okay, I'll remove the refine condition for the same as well

apps/web/src/components/forms/password.tsx
adithyaakrishna: Ahh! True, that would prevent current/existing users from logging in, thanks for pointing it out, will fix it in the next commit :)

packages/trpc/server/password.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from 'zod';
```
Mythie, line 1: This doesn't feel like the right place for it but I'm not sure where is. `packages/lib/universal/password.ts` might be an idea? 🤔

dguyen: I think `packages/trpc/server/auth-router/schema.ts` or `packages/lib/types/password`

adithyaakrishna:
GitHub Bot, why bro? Why so mean? Take some vacation 😭

packages/trpc/server/password.ts
dguyen: @adithyaakrishna

apps/web/src/components/forms/password.tsx
```diff
@@ -20,9 +21,9 @@ import { FormErrorMessage } from '../form/form-error-message';
 
 export const ZPasswordFormSchema = z
   .object({
-    currentPassword: z.string().min(6).max(72),
```
dguyen, line 23: Just to check, min(6) overwrites the min(8) in `ZPasswordSchema` right?

dguyen: Yeah I just wanted make sure that it actually overwrites it

adithyaakrishna: It does, as the preference is given to the min(6) in this case

dguyen: I think you should create a ZCurrentPasswordSchema which is `z.string().min(6).max(72)` and use that wherever a user has to put in a current password.

If this changes goes through as it currently is, we'd be locking some people out of their accounts

adithyaakrishna:
@dguyen Updated the schemas and made the requested changes, could you please take another look?

apps/web/src/components/forms/password.tsx
adithyaakrishna: Updated it to have `ZCurrentPasswordSchema `

dguyen (approved):
lgtm

PR #486, opened 2023-10-02
feat: add safari clipboard copy support

### LINKED ISSUE #467: [BUG][SAFARI ONLY] Unable to copy document sharing link.
Steps to repro:
1. Click on any actions that is supposed to perform a copy the document link.
2. Open the inspector, Voila, Err!

Screenshot 2023-09-29 at 5 51 49 PM

dguyen:
Fixes #467

## Changes:

- Added `ClipboardItem` to handle copying async values for Safari
- Extracted `useCopyToClipboard` to @documenso/lib
- Extracted duplicate logic for copying shared links to it's own hook

### MacOS:

- Safari
- Firefox
- Chrome

### iOS:

- Safari

### Ubuntu:

- Chrome
- Firefox


packages/lib/client-only/hooks/use-copy-share-link.ts
```diff
@@ -0,0 +1,59 @@
+import { trpc } from '@documenso/trpc/react';
+import { TCreateOrGetShareLinkMutationSchema } from '@documenso/trpc/server/share-link-router/schema';
+import { useToast } from '@documenso/ui/primitives/use-toast';
```
Mythie, line 3: Feels like this might cause a circular dep if stuff in UI depends on lib, consider an onSuccess callback instead :)

dguyen: I believe this would be no different then putting it in a hooks folder on `apps/marketing` or` apps/web`

The only reason this is in the lib package is to reduce duplicate code, using `onSuccess` would also increase duplicate code

Mythie: No I mean we have things in the `ui` package that require things from `lib` right, so having a lib item that relies on `ui` is circular. I'd move it to the `ui` package potentially.

dguyen: Okay so I'll just dump `useCopyShareLink` into `packges/ui/hooks`? Doesn't really feel right

I'd rather extract `useToast` hook out of the UI library, but that relies on types from the toast component, which if we can't really extract

Mythie: I think for this we can cheat and accept an optional `onError` callback in the hook that will do the toast when used within the webapp or marketing site 😇 

Saves bikeshedding on where things should be!

dguyen: I don't really agree but it's done, pls look

PR #489, opened 2023-10-02
fix: signature text overflow truncated for longer signature texts

arjunbharti:
Bug:

https://github.com/documenso/documenso/assets/60930192/1a368f70-ba4b-401d-94ee-e822d2ac58c8

After fix:
Screenshot 2023-10-02 at 5 58 38 PM


dguyen:
@arjunbharti thanks for that

/tip 25

PR #495, opened 2023-10-02
fix: fetch the correct number of open issues using the github search api

dephraiim:


apps/marketing/src/app/(marketing)/open/page.tsx
```diff
@@ -43,18 +47,28 @@ export type StargazersType = z.infer<typeof ZStargazersLiveResponse>;
...
 
 export default async function OpenPage() {
-  const {
-    forks_count: forksCount,
-    open_issues: openIssues,
-    stargazers_count: stargazersCount,
-  } = await fetch('https://api.github.com/repos/documenso/documenso', {
-    headers: {
-      accept: 'application/vnd.github.v3+json',
+  const { forks_count: forksCount, stargazers_count: stargazersCount } = await fetch(
```
dguyen, line 50: Is there a reason we aren't doing all these fetch requests in parallel?

https://nextjs.org/docs/app/building-your-application/data-fetching/patterns#parallel-data-fetching

dephraiim: I've updated it to use parallel routes

Mythie:
@dephraiim If you're good we're good

PR #502, opened 2023-10-03
feat: add playwright

catalinpit:


packages/prisma/schema.prisma
```diff
@@ -133,11 +133,11 @@ model DocumentData {
 }
 
 model DocumentMeta {
-  id       String    @id @default(cuid())
-  subject  String?
-  message  String?
-  documentId Int @unique
-  document Document @relation(fields: [documentId], references: [id], onDelete: Cascade)
+  id         String   @id @default(cuid())
```
catalinpit, line 136: have no idea what all this stuff is

.env.example
```diff
@@ -15,6 +15,11 @@ NEXT_PRIVATE_DATABASE_URL="postgres://documenso:password@127.0.0.1:54320/documen
 # Defines the URL to use for the database when running migrations and other commands that won't work with a connection pool.
 NEXT_PRIVATE_DIRECT_DATABASE_URL="postgres://documenso:password@127.0.0.1:54320/documenso"
 
+# [[E2E Tests]]
+E2E_TEST_AUTHENTICATE_USERNAME="Test User"
+E2E_TEST_AUTHENTICATE_USER_EMAIL="testuser@mail.com"
+E2E_TEST_AUTHENTICATE_USER_PASSWORD="test_password"
```
dguyen, line 21: These don't seem to be used anywhere?

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+          - 5432:5432
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - name: Install dependencies
+        run: npm ci
+      - name: Copy env
+        run: cp .env.example .env
```
dguyen, line 32: I thought this needs to be `.env.local`?

packages/lib/server-only/user/delete-user.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  if (!user) {
+    throw new Error(`User with name ${name} not found`);
+  }
+
+  return await prisma.user.delete({
+    where: {
+      id: user.id,
+    },
+  });
```
dguyen, line 20: It feels a little dodgy to delete by name due to no unique restraints.

Is there a reason we can't just do something like this?

```
export const deleteUserByEmail = async (email string) => {
  return prisma.user.delete({
    where: {
      email,
    },
  });
});
```

packages/web-tests/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "web-tests",
```
dguyen, line 2: Just checking, in the future if we want tests for the marketing website, will it be a separate package?

We also might want to have it as `@documenso/web-tests` 

or `@documenso/app-tests` if we can include marketing in this package in the future

packages/web-tests/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test:dev": "playwright test",
+    "test:e2e": "start-server-and-test \"(cd ../../apps/web && npm run start)\" http://localhost:3000 \"playwright test\""
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
dguyen, line 12: Not sure how licenses works, but is this the correct license?

Our packages seem to use AGPLv3/AGPL-3.0/MIT

packages/lib/server-only/user/delete-user.ts
catalinpit: It's deleting by ID, not name. Although, I can change it to email if you think it's better.

.github/workflows/e2e-tests.yml
catalinpit: It's `.env.example` in Documenso, right?

packages/lib/server-only/user/delete-user.ts
dguyen: The ID is found by fetching a user by their name though

I think it's better to just delete by unique value, in this case email is unique

.github/workflows/e2e-tests.yml
dguyen: Yeah but for the actual env file isn't it supposed to be `.env.local`? Unless this is for prisma

https://github.com/documenso/documenso/blob/feat/refresh/apps/web/next.config.js#L6

packages/lib/server-only/user/delete-user.ts
catalinpit: Ah, I get what you meant. Well, I figured that "test user" is unlikely to clash with other usernames, but I changed it to the email anyways. Thanks!

.github/workflows/e2e-tests.yml
catalinpit: Tests file cause they can't access the env vars if I change it to `.env.local`

Mythie (approved):

Mythie:
The end of an era! Thanks for being patient with me 😄

Issue #499, opened 2023-10-03
[bug] Email Requesting Signature shows "Completed Document" in Preview Text


daallgeier:
When receiving an email from documenso with a new document asking to sign the Preview Text (next or under the email subject in an inbox) shows the Preview as "Completed Document" while it should be "Signature Requested" or something similar.

How to reproduce?
1. Send signature request to gmail
2. Be in Inbox 
3. Look for "completed document" next to the subject 

![image](https://github.com/documenso/documenso/assets/9637909/0358cf92-7a27-461a-956a-736e7ff2dc75)

Other suggestions for a good Preview Text
- Signature Requested by {sender.name} on {document.name}
- Signature Requested on {document.name}
- Please sign {document.name}
- Please sign {document.name} by {sender.name} by {date.due}

catalinpit:
Thanks for reporting this, @daallgeier!

dguyen:
Thanks for finding this 😄 

/tip 25

PR #505, opened 2023-10-03
fix: email requesting signature shows "completed document" in preview…

18feb06:
… text #499
fix: #499


adithyaakrishna:
@18feb06 The PR seems to contain unnecessary changes with respect to some other files, could you please revert those and keep the ones that are needed for this issue?

18feb06:
@adithyaakrishna okay

PR #513, opened 2023-10-03
fix: changed text of stepper

### LINKED ISSUE #500: [UX Improvement] Progress Numbering on New Document is confusingn
When creating a new document the progress numbering is confusing on the first step to "add signers".

`Add Signers (1/3)` makes me think I can only add 3 signers which obviously isn't the case.

Hallidayo:
### Issue Type

UI Bug

### Current Issue

The stepper titles are a little confusing

### Fix

Changed to use Step of:

Screenshot 2023-10-03 at 19 54 25

Closes #500


dguyen:
Cheers!

/tip 25

PR #529, opened 2023-10-05
feat: add dark mode toggle

adithyaakrishna:
**Description:**

- This PR combines the dark mode and light mode buttons in the footer of https://documenso.com/
- Simplified the field mappers
- Simplified the recipient formatter


packages/lib/utils/recipient-formatter.ts
```diff
@@ -3,7 +3,7 @@ import { Recipient } from '@documenso/prisma/client';
 export const recipientInitials = (text: string) =>
   text
     .split(' ')
-    .map((name: string) => name.slice(0, 1).toUpperCase())
+    .map((name: string) => name[0].toUpperCase())
```
dguyen, line 6: I don't think this safeguards against empty strings, it's probably best to leave it as slice unless there is an issue.

apps/marketing/src/components/(marketing)/single-player-mode/create-single-player-document.action.ts
```diff
@@ -214,11 +213,14 @@ const mapField = (
...
-    .with(FieldType.NAME, () => signer.name)
-    .otherwise(() => '');
+  const customText =
+    field.type === FieldType.DATE
+      ? DateTime.now().toFormat('yyyy-MM-dd hh:mm a')
+      : field.type === FieldType.EMAIL
+      ? signer.email
+      : field.type === FieldType.NAME
+      ? signer.name
+      : '';
```
dguyen, line 223: Sorry but we have decided to use `match` for the majority of use cases such as this in our codebase.

Nested ternaries are rather unreadable, messy and more complex than `match`.

Note: If `match` did not exist in this scenario a switch case would be preferred.

apps/marketing/src/components/(marketing)/footer.tsx
```diff
@@ -81,15 +81,26 @@ export const Footer = ({ className, ...props }: FooterProps) => {
...
+                <Sun className="h-5 w-5" />
+                <span className="sr-only">Light</span>
+              </>
+            ) : (
+              <>
+                <Moon className="h-5 w-5" />
+                <span className="sr-only">Dark</span>
+              </>
+            )}
           </button>
```
dguyen, line 104: @Mythie what do you think of combining the buttons?

dguyen (changes requested):

apps/marketing/src/components/(marketing)/footer.tsx
Mythie: I think I prefer them individually since it matches the current app, but am open to opinions 😄

apps/marketing/src/components/(marketing)/single-player-mode/create-single-player-document.action.ts
Mythie: Yeah sorry, if you compare match with the ternary here it becomes pretty clear how powerful it is from a readability perspective 😇

packages/lib/utils/recipient-formatter.ts
Mythie: That's correct!
image

apps/marketing/src/components/(marketing)/footer.tsx
dguyen: Just noting our designer is going to look into this, I'll update this once we decide on how to proceed.

dguyen: @adithyaakrishna our designer came back with this:

https://www.figma.com/proto/Qm2heQggjZyygnSQdS22Xa/Design-exploration?type=design&node-id=101-11&t=kwthccpzca53L9vd-1&scaling=scale-down&page-id=0%3A1&starting-point-node-id=101%3A11&show-proto-sidebar=1

If you want to pass on implementing this please let us know so we can release it to the community to implemented

adithyaakrishna: @dguyen This looks super cool! I can work on this. Is it fine if I send a fix over the weekend?

apps/marketing/src/components/(marketing)/single-player-mode/create-single-player-document.action.ts
adithyaakrishna: @Mythie Makes sense, I guess this aligns with another ternary change which we had discussed wrt another issue, I will revert this :)

packages/lib/utils/recipient-formatter.ts
adithyaakrishna: Oops! Thanks for pointing it out, I hadn't considered the empty string case. I will revert this change as well :)

apps/marketing/src/components/(marketing)/footer.tsx
dguyen: Yeah that's fine!

adithyaakrishna: @dguyen Could you please share the Figma link for the same? the above one seems to be for the prototype. I thought to use the CSS styles from the Figma file itself

dguyen: @adithyaakrishna I currently don't have access to the figma file as our designer is currently on leave

I'll try follow up with this once he's back, but it might take some time (days)

adithyaakrishna: Could you please let me know if the above changes looks good?

apps/marketing/src/components/(marketing)/single-player-mode/create-single-player-document.action.ts
```diff
@@ -214,11 +213,14 @@ const mapField = (
   field: TCreateSinglePlayerDocumentSchema['fields'][number],
   signer: TCreateSinglePlayerDocumentSchema['signer'],
 ) => {
-  const customText = match(field.type)
-    .with(FieldType.DATE, () => DateTime.now().toFormat('yyyy-MM-dd hh:mm a'))
-    .with(FieldType.EMAIL, () => signer.email)
-    .with(FieldType.NAME, () => signer.name)
-    .otherwise(() => '');
+  const customText =
```
Mythie, line 216: Just confirming that this is being reverted

packages/lib/utils/recipient-formatter.ts
Mythie, line 6: Same here

packages/ui/primitives/theme-switcher.tsx
```diff
@@ -0,0 +1,33 @@
...
+  theme?: string;
+  // eslint-disable-next-line no-unused-vars
+  setTheme: (theme: string) => void;
+}
+
+export const ThemeSwitcher = (props: IProps) => {
+  const { theme, setTheme } = props;
+
+  return (
+    <ToggleGroup.Root
```
Mythie, line 14: Feel like we can solve this with some cheeky tailwind and framer motion. Something tells me that the `layoutId` will play a part in it.

apps/marketing/src/components/(marketing)/single-player-mode/create-single-player-document.action.ts
adithyaakrishna: Oops! Thanks, Lucas. I had missed that one, just sent a commit to revert it :)

packages/lib/utils/recipient-formatter.ts
adithyaakrishna: Sent a commit to revert this as well :)

packages/ui/primitives/theme-switcher.tsx
adithyaakrishna: I'm a bit new to framer motion, but understood the basic stuff on how `motion.div` and all works. Plus ChatGPT is useful sometimes xD

Just sent a commit below to use framer motion for this functionality :)

packages/ui/primitives/theme-switcher.tsx
```diff
@@ -0,0 +1,63 @@
...
+    >
+      <motion.div
+        className="absolute h-8 w-10 rounded-full bg-white shadow-sm dark:bg-stone-600"
+        initial={false}
+        animate={theme}
+        variants={variants}
+        transition={{ duration: 0.2 }}
+      ></motion.div>
+      <button
+        className={`z-10 flex h-8 w-10 items-center justify-center rounded-full ${getIconColor(
```
dephraiim, line 36: Can you replace this with `cn` utility from `@documenso/ui/lib/utils`?

packages/ui/primitives/theme-switcher.tsx
```diff
@@ -0,0 +1,63 @@
...
+        )}`}
+        onClick={() => setTheme('system')}
+      >
+        <Monitor className="h-5 w-5" />
+      </button>
+    </div>
+  );
+};
+
+export default ThemeSwitcher;
```
dephraiim, line 63: We prefer to use named exports over default exports.

packages/ui/primitives/theme-switcher.tsx
```diff
@@ -0,0 +1,63 @@
+import React, { FC } from 'react';
+
+import { motion } from 'framer-motion';
+import { Monitor, MoonStar, Sun } from 'lucide-react';
+
+interface ThemeSwitcherProps {
```
dephraiim, line 6: We prefer types over interfaces to avoid mixing the two

packages/ui/primitives/theme-switcher.tsx
```diff
@@ -0,0 +1,63 @@
...
+  theme?: string;
+  // eslint-disable-next-line no-unused-vars
+  setTheme: (theme: string) => void;
+}
+
+const ThemeSwitcher: FC<ThemeSwitcherProps> = ({ theme, setTheme }) => {
```
dephraiim, line 12: We tend to avoid using the `React.FC` type

dephraiim (changes requested):

packages/ui/primitives/theme-switcher.tsx
adithyaakrishna: Done ✅

packages/ui/primitives/theme-switcher.tsx
adithyaakrishna: Ohh that makes sense why I didn't see many usages interfaces, I will fix that one up as well :)

packages/ui/primitives/theme-switcher.tsx
```diff
@@ -0,0 +1,65 @@
...
+    >
+      <motion.div
+        className="absolute h-8 w-10 rounded-full bg-white shadow-sm dark:bg-stone-600"
+        initial={false}
+        animate={theme}
+        variants={variants}
+        transition={{ duration: 0.2 }}
+      ></motion.div>
+      <button
+        className={cn(
```
dephraiim, line 40: This is a great use of `cn`

```js
 className={cn(
      'z-10 flex h-8 w-10 items-center justify-center rounded-full',
      getIconColor('light'),
)}
```

No need for template strings

dephraiim (changes requested):

packages/ui/primitives/theme-switcher.tsx
adithyaakrishna: Updated it with the below commit :)

packages/ui/primitives/theme-switcher.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const ThemeSwitcher = () => {
+  const isMounted = useIsMounted();
+
+  const { theme, setTheme } = useTheme();
+
+  const variants = {
+    light: { x: 0 },
+    dark: { x: 'calc((2.15rem + 0.30rem) * 1)' },
+    system: { x: 'calc((2rem + 0.5rem) * 2)' },
```
dguyen, line 16: How are these calculated? Maybe we can extract them into variables which can be used as documentation?

Mythie (approved):

PR #547, opened 2023-10-09
fix: bypass signature fix (#536)

### LINKED ISSUE #536: [Bug] Ability to mark Document signing as completed without actually signing the document
Ability to mark Document signing as completed without actually signing the document

https://github.com/documenso/documenso/assets/145282438/a7464dde-541d-4bc9-8d04-b04011788a42

As shown in the video, we are able to click on complete button without signing.Ideally it should ask the user to sign the document.
Although it  does  not proceed on the ui but the document signing is marked completed.
At last if we download the document we do not see any signature in the final document.

Steps to reporduce

1.Create a new documenso signing with at least two receipients.
2.sign the document normally with one user.
3. with other user ,click on the signing link 
4.click on the signature field . make signature, after that click on cross (x) 
5.click on signature again.
6.now click on complete , and wait for 30 seconds.
7.Ui will not proceed but spinner icon keeps on spinning
8.Go to email and click on the same link to sign again.
9. it will show that document is already signed, download the document
10.there will be no signatures
11. Now go to the documenso dashboard, it will be marked as completed. download the file. it will have no signatures.

Expected Behavior
1.When we click on complete without signing, we should get a prompt to complete signing
2.Document should not be marked as completed.

Current Behavior
1.the spinner keeps on spinning
2.document is marked as completed without any signature

Environment
OS:ubuntu 22
Browser:Chrome Version 117.0.5938.132 (Official Build) (64-bit)

Abhinav-Developer-23:
https://github.com/documenso/documenso/assets/145282438/ce002b13-300d-4ef5-841c-9d51280527b9

fixes #536

### DIFF
--- a/apps/web/src/app/(signing)/sign/[token]/signature-field.tsx
+++ b/apps/web/src/app/(signing)/sign/[token]/signature-field.tsx
@@ -48,7 +48,7 @@ export const SignatureField = ({ field, recipient }: SignatureFieldProps) => {
 
   const [showSignatureModal, setShowSignatureModal] = useState(false);
   const [localSignature, setLocalSignature] = useState<string | null>(null);
-  const [isDocumentSigned, setIsDocumentSigned] = useState(false);
+  const [isLocalSignatureSet, setIsLocalSignatureSet] = useState(false);
   const state = useMemo<SignatureFieldState>(() => {
     if (!field.inserted) {
       return 'empty';
@@ -62,15 +62,15 @@ export const SignatureField = ({ field, recipient }: SignatureFieldProps) => {
   }, [field.inserted, signature?.signatureImageAsBase64]);
 
   useEffect(() => {
-    if (showSignatureModal == false && isDocumentSigned == false) {
+    if (!showSignatureModal && !isLocalSignatureSet) {
       setLocalSignature(null);
-      return;
     }
-  }, [showSignatureModal]);
+  }, [showSignatureModal, isLocalSignatureSet]);
+
   const onSign = async (source: 'local' | 'provider' = 'provider') => {
     try {
       if (!providedSignature && !localSignature) {
-        setIsDocumentSigned(false);
+        setIsLocalSignatureSet(false);
         setShowSignatureModal(true);
         return;
       }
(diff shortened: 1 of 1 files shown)


Mythie:

https://documen.so/cla

Abhinav-Developer-23:
Fixes issue #536

Abhinav-Developer-23:
Hi, @Mythie i have accepted the contributor license agreement.

Mythie:
This seems to just ensure that the widget on the side has a signature in it even if it's not used on the document correct?

apps/web/src/app/(signing)/sign/[token]/signature-field.tsx
```diff
@@ -61,9 +61,16 @@ export const SignatureField = ({ field, recipient }: SignatureFieldProps) => {
     return 'signed-text';
   }, [field.inserted, signature?.signatureImageAsBase64]);
 
+  useEffect(() => {
+    if (showSignatureModal == false && isDocumentSigned == false) {
+      setLocalSignature(null);
+      return;
+    }
+  }, [showSignatureModal]);
```
Comment (dguyen) [inline apps/web/src/app/(signing)/sign/[token]/signature-field.tsx:69]: I think you should include all the required items in the dependency array

Other points:
- return is redundant
- `===` is preferred over `==` 
- `!value` is also generally preferred

```
  useEffect(() => {
    if (!showSignatureModal && !isDocumentSigned) {
      setLocalSignature(null);
    }
  }, [showSignatureModal, isDocumentSigned]);
```

Could you also add some spacing below this useEffect?

apps/web/src/app/(signing)/sign/[token]/signature-field.tsx
```diff
@@ -48,7 +48,7 @@ export const SignatureField = ({ field, recipient }: SignatureFieldProps) => {
 
   const [showSignatureModal, setShowSignatureModal] = useState(false);
   const [localSignature, setLocalSignature] = useState<string | null>(null);
-
+  const [isDocumentSigned, setIsDocumentSigned] = useState(false);
```
Comment (dguyen) [inline apps/web/src/app/(signing)/sign/[token]/signature-field.tsx:51]: I think `isDocumentSigned` should be updated to something like `isLocalSignatureSet` or something

dguyen:
Personally I think the better way would be to drop the provider param and do something like this:

```
const onSign = async (signature?: string) => {
  const signatureValue = signature ?? providedSignature;

  try {
    if (!signatureValue) {
      setLocalSignature(null);
      setShowSignatureModal(true);
      return;
    }

    // ...
```

which means we don't need a useEffect and state to ensure a signature is present

what do you think @Mythie

Abhinav-Developer-23:
@dguyen I have made the improvements that you suggested in the code.

dguyen:
@Abhinav-Developer-23 looks good to me, thanks!

/tip 45

PR #548, opened 2023-10-09
fix(pre-commit): added lint staged to install and add all packages from package.json and add to commit

### LINKED ISSUE #546: Add a lint staged task for package.json changes
## Improvement Description

Add a lint-staged task that identifies if any `package.json` files have been modified and then runs `npm install` to ensure we always have the latest package.json file.

## Rationale

Whenever I encounter an issue I like to try and solve it for good so we don't have to worry about it again. As per #523 we accidentally forgot to commit the package-lock file and thus caused CI failures, could have happened to anyone so we should try and alleviate the burden of remembering such a mundane thing.

## Proposed Solution

Create the lint staged task as per above

## Alternatives (optional)

N/A

## Additional Context

N/A

## Checklist

- [x] I have searched the existing issues and improvement suggestions to avoid duplication.
- [x] I have provided a clear description of the improvement being suggested.
- [x] I have explained the rationale behind this improvement.

adityadeshlahre:
Fix #546 
Sorry for late PR i was stuck in some permission issue with files !
This might be the best solution !
**NOTE : I will edit the readme.md file after suggestions from maintainers.**

if any changes required let me know.
Thank you 

**SCRIPT : precommit runs before commit. [As shown in image below]**
image


lint-staged.config.cjs
```diff
@@ -1,3 +1,7 @@
 module.exports = {
   '**/*.{js,jsx,cjs,mjs,ts,tsx,cts,mts,mdx}': ['prettier --write'],
 };
+
+module.exports = {
+  '*.json': ['npm run precommit'],
```
Mythie, line 6: This should probably be more specific to package.json

package.json
```diff
@@ -16,7 +16,8 @@
     "dx:down": "docker compose -f docker/compose-services.yml down",
     "prisma:migrate-dev": "npm run with:env -- npm run prisma:migrate-dev -w @documenso/prisma",
     "prisma:migrate-deploy": "npm run with:env -- npm run prisma:migrate-deploy -w @documenso/prisma",
-    "with:env": "dotenv -e .env -e .env.local --"
+    "with:env": "dotenv -e .env -e .env.local --",
+    "precommit": "npm install && git add package.json package-lock.json"
```
Mythie, line 20: Iirc lint staged will happily modify staged files, is this just because package-lock might not be staged?

adityadeshlahre: yeah after running the script package-lock.json didn't get committed that's why tryed with script !

lint-staged.config.cjs
adityadeshlahre: module.exports = {
  '**/*/package.json': ['npm run precommit'],
};

this should work for nested directories too !

ElTimuro:
image

lint-staged.config.cjs
```diff
@@ -1,3 +1,7 @@
 module.exports = {
   '**/*.{js,jsx,cjs,mjs,ts,tsx,cts,mts,mdx}': ['prettier --write'],
 };
+
+module.exports = {
+  '**/*/package.json': ['npm run precommit'],
+};
```
adithyaakrishna, line 7: EOL is missing in this file :)

Mythie:
Thanks for your contribution! Sorry it took forever to get across 😬 

its-done-frodo

PR #557, opened 2023-10-13
fix: fix for Accepting signatures or text fields with white space only  #551

Abhinav-Developer-23:
fixes #551 
There are multiple places where text fields can be filled with white space only.
In this PR I have fixed on 4 pages. Please find the working demo video.

https://stg-marketing.documenso.com/  (both signature and name)
https://stg-marketing.documenso.com/pricing  (when we click on signup now)
regular sign up page  (http://localhost:3000/signup)
profile page (http://localhost:3000/settings/profile)

https://github.com/documenso/documenso/assets/145282438/58ec4dca-0bfc-44ed-95db-51fbbd0ab394

PR #558, opened 2023-10-14
fix(bug): name field can be updated with spaces #555

### LINKED ISSUE #555: bug(name_field): name field can be update with "spaces_only"
I was trying to update names. There i found that name can be updated with spaces only!

## Issue Description

In profile section in name tab name can be updated with spaces only "        ".

## Steps to Reproduce

https://www.loom.com/share/fbb762fd704844caac63b88ef2f83e91?sid=3729b00c-b1c5-4df5-b7bb-09f13029f6aa

## Expected Behavior

It should trim the name and update it else should give same error which it gives when we try to update name without any character.

## Current Behavior
In profile section in name tab name can be updated with spaces only "        ".

## Environment

- OS: [Arch Linux]
- Browser: [Brave]
- Version: [v1.59.117]

## Checklist

adityadeshlahre:
**fixes #555** 

**I have fixed in two places in singup form and profile update page form.
this fix improves UX as It disabled the update profile / singup button when ever the fullname was emtpy .
as shown in video below !**

**[updated video]**

https://www.loom.com/share/f0591258760349bc9871675a480972f4?sid=7617e340-a854-4f72-b70e-2756594686b4

If any changes required let me know !
Thank You :smile:

### DIFF
--- a/packages/trpc/server/field-router/schema.ts
+++ b/packages/trpc/server/field-router/schema.ts
@@ -3,7 +3,7 @@ import { z } from 'zod';
 export const ZSignFieldWithTokenMutationSchema = z.object({
   token: z.string(),
   fieldId: z.number(),
-  value: z.string().refine((value) => value.trim() !== '' || value === ''),
+  value: z.string().trim(),
   isBase64: z.boolean().optional(),
 });
 
--- a/packages/trpc/server/field-router/schema.ts
+++ b/packages/trpc/server/field-router/schema.ts
@@ -3,7 +3,7 @@ import { z } from 'zod';
 export const ZSignFieldWithTokenMutationSchema = z.object({
   token: z.string(),
   fieldId: z.number(),
-  value: z.string(),
+  value: z.string().refine((value) => value.trim() !== '' || value === ''),
   isBase64: z.boolean().optional(),
 });
 
--- a/apps/web/src/app/(signing)/sign/[token]/form.tsx
+++ b/apps/web/src/app/(signing)/sign/[token]/form.tsx
@@ -99,7 +99,7 @@ export const SigningForm = ({ document, recipient, fields }: SigningFormProps) =
                   id="full-name"
                   className="bg-background mt-2"
                   value={fullName}
-                  onChange={(e) => setFullName(e.target.value)}
+                  onChange={(e) => setFullName(e.target.value.trimStart())}
                 />
               </div>
 
--- a/apps/web/src/app/(signing)/sign/[token]/name-field.tsx
+++ b/apps/web/src/app/(signing)/sign/[token]/name-field.tsx
@@ -45,26 +45,32 @@ export const NameField = ({ field, recipient }: NameFieldProps) => {
 
   const [showFullNameModal, setShowFullNameModal] = useState(false);
   const [localFullName, setLocalFullName] = useState('');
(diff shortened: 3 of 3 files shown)


dguyen:
The majority of these changes can be simplified down to a ZOD schema change.

By adding `trim()` to the zod schema field you want to disallow whitespaces for.

Example:

```
export const ZProfileFormSchema = z.object({
  name: z.string().trim().min(1),
  signature: z.string().min(1, 'Signature Pad cannot be empty'),
});
```

In addition, we don't want to disable the submit button when name is whitespace. We want the user to be able to click it to see the validation trigger.

adityadeshlahre:
Hey @dguyen sorry for over complicating changes in previous push !
at first commit i was doing that zod validation only but UX was not good thats thought of other change.
Again sorry for that.

Recently i found out that there are some issues with {SIGNING PAGE} Attached video

https://www.loom.com/share/19d24660898447819d97e6a3f9a05bb9?sid=19fabcdb-5b3e-4f8b-ac01-834f40f3c936


There are few issues 
1. user can sing on [signing-field] with "            " spaces only.
2. If name was not entered by default Then if user clicks on [signing-field] the popup-dialog-box open there also user can sign with "             " spaces only on press on [sign] button
3. Similarly user can send "           " spaces only to [signing-field] on X [close button] of the popup-dialog-box


Should i open new issue for above issue or fix in this only !

Thank You

dguyen:
Ah apologies, it appears that these fields are already being covered in the following PR #557 which was created before this.

If you find any other areas that aren't covered in the PR linked above, please go ahead and add them here

Otherwise this PR will probably be closed in favor of the other one

adityadeshlahre:
Hey @dguyen seem like this one is not covered !

This is performed on **[https://stg-app.documenso.com/]**

https://www.loom.com/share/d85cdd3df9c8479ebf3156ef47b13f84?sid=d5eae026-450e-48da-948d-7134a1129afa

dguyen:
Please feel free to fix those up in this PR :)

adityadeshlahre:
fixed the DIALOG box.
user cannot submit " " spaces as name !
If the input field contains only spaces, the form submission will be prevented.

https://www.loom.com/share/65614afc957e4ed2981b411e86a9978f?sid=ce4c8465-ad24-40fa-89d9-efecee993e53

dguyen:
I think it would be more clean if you just trimmed the start of the input for `name-field.tsx`:

```
  onChange={(e) => setLocalFullName(e.target.value.trimStart())}
```

And the name input field in `apps/web/src/app/(signing)/sign/[token]/form.tsx`:

```
  onChange={(e) => setFullName(e.target.value.trimStart())}
```

Which should prevent any whitespace only names

And then also update the `ZSignFieldWithTokenMutationSchema` schema so it trims the `value` field.

packages/trpc/server/field-router/schema.ts
```diff
@@ -3,7 +3,7 @@ import { z } from 'zod';
 export const ZSignFieldWithTokenMutationSchema = z.object({
   token: z.string(),
   fieldId: z.number(),
-  value: z.string(),
+  value: z.string().refine((value) => value.trim() !== '' || value === ''),
```
dguyen, line 6: You should be able to just do `z.string().trim()`

dguyen:
Looks good to me! Thanks for this :)

/tip 25

PR #563, opened 2023-10-14
chore: update zod to 3.22.4

adithyaakrishna:
Description:

- This PR updates zod to `3.22.4` as the previous versions had ReDoS vulnerability

More Info: https://github.com/colinhacks/zod/pull/2824 and https://github.com/colinhacks/zod/issues/2609


catalinpit (approved):
thanks!

dguyen (changes requested):
Could you also upgrade them across the other packages? I think since a recent commit there were a few other packages that have zod now.

adithyaakrishna:
@dguyen Yes sure, will send a commit for that too :)

adithyaakrishna:
@dguyen Done, I updated `next-auth` as well, also, I was wondering if I could just update all the packages instead of only Zod?

dguyen:
What was the reason for upgrading next-auth? We prefer to try stay away from unnecessary updates since they generally break things and requires testing.

In this case I think upgrading Zod only is fine since there's a vulnerability.

adithyaakrishna:
@dguyen Makes sense, I have reverted the change to `next-auth` with the above commit :)

dguyen:
Something in the package lock file is breaking the build.

When I delete it and re-run everything it works.

Could you undo everything, and do the Zod upgrade incrementally so the package lock file is correctly updated?

dguyen:
Cheers 👍 

/tip 25

PR #566, opened 2023-10-15
feat: plan limits

Mythie:
Adds limits to plans to allow for a free plan that has limited functionality.

Screenshots:

image
image


packages/prisma/schema.prisma
```diff
@@ -31,7 +31,7 @@ model User {
   accounts           Account[]
   sessions           Session[]
   Document           Document[]
-  Subscription       Subscription[]
+  Subscription       Subscription?
```
dguyen, line 34: Would this need a migration?

Mythie: No this is just related to code generation, since we already have a unique constraint on `Subscription.userId` there are no other migrations required.

dguyen (approved):

Issue #577, opened 2023-10-18
[BUG] Loading State in Wrong Subscribe Button 


PeterKwesiAnsah:
## Issue Description
I tried to subscribe to a plan. Upon clicking on a subscribe button, two loading states were triggered instead of one.

## Steps to Reproduce

1. Go To the Billing Page in Settings
2. Click on a subscribe button in a Plan card

## Expected Behavior

One loading state

## Current Behavior
Two loading  button states show

## Screenshots (optional)

![Screenshot (78)](https://github.com/documenso/documenso/assets/31078314/7463c9cc-1fc6-49f1-89ae-b5361fe618cd)

## Environment

- OS: Linux
- Browser: Chrome

## Checklist

- [x] I have searched the existing issues to make sure this is not a duplicate.
- [x] I have provided steps to reproduce the issue.
- [x] I have included relevant environment information.
- [x] I have included any relevant screenshots.
- [x] I understand that this is a voluntary contribution and that there is no guarantee of resolution.

mittalsam98:
I would like to pick this up

PeterKwesiAnsah:
@mittalsam98 I'm already taking a swing at it

Mythie:
One thing to note and a reason why I left this the way it is, we don't want a user to click a second subscription button while a checkout is loading. This means that while loading all buttons should be disabled.

catalinpit:
Hey @PeterKwesiAnsah, thanks for opening the issue & PR! However, we recently updated the contribution guide, which states that you should `Comment on the issue with your intention to work on it` and get it assigned to you before starting to work on it.

In this case, @mittalsam98 should've been assigned the issue.

The reason for this update is to bring more order and improve the contribution process. Thanks for understanding!

![IMG_7832](https://github.com/documenso/documenso/assets/25515812/dd3e9fda-5d04-4436-9ab4-86aa715ad773)
*Source: [Contribution Guide #taking-issues](https://github.com/documenso/documenso/blob/feat/refresh/CONTRIBUTING.md#taking-issues)*

PeterKwesiAnsah:
@catalinpit Got it. Thanks Man!

mittalsam98:
Hey @catalinpit , thanks for assigning this to me. I will try to raise a fix ASAP

catalinpit:
Hey @mittalsam98, the issue reporter already started working on this. Normally, you would have priority since you asked first, but can we leave it like this for this one, please? To avoid confusion.

[...]

PR #580, opened 2023-10-19
chore: update bug report template

dguyen:
## Description

Update the GitHub report template checklist to include whether the issue author wants to handle creating a PR for the issue if approved. 

## Additional Notes

Not actually sure if this is a good spot to put it


dephraiim:
Due to this, https://github.com/documenso/documenso/issues/577#issuecomment-1776597354, I think this is okay and it should be merged else when I create an issue, I will have to state again that I want to work on it.

PR #581, opened 2023-10-19
feat: migrated to signpdf from node-signpdf

adithyaakrishna:
**Description:**

- `node-signpdf` is deprecated according to https://www.npmjs.com/package/node-signpdf and is replaced by `signpdf` with each package now being a separate one and hence, this PR updates the same :)


dephraiim (approved):

Mythie:
Pausing this, the deprecation was only recently and messes with a few moving parts 😄

Mythie:
Closing, I can reopen later against main when we're ready

PR #584, opened 2023-10-20
fix: change from forgotten to forgot password

anikdhabal:
Forgotten your password is not a correct phrase. So, I changed it to 'forgot' in dev mode.
Screenshot 2023-10-20 101525


dguyen (approved):
Cheers for that catch!

PR #585, opened 2023-10-21
fix: disable update profile and password button when field value is not changed or updated

anikdhabal:
The 'Update Profile' & 'Update Password' button is currently clickable even when no changes have been made to the fields. It should be disabled if any field values remain unchanged or not updated.

Now:-

Screenshot 2023-10-21 153451

Screenshot 2023-10-21 230938

### DIFF
--- a/apps/web/src/app/(unauthenticated)/forgot-password/page.tsx
+++ b/apps/web/src/app/(unauthenticated)/forgot-password/page.tsx
@@ -5,7 +5,7 @@ import { ForgotPasswordForm } from '~/components/forms/forgot-password';
 export default function ForgotPasswordPage() {
   return (
     <div>
-      <h1 className="text-4xl font-semibold">Forgotten your password?</h1>
+      <h1 className="text-4xl font-semibold">Forgot your password?</h1>
 
       <p className="text-muted-foreground mt-2 text-sm">
         No worries, it happens! Enter your email and we'll email you a special link to reset your
--- a/apps/web/src/app/(unauthenticated)/forgot-password/page.tsx
+++ b/apps/web/src/app/(unauthenticated)/forgot-password/page.tsx
@@ -5,7 +5,7 @@ import { ForgotPasswordForm } from '~/components/forms/forgot-password';
 export default function ForgotPasswordPage() {
   return (
     <div>
-      <h1 className="text-4xl font-semibold">Forgot your password?</h1>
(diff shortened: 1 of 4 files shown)


adithyaakrishna:
@anikdhabal Could you please check the same for the password page as well?

anikdhabal:
> @anikdhabal Could you please check the same for the password page as well?

Ok let me fix that.

dephraiim (approved):

adithyaakrishna:
I do have second thoughts about this, in enterprises, we usually don't prevent any users from doing any action which in this is clicking the button if the fields haven't changed.

What we do at Red Hat is something like,

- The button is clickable/enabled even though fields haven't changed
- Let's say a user clicks the button without changing anything, then we show an error message under the fields or an error ring to the input fields, following that, we disable the button

@dephraiim @dguyen @Mythie would love to know your opinion on this 👀

anikdhabal:
> I do have second thoughts about this, in enterprises, we usually don't prevent any users from doing any action which in this is clicking the button if the fields haven't changed.
> 
> What we do at Red Hat is something like,
> 
> * The button is clickable/enabled even though fields haven't changed
> * Let's say a user clicks the button without changing anything, then we show an error message under the fields or an error ring to the input fields, following that, we disable the button
> 
> @dephraiim @dguyen @Mythie would love to know your opinion on this 👀

I don't think so. It's not a major issue or bug to add any extra things.It was a simple fix that prevent to show the toast message every time whenever click the button without any change.
Org like cal.com does this same thing.

catalinpit:
Hey @anikdhabal, thanks for the PR! The behavior is fine as it is for now. Disabling the button might confuse people since people are used to profile pages having the button enabled. If the button is disabled, people might get the idea that they can't update their profile or something similar.

As an example, check GitHub. If you go to the profile page, you can see that the button is enabled even if you don't change anything.

Thanks a lot for the PR once again! It looks good, and we would've merged it for sure if we decided to disable the button.

anikdhabal:
> Hey @anikdhabal, thanks for the PR! The behavior is fine as it is for now. Disabling the button might confuse people since people are used to profile pages having the button enabled. If the button is disabled, people might get the idea that they can't update their profile or something similar.
> 
> As an example, check GitHub. If you go to the profile page, you can see that the button is enabled even if you don't change anything.
> 
> Thanks a lot for the PR once again! It looks good, and we would've merged it for sure if we decided to disable the button.

Thanks for your review. Not an issue .

PR #588, opened 2023-10-23
chore: use minio as s3 storage for document during development

dephraiim:


docker/compose-services.yml
```diff
@@ -17,3 +17,20 @@ services:
       - 9000:9000
       - 2500:2500
       - 1100:1100
+
+  minio:
+    image: minio/minio
+    container_name: s3
+    ports:
```
catalinpit, line 24: Maybe we could have 2 comments such as:

- 9002:9002 # API
- 9001:9001 # Console

Not super important, so you can disregard it. I just thought that it might help people quickly identify which one is what.

docker/compose-services.yml
```diff
@@ -17,3 +17,20 @@ services:
...
+      - 9002:9002
+      - 9001:9001
+    volumes:
+      - minio:/data
+    environment:
+      MINIO_ROOT_USER: documenso
+      MINIO_ROOT_PASSWORD: password
+    entrypoint: sh
+    command: -c 'mkdir -p /data/documenso && minio server /data --console-address ":9001" --address ":9002"'
```
catalinpit, line 33: 1. is the `entrypoint` field necessary?
2. is this part `-c 'mkdir -p /data/documenso && minio` from the `command` necessary? doesn't it work without it?

I'm asking because I don't have much Docker experience, so I'm trying to understand. :-D

catalinpit (approved):

docker/compose-services.yml
dephraiim: Yes and yes, `entrypoint` allows me to run the command after the container has been initiated. The command allows me to create a bucket with name `documenso` else the user would have to create the bucket manually which is easier to miss and you get a lot of errors if you don't have the bucket available

.env.example
```diff
@@ -22,17 +22,17 @@ E2E_TEST_AUTHENTICATE_USER_PASSWORD="test_password"
...
+NEXT_PRIVATE_UPLOAD_REGION="unknown"
 # REQUIRED: Defines the bucket to use for the S3 storage transport.
-NEXT_PRIVATE_UPLOAD_BUCKET=
+NEXT_PRIVATE_UPLOAD_BUCKET="documenso"
 # OPTIONAL: Defines the access key ID to use for the S3 storage transport.
-NEXT_PRIVATE_UPLOAD_ACCESS_KEY_ID=
+NEXT_PRIVATE_UPLOAD_ACCESS_KEY_ID="documenso"
 # OPTIONAL: Defines the secret access key to use for the S3 storage transport.
-NEXT_PRIVATE_UPLOAD_SECRET_ACCESS_KEY=
+NEXT_PRIVATE_UPLOAD_SECRET_ACCESS_KEY="password"
```
Mythie, line 35: I think we can stick with database for the example env but retain this as an option for those interested in exploring the universal upload.

dephraiim: Okay, I'll revert back to the database

docker/compose-services.yml
dephraiim: Going to maintain it like this for now, else we will have to add comments to all the other ports for the mailserver and the database

Mythie (approved):

Issue #470, opened 2023-09-29
BUG Signer fields cannot be edited/removed after modifying the signer.


kandelin16:
<img width="1211" alt="image" src="https://github.com/documenso/documenso/assets/78173482/0b47f36a-106b-48b3-9297-5a70fd9b8315">

In this photo `employee@farmhand.pro` was changed to `hyungseok5926@gmail.com`, but the old `employee@farmhand.pro` fields did not change, nor can they be edited at this point.

However, if I delete the modified email the fields are all removed.

To Replicate
1. Create Signer 
2. Create Fields 
3. Go back and change email address of signer 
4. The fields of the old email are still there and cannot be edited or removed

18feb06:
Sounds interesting...Working on this

dguyen:
hey @kandelin16, thanks for finding this!

Currently the data is actually being successfully updated in the backend, which we can tell by refreshing the page and seeing the fields being updated to the correct value.

I looked a bit into this and found that it's likely occurring because:

1. Between the steps we use `router.refresh` to get the updated fields/recipients
2. Since this occurs in the background, the `localFields` state in the `AddFields` component is not updated with the new values

There's a few ways to fix this:

1. Set the key of the `AddFieldsFormPartial` component to change based on the recipients & fields array
2. Update the logic of the `EditDocument` flow to store a local copy of the recipient/fields which is refreshed inbetween each step(so don't rely on router.refresh, this will stop the minor flickering between steps if things change)
3. Update `AddFields` component to update the local fields when recipients changes

Personally I'm in favor of the second option, but that would require the most changes out of the 3.

What do you guys think @catalinpit @dephraiim? Also welcome to suggestions from the community.

Abhinav-Developer-23:
@dguyen we can add `  await new Promise((resolve) => {
        setTimeout(resolve, 500);
      });`
      after router.refresh();  to wait for 500 ms . this method is working, I checked in the local. it does not cause a noticeable delay on the UI. It also requires minimal code changes

alfredlouisw8:
hi can i work on this issue?

ElTimuro:
> @dguyen we can add ` await new Promise((resolve) => { setTimeout(resolve, 500); });` after router.refresh(); to wait for 500 ms . this method is working, I checked in the local. it does not cause a noticeable delay on the UI. It also requires minimal code changes

Could work, but timeout are always hacky and carry a risk of not working in slow environments

PR #596, opened 2023-10-26
fix: signer fields cannot be edited/removed after modifying the signer

alfredlouisw8:
fixes: #470 

The approach I use is using a state variable to store the recipients and signer so that it changes on every step

any feedback for the code is appreciated! :)


alfredlouisw8:
Sure!

catalinpit (approved):
Looks good to me. But let's wait and see what @Mythie or @dguyen or @dephraiim says as well.

dguyen:
Hi @alfredlouisw8, apologies for the delay

This should be resolved as part of #1039 which uses a similar solution to yours, so we appreciate your initial investigation on this topic 🙏

PR #600, opened 2023-10-27
feat: improving incorrect default signature behaviour #594

### LINKED ISSUE #594: [improvement] incorrect default signature
## Issue Description

Suppose a user has 2 emails. They are signed in with email1 on documenso and email2 on Google Mail. When they receive an email to sign a document on email2, the default signature on the signature pad is from email1.signature

## Steps to Reproduce
1. create a signer using email1
2. add email2 in the recipient
3. sign in to Gmail using email2
4. open the link. The default signature will be of email1
## Expected Behavior

Signature pad should email empty

## Current Behavior

We get the signature of email1

## Screenshots (optional)

https://github.com/documenso/documenso/assets/145282438/dd99d0f8-b0c4-4cf8-b35c-79038ace68c9

## Environment

- OS: ubuntu 22
- Browser:Chrome

## Checklist

18feb06:
This PR fixes #594 .

This PR improves default signature behaviour so that when a user is signed in, if signer's email is not same as logged in user's email, then signature pad will be empty.


catalinpit (changes requested):
Nice! Could you also update the `value` for the "Full Name" input field, please?

Something like this:

```
value={session && session.user.email === recipient.email ? fullName : undefined}
```

At the moment, it shows the name of the recipient even you're logged in with another user.

Thanks!

18feb06:
@catalinpit change commited

Mythie:
I'd advise looking into the layout for signing which contains the provider for name/signature etc, we should be able to avoid the current changes by changing above the provider :)

18feb06:
@Mythie sounds great

apps/web/src/app/(signing)/sign/[token]/page.tsx
```diff
@@ -56,7 +56,11 @@ export default async function SigningPage({ params: { token } }: SigningPageProp
   }
 
   return (
-    <SigningProvider email={recipient.email} fullName={recipient.name} signature={user?.signature}>
+    <SigningProvider
+      email={recipient.email}
+      fullName={user?.email === recipient.email ? recipient.name : undefined}
```
Comment (Mythie) [inline apps/web/src/app/(signing)/sign/[token]/page.tsx:61]: So for this one we probably want something like `user?.email === recipient.email ? user.name : recipient.name` since we should use the name that the author has specified while allowing the recipient to change it if necessary.

Mythie:
Thanks for taking the suggested approach, it looks much cleaner now!

apps/web/src/app/(signing)/sign/[token]/page.tsx
Comment (18feb06) [reply apps/web/src/app/(signing)/sign/[token]/page.tsx:61]: ok.. no problem

Mythie:
LGTM!
cat-button-explosion

PR #613, opened 2023-10-31
fix: does not render cancel button if history length is <=1

### LINKED ISSUE #604: [Bug] Cancel button does nothing
Cancel Button on the document signing page does nothing

## Steps to Reproduce

1. Go to the received mail for document signing
2. Open the link
3. Click on cancel.

## Expected Behavior

It should close the page

## Current Behavior

Nothing happens

## Screenshots (optional)

https://github.com/documenso/documenso/assets/145282438/a7b01d49-3a3d-4347-b4e6-12f9e7172da4

## Environment

- OS: Ubuntu
- Browser: Chrome

## Checklist

Abhinav-Developer-23:
fixes #604 

https://github.com/documenso/documenso/assets/145282438/74872c3e-22b7-47f5-a13e-7dbbe56f0803


Mythie:

https://documen.so/cla

apps/web/src/app/(signing)/sign/[token]/form.tsx
```diff
@@ -121,16 +121,17 @@ export const SigningForm = ({ document, recipient, fields }: SigningFormProps) =
...
-                type="button"
-                className="dark:bg-muted dark:hover:bg-muted/80 w-full  bg-black/5 hover:bg-black/10"
-                variant="secondary"
-                size="lg"
-                disabled={typeof window !== 'undefined' && window.history.length <= 1}
-                onClick={() => router.back()}
-              >
-                Cancel
-              </Button>
+              {window.history.length > 1 && (
```
Comment (Mythie) [inline apps/web/src/app/(signing)/sign/[token]/form.tsx:124]: `typeof window` needs to be checked to avoid SSR issues, realistically this shouldn't be any different than the disable statement above so it should still fail in the same situations the issue reported.

Comment (Abhinav-Developer-23) [reply apps/web/src/app/(signing)/sign/[token]/form.tsx:124]: I guess it will always be disabled in Gmail because Gmail always opens the document in a new tab and it has no history. 
I removed `typeof window` because in dev server, it was showing error "Error: Text content does not match server-rendered HTML.Warning: Text content did not match. Server: "Complete" Client: "Cancel"
"

Comment (Mythie) [reply apps/web/src/app/(signing)/sign/[token]/form.tsx:124]: That would be the intention, it should be disabled for people opening from the email while those accessing it via the dashboard (inbox) would have the ability to cancel and go back.

Again I still think disabling is the way to go here, otherwise to handle this in an SSR friendly manner you'll need to use state and an effect to deal with hydration mismatches.

Comment (Abhinav-Developer-23) [reply apps/web/src/app/(signing)/sign/[token]/form.tsx:124]: Okay, I'll leave it as it is. I initially thought I needed to resolve this issue since it was assigned to me. I'll close this pull request and the associated issue.

Issue #603, opened 2023-10-28
Document Sharing Link Redirects to Homepage


daallgeier:
## Issue Description

Opening a sharing link to a document throws a console error and redirects to documenso.com (example my pledge: https://app.documenso.com/share/sdwlb8c7t1pkvu) 

```
8593596e-8fa50b0cbb75b2d0.js:9 Uncaught Error: Minified React error #419; visit https://reactjs.org/docs/error-decoder.html?invariant=419 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at 8593596e-8fa50b0cbb75b2d0.js:9:63944
    at Qh (8593596e-8fa50b0cbb75b2d0.js:9:64832)
    at rx (8593596e-8fa50b0cbb75b2d0.js:9:120339)
    at Lk (8593596e-8fa50b0cbb75b2d0.js:9:94328)
    at 8593596e-8fa50b0cbb75b2d0.js:9:94150
    at bf (8593596e-8fa50b0cbb75b2d0.js:9:94157)
    at mf (8593596e-8fa50b0cbb75b2d0.js:9:90680)
    at MessagePort.P (499-e5ab349bd0ad5c7b.js:25:1665)
(anonymous) @ 8593596e-8fa50b0cbb75b2d0.js:9
Qh @ 8593596e-8fa50b0cbb75b2d0.js:9
rx @ 8593596e-8fa50b0cbb75b2d0.js:9
Lk @ 8593596e-8fa50b0cbb75b2d0.js:9
(anonymous) @ 8593596e-8fa50b0cbb75b2d0.js:9
bf @ 8593596e-8fa50b0cbb75b2d0.js:9
mf @ 8593596e-8fa50b0cbb75b2d0.js:9
P @ 499-e5ab349bd0ad5c7b.js:25
```

## Steps to Reproduce

1. go to /documents 
2. Click the "share" button (wait for wayyy too long for it to be copied into clipboard)
3. Paste into new (incognito) tab 
4. Get redirected to documenso.com

## Expected Behavior

I expected to see my signed supporter pledge

## Current Behavior

I am seeing the homepage

## Environment

- OS: macOS Sonoma
- Browser: Chrome
- Version: 118

## Checklist

- [x] I have searched the existing issues to make sure this is not a duplicate.
- [x] I have provided steps to reproduce the issue.
- [x] I have included relevant environment information.
- [x] I have included any relevant screenshots.
[...]

PR #615, opened 2023-10-31
fix: added the share btn in the UI and prewarm fetch

luckyklyist:
Working issue no : #603 

Added the share UI btn from the document-share-btn.tsx and fetch the opengraph part in the btn .

https://github.com/documenso/documenso/assets/35479077/5d6e7f17-9bb8-49ab-9786-88cab531f862


Mythie:

https://documen.so/cla

apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
```diff
@@ -174,6 +175,8 @@ export const DataTableActionDropdown = ({ row }: DataTableActionDropdownProps) =
         </DropdownMenuItem>
       </DropdownMenuContent>
 
+      <DocumentShareButton token={String(row.Recipient[0]?.token)} documentId={row.id} />
```
Mythie, line 178: This can just be `recipient.token` since we do a matching in the code above. No need to cast to string, we should update the share dialog to have token as an optional prop.

packages/ui/components/document/document-share-button.tsx
```diff
@@ -71,7 +71,7 @@ export const DocumentShareButton = ({ token, documentId, className }: DocumentSh
 
   const onTweetClick = async () => {
     let { slug = '' } = shareLink || {};
-
+    await fetch(`${process.env.NEXT_PUBLIC_WEBAPP_URL}/share/${slug}/opengraph`);
```
Mythie, line 74: Awesome! Mind dropping a comment as to why we're doing this so no one deletes it in the future thinking we're being silly?

We may also want this after our `!slug` check to ensure we always have a valid share link to prewarm.

luckyklyist: Sure, I've just added the changes.

packages/ui/components/document/document-share-button.tsx
```diff
@@ -71,7 +71,6 @@ export const DocumentShareButton = ({ token, documentId, className }: DocumentSh
 
   const onTweetClick = async () => {
     let { slug = '' } = shareLink || {};
-
     if (!slug) {
       const result = await createOrGetShareLink({
         token,
```
Mythie, line 76: Apologies if it wasn't clear, we can make the `token` prop that this component accepts optional. If you look at the `createOrGetShareLink` implementation you'll find it's optional there too defaulting to the authenticated user if available.

This means that we don't need to using nullish coalescing in the `data-table-action-dropdown` component 😄

luckyklyist: sure , just made the token props as optional .

Mythie:
Awesome work 🎉 Happy first contribution!

thank-you-stronk

luckyklyist:
Currently, in the dashboard, the UI looks like this for share. 
Screenshot 2023-11-02 at 10 24 47 AM
Should I make changes to replace the sharing element here, or is it fine as it is?
Screenshot 2023-11-02 at 10 25 34 AM

PR #617, opened 2023-11-01
fix: seal document for single player mode

catalinpit:


dguyen:
I think instead of `sealDocument` it might be better to use `signpdf` directly since `create-single-player-document.action` was originally intended to handle everything, i just forgot to slap in the `signpdf` when I first did it

Mythie:
> I think instead of `sealDocument` it might be better to use `signpdf` directly since `create-single-player-document.action` was originally intended to handle everything, i just forgot to slap in the `signpdf` when I first did it
> 
> Reasoning:
> 
> Everything in `sealDocument` besides `signpdf` appears to be redundant, including 4 DB requests

also relevant, feels good to reuse tho

dguyen (approved):

PR #624, opened 2023-11-02
test: add test for document flow

G3root:


packages/app-tests/e2e/test-auth-flow.spec.ts
```diff
@@ -10,11 +12,11 @@ test.use({ storageState: { cookies: [], origins: [] } });
 */
 test.describe.configure({ mode: 'serial' });
 
-const username = process.env.E2E_TEST_AUTHENTICATE_USERNAME;
-const email = process.env.E2E_TEST_AUTHENTICATE_USER_EMAIL;
-const password = process.env.E2E_TEST_AUTHENTICATE_USER_PASSWORD;
+const username = 'Test User';
```
dephraiim, line 15: Why are these hardcoded instead of using the environmental variables?

packages/app-tests/fixtures/fixtures.ts
```diff
@@ -0,0 +1,24 @@
+import { test as base } from '@playwright/test';
+import type { Page } from '@playwright/test';
+import fs from 'node:fs';
+import path from 'node:path';
+
+import { createUsersFixture } from './user';
+
+interface Fixtures {
```
dephraiim, line 8: Can you use type instead?

packages/app-tests/fixtures/user.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import { APP_BASE_URL } from '@documenso/lib/constants/app';
+import { createUser as createUser_ } from '@documenso/lib/server-only/user/create-user';
+import { prisma } from '@documenso/prisma';
+
+type TUser = Awaited<ReturnType<typeof createUser>>;
+
+type TUserFixture = ReturnType<typeof createUserFixture>;
+
+interface loginUsingApiParams {
```
dephraiim, line 12: Can you use type instead?

packages/app-tests/e2e/test-document-flow.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  test.afterEach(async ({ users }) => users.deleteAll());
+
+  test('user can upload documents', async ({ page, samplePdf }) => {
+    await page.goto('/documents');
+    await expect(page).toHaveURL('/documents');
+
+    await page.getByTestId('document-dropzone').setInputFiles({
+      buffer: samplePdf,
+      mimeType: 'application/pdf',
+      name: 'test.pdf',
```
dephraiim, line 27: Is there an existing `test.pdf` file?

dephraiim (changes requested):

packages/app-tests/e2e/test-auth-flow.spec.ts
G3root: Using single static credentials for all the tests is causing conflicts when some tests run in parallel. Given that the user fixture generates a random user, we may not require those environment variables, so these values are hard coded only for this specific test.

packages/app-tests/e2e/test-document-flow.spec.ts
G3root: No it's just a placeholder name

packages/ui/primitives/document-dropzone.tsx
```diff
@@ -150,7 +150,7 @@ export const DocumentDropzone = ({
             </motion.div>
           </div>
 
-          <input {...getInputProps()} />
+          <input data-testid="document-dropzone" {...getInputProps()} />
```
catalinpit, line 153: Although testing by test IDs is the most resilient way of testing, I'm still not very sure about this approach for all tests.

The reason is that testing by test IDs is not user-facing, which means it doesn't resemble how the app is used by the users. Since these are created for testing only, the users don't see and interact with them.

Playwright recommends using this approach when you can't locate by role or text.

[Source](https://playwright.dev/docs/locators#locate-by-test-id)

Looking forward to seeing what others think as well.

packages/app-tests/e2e/test-auth-flow.spec.ts
catalinpit: Makes sense. I think I've seen the same thing in other repos as well.

packages/app-tests/e2e/test-auth-flow.spec.ts
```diff
@@ -10,11 +12,11 @@ test.use({ storageState: { cookies: [], origins: [] } });
...
+const email = 'testuser@email.com';
+const password = 'test_password';
 
-test('user can sign up with email and password', async ({ page }: { page: Page }) => {
+test('user can sign up with email and password', async ({ page }) => {
```
catalinpit, line 19: why remove the types?

packages/app-tests/e2e/test-document-flow.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  { email: 'signer1@gmail.com', name: 'signer 1' },
+  { email: 'signer2@gmail.com', name: 'signer 2' },
+  { email: 'signer3@gmail.com', name: 'signer 3' },
+];
+
+test.describe('test document flow', async () => {
+  test.beforeEach(async ({ users }) => {
+    const user = await users.create();
+    await user.apiLogin();
+  });
```
catalinpit, line 17: empty line after `beforeEach`, please.

packages/app-tests/e2e/test-document-flow.spec.ts
catalinpit: why not use the pdf name as the name? instead of using `test.pdf`?

packages/app-tests/e2e/test-document-flow.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    await page.goto('/documents');
+    await expect(page).toHaveURL('/documents');
+
+    await page.getByTestId('document-dropzone').setInputFiles({
+      buffer: samplePdf,
+      mimeType: 'application/pdf',
+      name: 'test.pdf',
+    });
+
+    await expect(page).toHaveURL(/\/documents\/\d+/);
```
catalinpit, line 43: lines 34-43 are repeated in 2 tests. I suppose they'll repeat again if we add another test for this page. Can we re-use them somehow or re-structure the tests?

packages/app-tests/e2e/test-document-flow.spec.ts
G3root: used pdf's name as the name

packages/app-tests/e2e/test-auth-flow.spec.ts
G3root: now the test function is typed, no need to typecast explicitly

packages/app-tests/e2e/test-document-flow.spec.ts
G3root: done 👍🏼

packages/app-tests/e2e/test-document-flow.spec.ts
G3root: now refactored

G3root:
closing in favour of #745

Issue #601, opened 2023-10-27
Issue with Document Listing After Uploading


poorna-prakash-sr:
## Issue Description

After uploading a document and clicking the back button without making any changes to the list, the recently uploaded document does not immediately appear. It either requires a manual refresh or it takes some time before the document appears in the list.

## Steps to Reproduce

1. Navigate to the 'Documents' section.
2. Upload a new document.
3. Once the document is successfully uploaded, click the back button without making any alterations.
4. Observe that the recently uploaded document is not present in the list.

## Expected Behavior

I expected that after uploading a document and navigating back, the document would be immediately listed.

## Current Behavior

The recently uploaded document is not immediately visible. It requires a manual page refresh or appears after some delay.

## Screenshots

https://www.awesomescreenshot.com/video/21964541?key=c1302a53ca4baefd487abdf7774e1327

## Environment

- OS: Ubuntu 22
- Browser: Chrome

## Checklist

- [x] I have searched the existing issues to make sure this is not a duplicate.
- [x] I have provided steps to reproduce the issue.
- [x] I have included relevant environment information.
- [x] I have included any relevant screenshots.
- [x] I understand that this is a voluntary contribution and that there is no guarantee of resolution.

mittalsam98:
Hey Can I work on this?

catalinpit:
> Hey Can I work on this?

Please go ahead. Thanks!

mittalsam98:
Thansk @catalinpit I will soon raise a PR for this

Mythie:
Just noting that using `router.refresh()` alongside a a `router.push()` will cause some navigation weirdness where the push may get cancelled.

ephraimduncan:
Closing this as it has been fixed.

PR #629, opened 2023-11-05
fix: document listing not updating after uploading

mittalsam98:
fixes : #601 

Issue title : Issue with Document Listing After Uploading

Fixes loom video
https://www.loom.com/share/04cf2bbb7d24446ca17b9647b4d9112c

Changes made : Setting prefetching to false in Link

Docs : https://nextjs.org/docs/pages/api-reference/components/link

Another option is to use `router.refresh()`, would love to hear any feedback from anyone

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/[id]/page.tsx
+++ b/apps/web/src/app/(dashboard)/documents/[id]/page.tsx
@@ -55,7 +55,11 @@ export default async function DocumentPage({ params }: DocumentPageProps) {
 
   return (
     <div className="mx-auto -mt-4 w-full max-w-screen-xl px-4 md:px-8">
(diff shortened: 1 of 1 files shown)


dguyen:
Just noting this won't resolve the issue if the user presses the browser's back button instead of clicking the link to go back

mittalsam98:
yeah it won't work when we will click on browser back button most probably , missed this case, I will check this and fix it ASAP

Mythie:
Re-closing on account of navigating to the document post-upload. Admittedly next.js navigations can be slow at times which leaves you on the page which hasn't refreshed but this isn't such a large issue that I'm overly concerned.

Sorry for the delay and we appreciate the effort! 💪

PR #640, opened 2023-11-08
fix: improve the validation message for documenso app

### LINKED ISSUE #635: improvement: fix the validation message for documenso app
### Issue Description

There are some input fields where I believe improvements are needed for the error messages displayed when an input field is empty, and someone attempts to proceed to the next step.

1.) In the Sign-In form and the reset password form, when the password field is empty, it shows the message "String must contain at least 6 character(s)."

2.) The email field in the "Add Signer" form displays "String must contain at least 1 character(s)."

### Steps to Reproduce

_No response_

### Expected Behavior

1.)  When the password field is empty, the error message should be "Invalid Password," similar to the email field.

2.) Instead of "String must contain at least 1 character(s)," the email field should display "Email is required," similar to how the name field behaves.

### Current Behavior

_No response_

### Screenshots (optional)

Current behavior:-

Screenshot 2023-11-08 131922

Screenshot 2023-11-08 132454

Screenshot 2023-11-08 145659

_No response_

_No response_

### Version [e.g., 2.0.1]

_No response_

- [ ] I have provided steps to reproduce the issue.

anikdhabal:
fixes #635 

Screenshot 2023-11-08 195029
Screenshot 2023-11-08 192253
Screenshot 2023-11-08 191351


apps/web/src/components/forms/signup.tsx
```diff
@@ -21,7 +21,7 @@ import { useToast } from '@documenso/ui/primitives/use-toast';
 export const ZSignUpFormSchema = z.object({
   name: z.string().trim().min(1, { message: 'Please enter a valid name.' }),
   email: z.string().email().min(1),
-  password: z.string().min(6).max(72),
+  password: z.string().min(6, { message: 'Password should contain at least 6 characters' }).max(72),
```
catalinpit, line 24: Since it has min and max, should the message be something like "Password should contain between 6 and 72 characters"?

apps/web/src/components/forms/password.tsx
```diff
@@ -20,9 +20,18 @@ import { FormErrorMessage } from '../form/form-error-message';
 
 export const ZPasswordFormSchema = z
   .object({
-    currentPassword: z.string().min(6).max(72),
-    password: z.string().min(6).max(72),
-    repeatedPassword: z.string().min(6).max(72),
+    currentPassword: z
+      .string()
+      .min(6, { message: 'Password should contain at least 6 characters' })
```
catalinpit, line 25: Same comment as below for all the fields that have min and max.

apps/web/src/components/forms/signup.tsx
anikdhabal: Generallly we don't set password more then 72 characters. So, I think it's fine now.
What if we can add validation messages for both the min and max cases? For max- "Password should not contain more than 72 characters".

anikdhabal: @catalinpit what do you think?

catalinpit: You are right. Please add the message for the max value as well.

anikdhabal: Done

anikdhabal:
@catalinpit I've updated it.

catalinpit:
Looks good, thanks for the PR!

catalinpit (approved):

PR #648, opened 2023-11-12
feat: runtime env

Mythie:
Support runtime environment variables using server components.

This will mean docker images can change env vars for runtime as required.


Mythie:
Closing as the recent improvements to https://github.com/expatfile/next-runtime-env mean that I'll use that instead!

Issue #660, opened 2023-11-16
feat: Title of your own choice


tanayvaswani:
### Feature Description

- After uploading the document, title is set by default to the file name. 
- Adding a feature to change title while in draft mode. 

### Use Case

- This title will be shown at the Dashboard.
- Same title can be used while sending emails. 

### Proposed Solution

- It would be great if users can have the flexibility of putting a title of their own choice. 
- Can be done by allowing users to input title while in draft & disable the input once document is no longer a draft.

### Additional Context

- Once the document is changed from **Draft** to **Pending**, there will no longer be an option to change the title, to avoid confusions.  

### Please check the boxes that apply to this feature request.

- [X] I have searched the existing feature requests to make sure this is not a duplicate.
- [X] I have provided a detailed description of the requested feature.
- [X] I have explained the use case or scenario for this feature.
- [X] I have included any relevant technical details or design suggestions.
- [X] I understand that this is a suggestion and that there is no guarantee of implementation.

Mythie:
Assigned 😁

tanayvaswani:
Done #663 ✅

PR #663, opened 2023-11-19
feat: add/update title of the Document

tanayvaswani:
Feature to add or update title to the documents. #660 

I've added the feature which provides the flexibility to users to have the custom title to their documents. 

These changes adds an input to the step form, which handles the title. See #660 for more information. 

**Added some action, types. 
Added some changes to the tRPC schema and tRPC document router.
Added some changes to the UI components.**

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx
+++ b/apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx
@@ -4,7 +4,7 @@ import { useState } from 'react';
 
 import { useRouter } from 'next/navigation';
 
-import { DocumentData, Field, Recipient, User } from '@documenso/prisma/client';
+import { DocumentData, DocumentStatus, Field, Recipient, User } from '@documenso/prisma/client';
 import { DocumentWithData } from '@documenso/prisma/types/document-with-data';
 import { cn } from '@documenso/ui/lib/utils';
 import { Card, CardContent } from '@documenso/ui/primitives/card';
@@ -14,6 +14,8 @@ import { AddSignersFormPartial } from '@documenso/ui/primitives/document-flow/ad
 import { TAddSignersFormSchema } from '@documenso/ui/primitives/document-flow/add-signers.types';
 import { AddSubjectFormPartial } from '@documenso/ui/primitives/document-flow/add-subject';
 import { TAddSubjectFormSchema } from '@documenso/ui/primitives/document-flow/add-subject.types';
+import { AddTitleFormPartial } from '@documenso/ui/primitives/document-flow/add-title';
+import { TAddTitleFormSchema } from '@documenso/ui/primitives/document-flow/add-title.types';
 import {
   DocumentFlowFormContainer,
   DocumentFlowFormContainerHeader,
@@ -25,6 +27,7 @@ import { useToast } from '@documenso/ui/primitives/use-toast';
 import { addFields } from '~/components/forms/edit-document/add-fields.action';
 import { addSigners } from '~/components/forms/edit-document/add-signers.action';
 import { completeDocument } from '~/components/forms/edit-document/add-subject.action';
+import { addTitle } from '~/components/forms/edit-document/add-title.action';
 
 export type EditDocumentFormProps = {
   className?: string;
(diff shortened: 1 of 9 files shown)


tanayvaswani:
@Mythie Here is a draft PR with current implementation (clip on discord). Do let me know if this needs some modification. Also should we need to move this on **Add Signers'** step instead of having a dedicated step?

Mythie:
Will check out this branch and run it shortly!

apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx
```diff
@@ -48,30 +51,60 @@ export const EditDocumentForm = ({
...
+      title: 'Add Title',
+      description: 'Add the title to the document.',
+      stepIndex: 1,
+    },
     signers: {
       title: 'Add Signers',
       description: 'Add the people who will sign the document.',
-      stepIndex: 1,
+      stepIndex: 2,
+      onBackStep: () => setStep('title'),
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx:68]: We should also add a check here for the document status to ensure we aren't able to go back when we're out of draft mode.

apps/web/src/components/forms/edit-document/add-title.action.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { updateTitle } from '@documenso/lib/server-only/document/update-title';
+import type { TAddTitleFormSchema } from '@documenso/ui/primitives/document-flow/add-title.types';
+
+export type AddTitleActionInput = TAddTitleFormSchema & {
+  documentId: number;
+};
+
+export const addTitle = async ({ documentId, title }: AddTitleActionInput) => {
+  'use server';
```
Mythie, line 11: Need to check for userId here otherwise I could theoretically update anyone's document

packages/lib/server-only/document/update-title.ts
```diff
@@ -0,0 +1,20 @@
+'use server';
+
+import { prisma } from '@documenso/prisma';
+
+export type UpdateTitleOptions = {
+  userId?: number;
```
Mythie, line 6: This is a mandatory option and should be reflected in the prisma query as well 😄

packages/trpc/server/document-router/router.ts
```diff
@@ -113,6 +115,17 @@ export const documentRouter = router({
       }
     }),
 
+  setTitleForDocument: authenticatedProcedure
+    .input(ZSetTitleForDocumentMutationSchema)
+    .mutation(async ({ input, ctx }) => {
+      const { documentId, title } = input;
```
Mythie, line 121: Same here, you can get the user id from `ctx.user.id` 😄

packages/ui/primitives/document-flow/add-signers.tsx
```diff
@@ -225,7 +228,9 @@ export const AddSignersFormPartial = ({
         <DocumentFlowFormContainerActions
           loading={isSubmitting}
           disabled={isSubmitting}
-          onGoBackClick={documentFlow.onBackStep}
+          onGoBackClick={
```
Mythie, line 231: If we make the change in the place I suggested above in the review we can circumvent the logic here if we wish to.

Mythie:
I tried it both ways in terms of design and I agree that yours feels better, I'm sure there's still work we could do to improve that experience but it's for a later time.

Once the comments above have been addressed we should be good to go 💪

packages/ui/primitives/document-flow/add-signers.tsx
Mythie: That’s fine, right now it’s not doing that though since disabled is only depending on the submission state so I guess we’d need to change that

tanayvaswani: I've addressed the above review (**edit-document.tsx**, first comment) of checking document status to ensure we aren't able to go back when we are out of draft mode. It does the work perfectly.

apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx
Comment (tanayvaswani) [reply apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx:68]: Done ✅

apps/web/src/components/forms/edit-document/add-title.action.ts
tanayvaswani: Yeah 👍

Mythie:
Hey there, I think we're all good to go now! Think I just forgot to hit the approve and merge button when I was doing this the other day 😆

Mythie (approved):

Mythie:
Awesome stuff!

thanks-adventure-time

PR #671, opened 2023-11-22
fix: add healthcheck endpoint

Mythie:


dguyen (approved):

dephraiim:
can you remove this?

https://github.com/documenso/documenso/blob/fbbc3b89c3eceef2129b03a86ed66be7b316a32a/packages/trpc/server/router.ts#L10-L12

PR #686, opened 2023-11-25
feat: enhance posthog event tracking

dephraiim:


apps/web/src/app/(dashboard)/documents/upload-document.tsx
```diff
@@ -18,10 +19,12 @@ import { useToast } from '@documenso/ui/primitives/use-toast';
 
 export type UploadDocumentProps = {
   className?: string;
+  userId?: number;
 };
 
-export const UploadDocument = ({ className }: UploadDocumentProps) => {
+export const UploadDocument = ({ className, userId }: UploadDocumentProps) => {
```
Mythie, line 25: We could just `useSession` instead of prop drilling?

packages/lib/universal/get-feature-flag.ts
```diff
@@ -20,6 +18,10 @@ export const getFlag = async (
 ): Promise<TFeatureFlagValue> => {
   const requestHeaders = options?.requestHeaders ?? {};
 
+  if (!LOCAL_FEATURE_FLAGS[flag]) {
+    return LOCAL_FEATURE_FLAGS[flag];
+  }
+
```
Mythie, line 24: 🤔

apps/web/src/app/(dashboard)/documents/upload-document.tsx
dephraiim: Sure, I will update it

packages/lib/universal/get-feature-flag.ts
dephraiim: I'll remove this.  The billing feature is not disabled when `NEXT_PUBLIC_FEATURE_BILLING_ENABLED` is set as false and you have environmental variables for PostHog available. This was a work around to remove documents limits locally for me to avoid dealing with stripe

Mythie (approved):

Issue #620, opened 2023-11-02
Forget password Options for OAuth Users


luckyklyist:
### Issue Description

For OAuth users, there is no password set in the database. Therefore, the 'Forgot Password' page will not be of any use to them. There should be an option to implement the creation of a password for OAuth users if they haven't set one would be a good idea.

### Steps to Reproduce

Log in with the OAuth provider Google, go to the profile page, and click on the 'Password' tab.

### Expected Behavior

There should be a way for the OAuth user to set the password.

### Current Behavior

When trying to add the password, the current password for the OAuth user is null, so there won't be forget password functionality for the oauth user.

### Screenshots (optional)

<img width="1335" alt="Screenshot 2023-11-02 at 1 31 40 PM" src="https://github.com/documenso/documenso/assets/35479077/c2b15636-7fd2-43cd-baea-fa64b386be37">

<img width="424" alt="Screenshot 2023-11-02 at 1 39 51 PM" src="https://github.com/documenso/documenso/assets/35479077/0772c6b6-6016-438a-bd0b-cd4df5c88a60">

### Browser [e.g., Chrome, Firefox]

Brave

### Please check the boxes that apply to this issue report.

- [X] I have searched the existing issues to make sure this is not a duplicate.
- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I have included any relevant screenshots.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.

luckyklyist:
After the approval of this issue, I would like to work on it.

Mythie:
Can we confirm if the forgot password flow works? So when signed out click on forgotten password and work from there :)

luckyklyist:
Yes, the 'Forgot Password' functionality is working. OAuth users can reset their password from there.

Should I add another tab to the settings page for 'Forgot Password'? Would that be beneficial for the UX?
<img width="1323" alt="Screenshot 2023-11-02 at 3 57 58 PM" src="https://github.com/documenso/documenso/assets/35479077/f5e6c41e-1abd-4b88-9436-13a1a9087ed8">

Mythie:
Reserving for a write up tomorrow on options and complications

G3root:
[...]

PR #689, opened 2023-11-27
fix: fix password form validation and update password logic for oauth…

luckyklyist:
Issue no : #620 

Fixed up the password form validation and updated the password logic for OAuth in response to issue #620. Introduced isPasswordNull routes and improved the password update process by checking for null values in the password field.

https://github.com/documenso/documenso/assets/35479077/b0da4339-6d80-4d2b-b0d2-24edbdba0ee5

### DIFF
--- a/apps/web/src/components/forms/password.tsx
+++ b/apps/web/src/components/forms/password.tsx
@@ -63,11 +63,8 @@ export const PasswordForm = ({ className }: PasswordFormProps) => {
     resolver: zodResolver(ZPasswordFormSchema),
   });
 
-  console.log(errors, 'form errors');
-
   const { mutateAsync: updatePassword } = trpc.profile.updatePassword.useMutation();
   const isPasswordNull = trpc.profile.isPasswordNullUser.useQuery().data;
-  console.log(isPasswordNull);
 
   const onFormSubmit = async ({ currentPassword, password }: TPasswordFormSchema) => {
     try {
--- a/apps/web/src/components/forms/password.tsx
+++ b/apps/web/src/components/forms/password.tsx
@@ -21,9 +21,11 @@ import { FormErrorMessage } from '../form/form-error-message';
 export const ZPasswordFormSchema = z
   .object({
     currentPassword: z
-      .string()
-      .min(6, { message: 'Password should contain at least 6 characters' })
-      .max(72, { message: 'Password should not contain more than 72 characters' }),
+      .union([
+        z.string().min(6, { message: 'Password should contain at least 6 characters' }),
+        z.string().max(72, { message: 'Password should not contain more than 72 characters' }),
+      ])
+      .optional(),
     password: z
       .string()
       .min(6, { message: 'Password should contain at least 6 characters' })
@@ -58,20 +60,19 @@ export const PasswordForm = ({ className }: PasswordFormProps) => {
     reset,
     formState: { errors, isSubmitting },
   } = useForm<TPasswordFormSchema>({
-    values: {
-      currentPassword: '',
-      password: '',
-      repeatedPassword: '',
-    },
     resolver: zodResolver(ZPasswordFormSchema),
   });
 
+  console.log(errors, 'form errors');
+
   const { mutateAsync: updatePassword } = trpc.profile.updatePassword.useMutation();
+  const isPasswordNull = trpc.profile.isPasswordNullUser.useQuery().data;
+  console.log(isPasswordNull);
 
   const onFormSubmit = async ({ currentPassword, password }: TPasswordFormSchema) => {
     try {
       await updatePassword({
-        currentPassword,
+        currentPassword: currentPassword || null,
         password,
       });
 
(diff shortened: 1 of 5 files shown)


packages/lib/server-only/user/update-password.ts
```diff
@@ -3,50 +3,71 @@ import { compare, hash } from 'bcrypt';
 import { prisma } from '@documenso/prisma';
 
 import { SALT_ROUNDS } from '../../constants/auth';
+import { isPasswordNull } from './is-password-null';
 
 export type UpdatePasswordOptions = {
   userId: number;
   password: string;
-  currentPassword: string;
+  currentPassword?: string;
```
Mythie, line 11: `currentPassword?: string | null;` right?

packages/lib/server-only/user/update-password.ts
```diff
@@ -3,50 +3,71 @@ import { compare, hash } from 'bcrypt';
...
 };
 
 export const updatePassword = async ({
   userId,
   password,
   currentPassword,
 }: UpdatePasswordOptions) => {
+  console.log('userId', userId, 'update passowrd');
```
Mythie, line 19: 🤓

packages/lib/server-only/user/update-password.ts
```diff
@@ -3,50 +3,71 @@ import { compare, hash } from 'bcrypt';
...
   const user = await prisma.user.findFirstOrThrow({
     where: {
       id: userId,
     },
   });
 
-  if (!user.password) {
+  if (!user) {
     throw new Error('User has no password');
   }
```
Mythie, line 29: This is probably redundant now

packages/lib/server-only/user/update-password.ts
```diff
@@ -3,50 +3,71 @@ import { compare, hash } from 'bcrypt';
...
 
-  const isCurrentPasswordValid = await compare(currentPassword, user.password);
-  if (!isCurrentPasswordValid) {
-    throw new Error('Current password is incorrect.');
-  }
+  const isUserPasswordNull = await isPasswordNull({ id: userId });
```
Mythie, line 31: This would mean we're retrieving the user twice, there should be no need for that?

packages/lib/server-only/user/update-password.ts
```diff
@@ -3,50 +3,71 @@ import { compare, hash } from 'bcrypt';
...
+      const hashedNewPassword = await hash(password, SALT_ROUNDS);
 
-  return updatedUser;
+      const updatedUser = await prisma.user.update({
+        where: {
+          id: userId,
+        },
+        data: {
+          password: hashedNewPassword,
+        },
```
Mythie, line 67: This should be simplified to only need an `if` branch if the password is not null where we can check and throw otherwise we proceed with the update. 

Right now it's doing the work twice and is a bit more confusing since `else` branches become hard to follow.

packages/trpc/server/profile-router/router.ts
```diff
@@ -110,4 +111,15 @@ export const profileRouter = router({
...
+    try {
+      const isUserPasswordNull = await isPasswordNull({ id: ctx.user?.id as number });
+      return isUserPasswordNull;
+    } catch (err) {
+      throw new TRPCError({
+        code: 'BAD_REQUEST',
+        message: 'An error occurred while checking the user existence. Please try again.',
+      });
+    }
+  }),
```
Mythie, line 124: We can probably just retrieve this during the server render and eliminate the need for the tRPC endpoint.

That said I do appreciate that you're just using the ctx here for the userId and not introducing it as a param which could be abused, good work!

apps/web/src/components/forms/password.tsx
```diff
@@ -21,9 +21,11 @@ import { FormErrorMessage } from '../form/form-error-message';
...
   .object({
     currentPassword: z
-      .string()
-      .min(6, { message: 'Password should contain at least 6 characters' })
-      .max(72, { message: 'Password should not contain more than 72 characters' }),
+      .union([
+        z.string().min(6, { message: 'Password should contain at least 6 characters' }),
+        z.string().max(72, { message: 'Password should not contain more than 72 characters' }),
+      ])
+      .optional(),
```
Mythie, line 28: This still feels clunky, would it work as a nullable like you've done at the bottom of this PR?

Mythie (commented):
If we go one level up from this form we should be able to retrieve the user there and pass in a prop determining if we should require `currentPassword` 😄

apps/web/src/components/forms/password.tsx
luckyklyist: What Zod validations should I add there to make it work?
```
currentPassword: z
      .string()
      .min(6, { message: 'Password should contain at least 6 characters' })
      .max(72, { message: 'Password should not contain more than 72 characters' })
      .nullable(),
  ```
  This gives error

packages/trpc/server/profile-router/router.ts
luckyklyist: Do you mean creating the Next.js server component to retrieve data directly?

packages/lib/server-only/user/update-password.ts
luckyklyist: 😅

Mythie:
Closing for now as we (Documenso) missed the mark on this one, apologies on my end. If this is still an issue with the current application let me know and I'll fix it personally 💚

PR #692, opened 2023-11-28
perf(web, lib): do not await inside promise statements

snax4a:
- removes `await` keyword from promises passed to `Promise.all`

Using `await` inside a `Promise.all` will make the awaited Promise resolve first, and only after that the Promise.all will be called.  This means the promises are run serially. In these two cases, we probably don't want that


snax4a:
> 
> https://documen.so/cla

I have filled out the CLA.

Mythie (approved):

Mythie:
What a silly mistake on our end! I think this came in as part of linting change surrounding promises 😅 

Thanks so much and happy first contribution! 🎉 

captain-america-salute

PR #711, opened 2023-12-02
feat: Allow for the deletion of any document

### LINKED ISSUE #696: [DOC-372] Allow for the deletion of any document
### **Description**

Expand deletion functionality to include documents in PENDING and COMPLETED states, alongside existing draft deletion.

### **Improvement Description**

* Allow deletion of documents in any state (DRAFT, PENDING, COMPLETED).
* Implement varying confirmation requirements based on the document state.
* Soft delete documents using a `deletedAt` column.

### **Rationale**

We've had a couple of users complain that they are unable to delete an erroneous document once it's left the draft state, this has hindered their ability to use Documenso and doesn't match that of other providers. To improve user satisfaction and match our competition, we should improve our current handling of deletions. It also hinders the user from handling their data as they please.

### **Proposed Solution**

* Maintain simple deletion for drafts.
* Require explicit confirmation (e.g., typing "delete me") for deleting PENDING or COMPLETED documents.
* Email recipients for documents that are PENDING so they are aware that it has been canceled.
* Implement soft deletion for documents, retaining access in the inbox.

### **Alternatives**

* Introduce a temporary holding area for deleted documents before permanent deletion. (Recycle Bin)
* User-customizable settings for deletion confirmations.

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-372](https://linear.app/documenso/issue/DOC-372/allow-for-the-deletion-of-any-document)</sub>

ksushant6566:
Resolves #696


packages/trpc/server/document-router/router.ts
```diff
@@ -97,15 +99,18 @@ export const documentRouter = router({
...
-        const { id } = input;
+        const { id, status } = input;
 
         const userId = ctx.user.id;
 
-        return await deleteDraftDocument({ id, userId });
+        if (status === DocumentStatus.DRAFT) {
+          return await deleteDraftDocument({ id, userId });
+        }
+        return await deleteNonDraftDocument({ id, userId, status });
```
Mythie, line 113: Should we simplify this by merging the two and just returning early for draft documents? We typically don't introduce logic in our tRPC routers as doing so makes them less portable for other API types.

ksushant6566: oh okay, I'll just make a `deleteDocument({ id, userId, status })` function that merges the functionality

apps/web/src/app/(dashboard)/documents/deleteWithConfirmation.tsx
```diff
@@ -0,0 +1,26 @@
+import { useState } from 'react';
+
+export const DeleteWithConfirmation = () => {
```
Mythie, line 3: This file doesn't appear to be used? 🤔

packages/email/templates/document-cancel.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export type DocumentCancelEmailTemplateProps = Partial<TemplateDocumentCancelProps>;
+
+export const DocumentCancelTemplate = ({
+  inviterName = 'Lucas Smith',
+  inviterEmail = 'lucas@documenso.com',
+  documentName = 'Open Source Pledge.pdf',
+  assetBaseUrl = 'http://localhost:3002',
+}: DocumentCancelEmailTemplateProps) => {
+  const previewText = `${inviterName} has cancelled the document ${documentName}, you don't need to sign it anymore.`;
```
Mythie, line 16: Feels like a redundant template string since we don't change it based on any conditions? 🤔

packages/lib/server-only/document/delete-document.ts
```diff
@@ -0,0 +1,96 @@
...
+
+import { FROM_ADDRESS, FROM_NAME } from '../../constants/email';
+
+export type DeleteDocumentOptions = {
+  id: number;
+  userId: number;
+  status: DocumentStatus;
+};
+
+export const deleteDocument = async ({ id, userId, status }: DeleteDocumentOptions) => {
```
Mythie, line 19: Good stuff on making this change 😄

Mythie:
I think this is mostly missing interactions with what happens when a signer visits the signing page for a deleted document etc.

If a document was deleted do we redirect them? Show an error page? If we do either of those what do we do for completed documents with a valid token, should we just show the normal completion page?

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
```diff
@@ -10,41 +13,46 @@ import {
...
-    });
+      onOpenChange(false);
+    },
+  });
 
-  const onDraftDelete = async () => {
+  const onDelete = async () => {
     try {
-      await deleteDocument({ id });
+      await deleteDocument({ id, status });
```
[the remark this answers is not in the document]
ksushant6566: yeah you are right!, lemme just make these changes and push quickly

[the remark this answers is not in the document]
ksushant6566: wait we are toasting on successful deletes, see OnSuccess of `trpcReact.document.deleteDocument.useMutation`

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
```diff
@@ -55,6 +63,11 @@ export const DeleteDraftDocumentDialog = ({
     }
   };
 
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
ksushant6566: okay, changing this

apps/web/src/app/(dashboard)/documents/deleteWithConfirmation.tsx
ksushant6566: oops

packages/email/templates/document-cancel.tsx
ksushant6566: In my opinion, it just makes it more readable & future-proof. But lmk if you want it changed?

packages/lib/server-only/document/delete-document.ts
```diff
@@ -0,0 +1,96 @@
...
+    const document = await prisma.document.findUnique({
+      where: {
+        id: id,
+        userId,
+      },
+      include: {
+        Recipient: true,
+        documentMeta: true,
+      },
+    });
```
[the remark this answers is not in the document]
ksushant6566: I mean we can do that, but maybe this seems a bit more readable?

packages/lib/server-only/document/delete-document.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    if (!document) {
+      throw new Error('Document not found');
+    }
+
+    if (document.Recipient.length === 0) {
+      throw new Error('Document has no recipients');
```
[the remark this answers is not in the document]
ksushant6566: oh yes, if there are no recipients then we don't need to do anything

ksushant6566:
> I think this is mostly missing interactions with what happens when a signer visits the signing page for a deleted document etc.
> 
> If a document was deleted do we redirect them? Show an error page? If we do either of those what do we do for completed documents with a valid token, should we just show the normal completion page?

I think for completed documents we can just show the normal page & for pending documents we can disable actions and show a message that **_this document is no longer available to sign_**

wdyt?

packages/lib/server-only/document/delete-document.ts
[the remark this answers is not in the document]
Mythie: That's what I was thinking :)

packages/lib/server-only/document/delete-document.ts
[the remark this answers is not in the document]
Mythie: This will involve two round trips to the database while there's potential for the one above to do a join, it's negligible in terms of performance but could add up later on. Happy to leave for now for the sake of readability though 😄

packages/email/templates/document-cancel.tsx
Mythie: Happy to keep, just noting that it doesn't do much 😄 

I try to avoid designing for a possible future and instead design for the today and the potential throw away

Mythie:
I'll do a round of UI review with this now as well, I've only been stinging you on code so far 👀

Mythie:
This will also need some guards for recipients adding signatures when a document has since been deleted 😄

ksushant6566:
> This will also need some guards for recipients adding signatures when a document has since been deleted 😄

Done, added a check in `sign-field-with-token.ts` .

ksushant6566:
Here is a loom video of the solution :)

https://www.loom.com/share/8f28b9ad4bb64a70b33da9d59fd2a95b?sid=3da77147-be41-486a-af86-cc8fd09afce2

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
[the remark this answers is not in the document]
Mythie: Right I always get mixed up since we've changed things up a little there!

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
```diff
@@ -78,7 +102,13 @@ export const DeleteDraftDocumentDialog = ({
               Cancel
             </Button>
 
-            <Button type="button" loading={isLoading} onClick={onDraftDelete} className="flex-1">
+            <Button
+              type="button"
```
Mythie, line 106: This should be a `variant="destructive"` so we know we're nuking something

apps/web/src/app/(signing)/sign/[token]/complete/page.tsx
```diff
@@ -92,11 +99,18 @@ export default async function CompletedSigningPage({
               Everyone has signed! You will receive an Email copy of the signed document.
             </p>
           ))
-          .otherwise(() => (
-            <p className="text-muted-foreground/60 mt-2.5 max-w-[60ch] text-center text-sm font-medium md:text-base">
-              You will receive an Email copy of the signed document once everyone has signed.
-            </p>
-          ))}
+          .otherwise(() =>
```
Comment (Mythie) [inline apps/web/src/app/(signing)/sign/[token]/complete/page.tsx:102]: Same thing

ksushant6566:
Yo, I believe everything works now, but still do confirm if the logic below makes sense:

What documents users can see:

**own documents**
_All documents created by me, but not the ones I deleted._

**received documents**

Pending documents: 

- not deleted: Yes
- deleted but signed: Yes
- deleted, not signed (Inbox):  No

Completed documents: 

- not deleted: Yes
- deleted: Yes

Note: Only owners can delete documents, users can't delete received documents. That will involve lot more work. (Maybe open another issue?)

PS: thanks a lot for your help throughout 🙌

Mythie:
This all sounds correct to me, going to run through a few rounds of testing before I hit the merge button :)

Mythie (approved):

PR #712, opened 2023-12-02
feat: allow deletion of any document

### LINKED ISSUE #696: [DOC-372] Allow for the deletion of any document

crediblebilal:
fixes #696

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
+++ b/apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
@@ -32,6 +32,7 @@ import {
 } from '@documenso/ui/primitives/dropdown-menu';
 
 import { ResendDocumentActionItem } from './_action-items/resend-document';
+import { DeleteDocumentDialog } from './delete-document-dialog';
 import { DeleteDraftDocumentDialog } from './delete-draft-document-dialog';
 import { DuplicateDocumentDialog } from './duplicate-document-dialog';
 
@@ -57,10 +58,10 @@ export const DataTableActionDropdown = ({ row }: DataTableActionDropdownProps) =
   const isOwner = row.User.id === session.user.id;
   // const isRecipient = !!recipient;
   const isDraft = row.status === DocumentStatus.DRAFT;
-  // const isPending = row.status === DocumentStatus.PENDING;
+  const isPending = row.status === DocumentStatus.PENDING;
   const isComplete = row.status === DocumentStatus.COMPLETED;
   // const isSigned = recipient?.signingStatus === SigningStatus.SIGNED;
-  const isDocumentDeletable = isOwner && row.status === DocumentStatus.DRAFT;
+  const isDocumentDeletable = isOwner;
 
   const onDownloadClick = async () => {
     let document: DocumentWithData | null = null;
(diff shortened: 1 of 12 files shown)


packages/lib/server-only/document/delete-document.ts
```diff
@@ -0,0 +1,23 @@
...
+  const documentId = id;
+  await sendDeletedEmail({ documentId, userId });
+  return await prisma.document.update({
+    where: {
+      id,
+    },
+    data: {
+      deletedAt: new Date(),
+    },
+  });
```
Mythie, line 22: We wouldn't send an email for DRAFT documents since they would have never been sent :)

crediblebilal:
@Mythie  The above case is fixed. If there is any other change, let me know.

apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
```diff
@@ -160,13 +159,20 @@ export const DataTableActionDropdown = ({ row }: DataTableActionDropdownProps) =
...
           onOpenChange={setDeleteDialogOpen}
         />
+      ) : (
+        <DeleteDocumentDialog
+          id={row.id}
+          open={isDeleteDialogOpen}
+          onOpenChange={setDeleteDialogOpen}
+        />
       )}
+
```
Mythie, line 175: Feels like we could have just one dialog that conditionally requires the extra step based on the document status.

apps/web/src/app/(signing)/sign/[token]/page.tsx
```diff
@@ -55,6 +55,10 @@ export default async function SigningPage({ params: { token } }: SigningPageProp
     redirect(`/sign/${token}/complete`);
   }
 
+  if (document.deletedAt !== null) {
+    redirect(`/sign/${token}/deleted`);
+  }
```
Comment (Mythie) [inline apps/web/src/app/(signing)/sign/[token]/page.tsx:60]: Nice, I like that this occurs after the completion check so for normal recipients with a completed document there's nothing more they need to do

packages/lib/server-only/document/send-delete-email.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+  });
+
+  if (!document) {
+    throw new Error('Document not found');
+  }
+
+  if (document.status == DocumentStatus.DRAFT) {
+    throw new Error('Document is a draft state');
+  }
```
Mythie, line 36: Right I see so with the part I unresolved this would always throw an error for draft documents meaning you wouldn't be able to delete them.

crediblebilal:
@Mythie are there any other changes required?

packages/lib/server-only/document/send-delete-email.ts
Mythie: Now made redundant but could be kept for extra safety I suppose

packages/lib/server-only/document/find-documents.ts
```diff
@@ -55,23 +55,32 @@ export const findDocuments = async ({
       OR: [
```
Mythie, line 55: These are all correct as far as I can tell

Issue #703, opened 2023-12-01
[OSSHACK] Opt-In Semantic Search


Mythie:
## Improvement Description

Integrate an opt-in semantic search feature into Documenso, leveraging OpenAI embeddings. This feature aims to enhance document search capabilities by providing more contextually relevant results based on the semantic content of documents and the user's search query.

The semantic search will only be available to subscribed users or to all users if a self-hoster has enabled the feature flag.

Users will be able to opt-in to semantic search by enabling the feature in their profile settings. The feature will be disabled by default as users might not want to use it due to privacy concerns or other reasons.

## Rationale

Semantic search represents a significant advancement over traditional keyword-based search by understanding the context and meaning behind the text. It can greatly improve user experience by providing more accurate and relevant search results.

## Proposed Solution

- Use the OpenAI Ada model to generate document embeddings.
- Store the embeddings within our database under a new table using `pg_vector`.
- Use a background job to generate embeddings for all existing documents if a user opts-in to semantic search.
- Use a background job to generate embeddings for new documents as they are added to the database for an opted-in user.
- Update the find documents method to use the semantic search feature if a user has opted-in.

## Caveats

- We would like to use Trigger.dev for background jobs but we should also consider how we might also implement this without a third-party when self-hosting if possible.
- The entire feature should be completely opt-in for self-hosters since they have full control over their data. We do not want to force required third-party services on them.

## Alternatives

- Using a different embedding model, such as BERT or RoBERTa.
- Using a different search engine, such as Elasticsearch or Algolia.

## Additional Context

- The implementation must ensure data privacy and security, considering the sensitive nature of the documents.
- This is a deviation from our no mandatory third-party services policy therefore we must make this a progressive enahancement rather than a mandatory feature.

VishalMCF:
Let's say we are using our cron job for the data security issue. As per my analysis, we can do the following:-
Whenever the user opts in for the semantic search, it should trigger a job request (publish) which needs to be sent to a message queue (as it will be persistent).  Then we can have a consumer in the same application instance that listens to the message events and executes the job. In that, we can run background processes without the need to have a timely cron job running. We should have a consumer group that will include all the pods in the production so that one pod gets one event only.
Now here we need to take care of the failure scenario. If any failure happens then that doc again needs to be sent again to the queue. After processing the embeddings we can store it as you suggested but my approach focuses on how to avoid trigger.dev. 
[...]

PR #713, opened 2023-12-02
feat: add document search to the command menu

### LINKED ISSUE #702: [OSSHACK] Add document search to the command menu
## Improvement Description

Add a search feature to the command menu that allows users to search for documents by title or recipient email. The search feature should be accessible from any page on the platform and should return a list of documents matching the search query. 

The search results should be limited to 3-5 documents to maintain interface clarity. 

The search feature should include both sent and received documents.

## Rationale

As we were a new platform search was not an issue since most users had a small number of documents. However, as the number of documents increases, users are required to paginate through tables to find specific documents. This is inefficient and time-consuming. A search feature would allow users to quickly find the documents they are looking for.

## Proposed Solution

- Add a search feature in the command bar.
- Enable search by document title or recipient email.
- Limit search results to 3-5 to maintain interface clarity.
- Include both sent and received documents in search results.

## Alternatives

- A separate dedicated search page or section as an alternative to integrating with the command bar.
- Advanced search filters (like date, document type, etc.) for more specific searches. (We will explore this later)

ksushant6566:
resolves #702


apps/web/src/components/(dashboard)/common/command-menu.tsx
```diff
@@ -53,6 +55,34 @@ export function CommandMenu({ open, onOpenChange }: CommandMenuProps) {
   const [search, setSearch] = useState('');
   const [pages, setPages] = useState<string[]>([]);
 
+  const [searchResults, setSearchResults] = useState<
+    { label: string; path: string; recipents: string[] }[]
+  >([]);
+
+  const { isLoading: isSearchingDocuments, refetch: findDocuments } =
```
Mythie, line 62: Why not use the `data` property instead? Using onSuccess on queries will be deprecated with React Query 5 afaik

apps/web/src/components/(dashboard)/common/command-menu.tsx
```diff
@@ -113,15 +143,42 @@ export function CommandMenu({ open, onOpenChange }: CommandMenuProps) {
...
+          if (exists) return 1;
+
+          const result = searchResults.find(
+            (result) => result.label.toLocaleLowerCase() === value.toLocaleLowerCase(),
+          );
+
+          if (result?.recipents.some((recipent) => recipent.includes(search))) return 1;
+          return 0;
+        },
+      }}
```
Mythie, line 160: Wonder if we can cheat and just use an `sr-only` thing that contains the recipients so we don't have to add a custom filter fn 🤔

apps/web/src/components/(dashboard)/common/command-menu.tsx
ksushant6566: makes sense, I'll make change it

apps/web/src/components/(dashboard)/common/command-menu.tsx
ksushant6566: not sure what you mean by sr-only? can you please explain?

packages/lib/server-only/document/search-documents-with-keyword.ts
```diff
@@ -0,0 +1,40 @@
...
+        {
+          Recipient: {
+            some: {
+              email: {
+                contains: query,
+              },
+            },
+          },
+        },
+      ],
```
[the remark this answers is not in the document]
ksushant6566: cools, I'll check and make the necessary changes

apps/web/src/components/(dashboard)/common/command-menu.tsx
ksushant6566: added a new `value` property in searchResults

```
const newResults = searchDocuments?.map((document) => ({
        label: document.title,
        path: `/documents/${document.id}`,
        value: document.title + document.Recipient.map((recipient) => recipient.email).join(' '),
      }));
```

with this we don't need a custom filter function anymore

packages/lib/server-only/document/search-documents-with-keyword.ts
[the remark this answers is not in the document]
ksushant6566: although the `findDocuments` does have a filter by `term` but searching `term` in recipients email might be a lil confusing & unintuitive for others so I just made the changes in `SearchDocumentsWithKeyword` to include received documents as well.

apps/web/src/components/(dashboard)/common/command-menu.tsx
```diff
@@ -53,6 +62,41 @@ export function CommandMenu({ open, onOpenChange }: CommandMenuProps) {
...
+  //       label: document.title,
+  //       path: `/documents/${document.id}`,
+  //       value: document.title + document.Recipient.map((recipient) => recipient.email).join(' '),
+  //     }));
+  //     setSearchResults(newResults);
+  //   }
+  // }, [searchDocuments]);
+
+  const searchResults = useMemo(() => {
+    if (!searchDocuments) return [];
```
Mythie, line 91: No single line if statements please, just wrap it in braces `{}` 😄

packages/lib/server-only/document/search-documents-with-keyword.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const searchDocumentsWithKeyword = async ({
+  query,
+  userId,
+}: SearchDocumentsWithKeywordOptions) => {
+  const user = await prisma.user.findFirstOrThrow({
+    where: {
+      id: userId,
+    },
+  });
```
Mythie, line 16: Seems like we aren't using this?

packages/lib/server-only/document/search-documents-with-keyword.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const documents = await prisma.document.findMany({
```
Mythie, line 18: Looks like we're missing a limit here, if I were to search for `a` I might return thousands of results

Mythie: Ideally this would be an option passed into the method

packages/trpc/server/document-router/schema.ts
```diff
@@ -83,3 +83,7 @@ export const ZDeleteDraftDocumentMutationSchema = z.object({
 });
 
 export type TDeleteDraftDocumentMutationSchema = z.infer<typeof ZDeleteDraftDocumentMutationSchema>;
+
+export const ZSearchDocumentsMutationSchema = z.object({
+  query: z.string().min(1),
```
Mythie, line 88: Which will meant his also needs to change as well

packages/lib/server-only/document/search-documents-with-keyword.ts
ksushant6566: done now this takes a `limit` variable with a default value of 5.

ksushant6566:
Hey @Mythie , just pushed requested changes. if (anything needs to be changed) { please let me know :) }

apps/web/src/components/(dashboard)/common/command-menu.tsx
```diff
@@ -29,13 +31,20 @@ const DOCUMENTS_PAGES = [
...
   { label: 'Pending documents', path: '/documents?status=PENDING' },
   { label: 'Inbox documents', path: '/documents?status=INBOX' },
 ];
 
 const SETTINGS_PAGES = [
-  { label: 'Settings', path: '/settings', shortcut: SETTINGS_PAGE_SHORTCUT.replace('+', '') },
+  {
+    label: 'Settings',
+    path: '/settings',
+    shortcut: SETTINGS_PAGE_SHORTCUT.replace('+', ''),
```
Mythie, line 46: What's this about btw?

apps/web/src/components/(dashboard)/common/command-menu.tsx
```diff
@@ -53,6 +62,32 @@ export function CommandMenu({ open, onOpenChange }: CommandMenuProps) {
...
+  // const [searchResults, setSearchResults] = useState<
+  //   { label: string; path: string; value: string }[]
+  // >([]);
+
+  const { data: searchDocuments, isLoading: isSearchingDocuments } =
+    trpcReact.document.searchDocuments.useQuery(
+      {
+        query: search,
+      },
+      // {
```
Mythie, line 74: We should add `keepPreviousData` so we can hold onto the old results while we're fetching 😄

packages/lib/server-only/document/search-documents-with-keyword.ts
```diff
@@ -0,0 +1,69 @@
...
+export type SearchDocumentsWithKeywordOptions = {
+  query: string;
+  userId: number;
+  limit?: number;
+};
+
+export const searchDocumentsWithKeyword = async ({
+  query,
+  userId,
+  limit = 5,
```
Mythie, line 12: Good choice

Mythie:
Noting that the searchDocuments endpoint is angry:

image

ksushant6566:
> Noting that the searchDocuments endpoint is angry:
> 
> image

Oh shit! lemme check

Mythie:
It's because of migrations aha! This PR has been built on top of the other PR for document deletion so it won't work on a fresh checkout like the one I did on Gitpod

ksushant6566:
> It's because of migrations aha! This PR has been built on top of the other PR for document deletion so it won't work on a fresh checkout like the one I did on Gitpod

Yes!

Mythie:
As a final request then I'll need a video recording showing the 3 flows:

- document I own appearing in search results
- document I received showing in search results
- search by recipient

Once this has been provided I'm happy to accept this submission 🙌

ksushant6566:
> As a final request then I'll need a video recording showing the 3 flows:
> 
> * document I own appearing in search results
> * document I received showing in search results
> * search by recipient
> 
> Once this has been provided I'm happy to accept this submission 🙌

cool give me 5 minutes

kritarth2121:
The cmdk library filters based on whether the item contains a string sequence of **search**  in order, but not necessarily right next to each other, 

but the filter you have implemented filters whether the items containing **search substring** right next to each other

This feature can be implemented with regex, if you want I can build it.
@Mythie

Mythie (dismissed):

ksushant6566:
Hey @Mythie , Since this is approved I'll start working on #703 , in this branch only, since we want the search results in the same command menu.

Mythie (approved):

PR #714, opened 2023-12-02
feat: add document search to the command menu

kritarth2121:


Mythie:

https://documen.so/cla

kritarth2121:
> 
> https://documen.so/cla

Accepted it now :)

apps/web/src/components/(dashboard)/common/command-menu.tsx
```diff
@@ -53,6 +55,33 @@ export function CommandMenu({ open, onOpenChange }: CommandMenuProps) {
...
+
+  const debouncedSearchString = useDebouncedValue(search, 1000);
+  const { isLoading: isSearchingDocuments, refetch: findDocuments } =
+    trpc.document.searchDocuments.useQuery(
+      {
+        query: search,
+      },
+      {
+        enabled: search !== '',
+        onSuccess: (data) => {
```
Mythie, line 70: These hooks will be deprecated in future versions of react-query and so we should avoid them 😄

Mythie:
I've noticed that this is mostly the same code as #713 almost to the letter, I'd suggest teaming up with @ksushant6566 if possible to make the claim.

If this does get across the line I'll likely still have to give the bounty to Sushant and award you a secondary smaller amount since it looks like this is mostly his code.

kritarth2121:
> I've noticed that this is mostly the same code as #713 almost to the letter, I'd suggest teaming up with @ksushant6566 if possible to make the claim.
> 
> If this does get across the line I'll likely still have to give the bounty to Sushant and award you a secondary smaller amount since it looks like this is mostly his code.

As you said, you will give a bounty to him I worked on it mostly, and I would like to add one point to it as well, which can help in the feature.

The cmdk command **filters** based on whether the label contains a string sequence of search terms in order, but not necessarily right next to each other, but the filter we have implemented takes into account that substring should be right next to each other, Else it will not filter it.

This feature can be implemented with regex, if you want I can build it in this.
@Mythie

PR #716, opened 2023-12-03
feat: add recipient roles

### LINKED ISSUE #705: [OSSHACK] Recipient Roles
## Improvement Description

Add the ability to specify a role for a recipient. This will include CC (Carbon Copy), Signer, Viewer, and Approver.

- Signers will be able to sign the document as per usual.
- Approvers may be able to sign the document but will also be able to approve the document without signing it.
- Viewers will be able to view the document but will not be able to sign or approve it.
- CCs will only receive a copy of the completed document.

## Rationale

Not all recipients of a given document need to sign it. Some recipients may only need to view the document, others may need to approve it, and others may need to sign it. Support for these roles will allow for more flexibility in the types of documents that can be created and sent.

## Proposed Solution

- Add a `role` enum to the `Recipient` model that contains the role types.
- Update the add signers step to allow for the selection of a role for each recipient defaulting to `Signer`.
- Update our signing flow to allow for the different roles.
- Update the recipient popover to display the role of each recipient in addition to their email.
- Update our complete document flow to allow for the different roles and email sending requirements.

## Alternatives

N/A

## Additional Context
- The design for this improvement can be found on our Figma: https://www.figma.com/file/mKYFxgmekwxXsoXCP9dXXx/Documenso-Design-System?type=design&node-id=4%3A336&mode=design&t=RCNR7fdt1cxJEReM-1

thehanimo:
Fixes #705


Mythie:

https://documen.so/cla

Mythie:
I'll have to review this one when I'm back at a computer since it's a bit hard to follow. I understand there's a bit of nuance to this issue so I'll try and get back to you asap

apps/web/src/app/(dashboard)/documents/data-table-action-button.tsx
```diff
@@ -72,6 +73,8 @@ export const DataTableActionButton = ({ row }: DataTableActionButtonProps) => {
     window.URL.revokeObjectURL(link.href);
   };
 
+  if (recipient?.role === RecipientRole.CC && isComplete === false) return <div></div>;
```
Mythie, line 76: This one sticks out to me as a question but until I run it locally I'm unsure what the question is. Just seems odd to return nothing here.

apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
```diff
@@ -108,12 +110,32 @@ export const DataTableActionDropdown = ({ row }: DataTableActionDropdownProps) =
       <DropdownMenuContent className="w-52" align="start" forceMount>
         <DropdownMenuLabel>Action</DropdownMenuLabel>
 
-        <DropdownMenuItem disabled={!recipient || isComplete} asChild>
-          <Link href={`/sign/${recipient?.token}`}>
-            <Pencil className="mr-2 h-4 w-4" />
-            Sign
-          </Link>
-        </DropdownMenuItem>
+        {recipient?.role !== RecipientRole.CC && (
```
Mythie, line 113: Perhaps a CC'er should be able to view since it seems to complicate the code

packages/lib/server-only/document/complete-document-with-token.ts
```diff
@@ -88,7 +88,20 @@ export const completeDocumentWithToken = async ({
       id: document.id,
       Recipient: {
         every: {
-          signingStatus: SigningStatus.SIGNED,
+          OR: [
```
Mythie, line 91: Is there a way we can simplify this query? Perhaps there isn't but it'd be interesting to find out 🙌🏻

apps/web/src/app/(dashboard)/documents/data-table-action-button.tsx
thehanimo: This is just assuming that the CC'er shouldn't be able to view the document until it has been completed.

Mythie: You are absolutely right, I think the thing tripping me up here is that it appears as though a CC'er can see the document in their account. 

We may want this but only once the document is completed, otherwise perhaps they only ever receive the email.

This may balloon the query needed for `findDocuments` which is a concern.

packages/lib/server-only/document/complete-document-with-token.ts
Mythie: I've thought about it, for those that don't need to do anything with a document (CC'er) we can mark them as signed from the beginning.

For viewers I'm unsure, we may want them to confirm viewing rather than having simply opened the document much like an approver needs to confirm approving.

packages/lib/server-only/document/seal-document.ts
```diff
@@ -45,7 +45,13 @@ export const sealDocument = async ({ documentId, sendEmail = true }: SealDocumen
...
   });
 
-  if (recipients.some((recipient) => recipient.signingStatus !== SigningStatus.SIGNED)) {
+  if (
+    recipients.some(
+      (recipient) =>
+        recipient.signingStatus !== SigningStatus.SIGNED &&
+        (recipient.role === RecipientRole.SIGNER || recipient.role === RecipientRole.APPROVER),
+    )
+  ) {
```
Mythie, line 54: The alternative is we make something like this an extracted method and use it in lieu of the the big query above and then reuse it in places like here.

packages/ui/primitives/document-flow/add-signers.types.ts
```diff
@@ -8,6 +8,7 @@ export const ZAddSignersFormSchema = z
         nativeId: z.number().optional(),
         email: z.string().email().min(1),
         name: z.string(),
+        role: z.string(),
```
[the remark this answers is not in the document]
Mythie: Yep we should use the `nativeEnum` method with the `SignerRole` enum from prisma or whatever we called it 😄

Mythie (commented):
Noting that we shouldn't be able to add fields for Viewers and Copy'ers

Mythie (commented):
Noting that the CC'er received an invitation to sign the document but they should only receive an email once the document has been completed.

packages/ui/primitives/document-flow/add-signers.types.ts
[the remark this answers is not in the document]
thehanimo: Totally forgot zod supported enums :)

apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx
```diff
@@ -230,7 +230,9 @@ export const EditDocumentForm = ({
             <AddFieldsFormPartial
               key={fields.length}
               documentFlow={documentFlow.fields}
-              recipients={recipients}
+              recipients={recipients.filter(
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx:233]: Nice

packages/lib/server-only/document/complete-document-with-token.ts
```diff
@@ -88,20 +88,7 @@ export const completeDocumentWithToken = async ({
...
-              ],
-            },
-            {
-              AND: [
-                { signingStatus: SigningStatus.NOT_SIGNED },
-                { OR: [{ role: RecipientRole.CC }, { role: RecipientRole.VIEWER }] },
-              ],
-            },
-          ],
+          signingStatus: SigningStatus.SIGNED,
```
Mythie, line 91: This feels better right?

packages/lib/server-only/document/resend-document.tsx
```diff
@@ -59,6 +59,8 @@ export const resendDocument = async ({ documentId, userId, recipients }: ResendD
 
   await Promise.all(
     document.Recipient.map(async (recipient) => {
+      if (recipient.role === RecipientRole.CC) return;
```
Mythie, line 62: No single line if statements please, the early return is good though

Mythie (commented):
Change since last review, now doing the whole PR once more incase I've missed anything before running the UI

packages/lib/server-only/document/send-document.tsx
```diff
@@ -47,6 +47,9 @@ export const sendDocument = async ({ documentId, userId }: SendDocumentOptions)
 
   await Promise.all(
     document.Recipient.map(async (recipient) => {
+      if (recipient.sendStatus === SendStatus.SENT || recipient.role === RecipientRole.CC) {
+        return;
+      }
       const { email, name } = recipient;
```
Mythie, line 53: newline here please, let it breathe

thehanimo:
>  While it's nice to see the roles it feels little cramped right now, the font weight also steals attention from the recipient email which likely isn't a good thing. Consider muting the role label and making it more subtle.

Do you think a normal font weight would do? What if we went smaller than `text-xs`?  Or maybe make it a pill/label?

packages/lib/server-only/document/seal-document.ts
thehanimo: The approach of CC having SIGNED by default was nice!

Mythie:
Noting that Eric dropped out but had a really cool thing for recipient selection in #717. He was kind enough to grant permission to use any code you deem useful there as well 😄

Mythie:
> > While it's nice to see the roles it feels little cramped right now, the font weight also steals attention from the recipient email which likely isn't a good thing. Consider muting the role label and making it more subtle.
> 
> Do you think a normal font weight would do? What if we went smaller than `text-xs`? Or maybe make it a pill/label?

I'd consider text-xs with a normal font weight. I'd probably also use our `text-muted-forground` class

Mythie:
Noting that this will be the winner, just working through minor items now.

thehanimo:
> Noting that Eric dropped out but had a really cool thing for recipient selection in #717. He was kind enough to grant permission to use any code you deem useful there as well 😄

Tha is unfortunate :/
What are you referring to when you say recipient selection?


> I'd consider text-xs with a normal font weight. I'd probably also use our `text-muted-forground` class

It was `text-muted-foreground` all along. I changed it to `text-muted-forground/50`. Looks better to me!

packages/lib/server-only/document/find-documents.ts
```diff
@@ -62,6 +62,18 @@ export const findDocuments = async ({
...
+              },
+            },
+          },
+        },
+        {
+          status: ExtendedDocumentStatus.COMPLETED,
+          Recipient: {
+            some: {
+              email: user.email,
+              role: RecipientRole.CC,
```
dguyen, line 76: These two filters seem weird, unless I'm tripping

Both:
- Check for a completed document
- Check that one of the recipients is the current user

And then they check for where the current user is:
- Non CC
- CC

Which would mean that it could be any role?

dguyen (dismissed):

Mythie (approved):

PR #718, opened 2023-12-03
feat: add stepper component

### LINKED ISSUE #708: [OSSHACK] Create a Stepper component
## Improvement Description

- Create a stepper component that will enhance the user experience and reduce the complexity of current multistep forms.
- The component should use Radix UI and/or Shadcn to the best of its ability.

## Rationale

- Current multistep forms are complex and potentially inefficient.
- A stepper component will standardize the process, leading to easier maintenance and better user experience.
- Utilizing Radix UI and Shadcn ensures consistency with existing design and functionality paradigms.

## Proposed Solution

- Develop a `Stepper` component with the following structure:

```html
<Stepper onComplete?={callback}>
  <Step>
    <Component />
    <!-- 
      Where in the component I could use `useStep` hook to get the current step and the `next` function to go to the next step.
      Additionally I would need a previous method and an `isFirst` and `isLast` boolean.
    -->
  </Step>
  ...
</Stepper>
```

- The `useStep` hook within each `Step` component should provide:
  - The current step information.
  - `next` and `previous` methods to navigate between steps.
  - `isFirst` and `isLast` booleans to identify the beginning and end of the steps.

## Alternatives

- Explore third-party stepper components that could be adapted to our needs. This hasn't been promising so far.
- Refactor existing multistep forms without a stepper component, which could be more time-consuming and less efficient.

## Additional Context

- The new stepper component must be fully compatible with our current stack.
- Documentation and examples should be provided for easy integration into existing and future forms.

## Things to consider building on top of
- https://headless-stepper.netlify.app/
- https://github.com/devrnt/react-use-wizard

mikezzb:
Fixes #708

Created a stepper component that will enhance the user experience and reduce the complexity of current multistep forms.

### Features

- Support controlled (parent component controls the steps) & uncontrolled mode (children controls the step)
- Support parent monitoring for uncontrolled mode using `onComplete` or `onStepChanged`
- Added generic type `WithStep<T>` as an interface for stepper children

### Example:

- Refactored `EditDocumentForm` using the new stepper


apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx
```diff
@@ -35,7 +35,12 @@ export type EditDocumentFormProps = {
   documentData: DocumentData;
 };
 
-type EditDocumentStep = 'title' | 'signers' | 'fields' | 'subject';
+enum EditDocumentStepEnum {
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx:38]: Typescript enums are _cursed_ so we avoid them, a const object is a nice middleground that we use elsewhere in the codebase 😄

packages/ui/primitives/stepper.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useEffect, useState } from 'react';
```
Mythie, line 1: This is a pretty cool implementation, would be nice to potentially use context here rather than cloneElement so we don't have to concern ourselves with injected props which can be hard to reason about.

Mythie:
> Should we integrate the next / prev button and the step progress bar inside Stepper or let the Stepper children handle it? I can modify it based on the updated design and will add more documentation and example usage after the design is confirmed.

Ideally stepper children would handle it which would be another argument for having it via context, overall this is the start of a nice solution but needs a bit of refinement before it can be accepted.

My ideal solution provides the kind of API described in the issue where my children can deal with most of their logic removing the need for the current "document flow" that is mostly a bandaid around not having this component/api.

Mythie:
As mentioned on the other PR, yourself and #720 are both bringing some fantastic ideas on how to solve this. I reckon working together would net an amazing result, hope ya'll consider it! 🙌

Mythie (approved):

PR #720, opened 2023-12-03
feat: add stepper component

### LINKED ISSUE #708: [OSSHACK] Create a Stepper component

G3root:
fixes #708

### DIFF
--- a/packages/ui/primitives/document-flow/document-flow-root.tsx
+++ b/packages/ui/primitives/document-flow/document-flow-root.tsx
@@ -1,11 +1,9 @@
 'use client';
 
-import React, { HTMLAttributes } from 'react';
-
-import { motion } from 'framer-motion';
+import type { HTMLAttributes } from 'react';
+import React from 'react';
 
 import { cn } from '@documenso/ui/lib/utils';
-import { Button } from '@documenso/ui/primitives/button';
 
 export type DocumentFlowFormContainerProps = HTMLAttributes<HTMLFormElement> & {
   children?: React.ReactNode;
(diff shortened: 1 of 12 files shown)


packages/ui/primitives/document-flow/document-flow-root.tsx
```diff
@@ -72,97 +70,3 @@ export const DocumentFlowFormContainerContent = ({
     </div>
   );
 };
-
```
G3root, line 75: unused code

Mythie: looks like it

packages/ui/primitives/stepper/stepper-context.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const useStepper = () => {
+  const context = useContext(StepperProvider);
+
+  if (!context) {
+    throw Error('useContext hook should be called inside the Stepper Provider');
+  }
+
+  return context;
+};
```
Mythie, line 29: If you're using this pattern it's best to make the default value null otherwise the `!context` won't get triggered since it'll use the value you provided as the `createContext` arg

packages/ui/primitives/stepper/step.tsx
```diff
@@ -0,0 +1,133 @@
+import type { ReactNode } from 'react';
```
Mythie, line 1: This might have specialised a little too much into our specific UI, let's try and keep it a little more headless so we could use it for other kinds of forms such as team creation 😄

Mythie:
Fwiw you and #718 are both bringing the heat, might be worth trying to pair up and make something awesome together! 🔥

packages/ui/primitives/stepper/step.tsx
```diff
@@ -0,0 +1,133 @@
...
+  };
+
+  const isPreviousEnabled =
+    canGoPrevious === true
+      ? false
+      : isFirstStep === true
+      ? true
+      : disablePrevious === true
+      ? true
+      : false;
```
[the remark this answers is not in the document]
G3root: 😂 Spaghetti Code Alert !!!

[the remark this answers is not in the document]
adithyaakrishna: I love Spaghetti code xD

Mythie:
Closing as we've gone with the other submission, still love your work though 💚

PR #722, opened 2023-12-03
feat: document thumbnails

### LINKED ISSUE #704: [OSSHACK] Document Thumbnails
## Improvement Description

Add a thumbnail generation feature for documents uploaded, thumbnails will be created using the first page of the uploaded document.

A high quality and low quality thumbnail should be generated for each document uploaded. The low quality thumbnail will be used in the documents table for quick viewing, while the high quality thumbnail will be used for a more detailed preview upon clicking the thumbnail.

## Rationale

Right now, the documents table is very plain and does not provide any visual information about the document. Thumbnails will allow users to quickly identify the document they are looking for.

This is extremely useful for users with many documents that have similar names since it provides a visual cue to help them find the document they are looking for.

## Proposed Solution

- Add a thumbnail generation feature for documents uploaded, thumbnails will be created using the first page of the uploaded document.
- Use either a background job or an endpoint to perform an upsert-like operation for retrieving the thumbnail.
- Store the thumbnails somewhere? Currently we have the univeral upload which is catered towards storing documents but could be used for storing thumbnails as well. Alternatively we also store data uris in the database which I'm less inclined to do but is an option.

## Alternatives

- Use a third party service to generate thumbnails. This would be the easiest solution but would require a third party service to be used which we would like to avoid.

## Additional Context

- While this might be a good candidate for background jobs we should consider how we can make this work without background jobs since we don't currently have a background job system in place.
- We would like to avoid third-party services since we would like to keep the project as self-contained as possible.

mikezzb:
Closes #704

### Implementation:

- Generate a thumbnail on the client side using the rendering of an invisible LazyPDFViewer
- Store the base 64 string of the thumbnail in the database
- Error handling: reject the thumbnail generation if timeout (500ms)
- Keep the aspect ratio of the thumbnail but enforce the max dimension
- Implemented the primitive UI to display thumbnails in the table

### Preview:

Screenshot 2023-12-03 at 05 07 48
(For compatibility, previous document without thumbnail will show default icon as last one)

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/upload-document.tsx
+++ b/apps/web/src/app/(dashboard)/documents/upload-document.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 
 import Link from 'next/link';
 import { useRouter } from 'next/navigation';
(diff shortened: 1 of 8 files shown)


Mythie:
Now this is cool and just the kind of hackathon implementation I was talking about!

Two main points before I do the individual code spam:

- Since you're resizing it's trivial to store the HQ version as well, consider how we might do that without also shipping too much at a limited Vercel lamba function. I added some detail of this on the issue.
- If the process times out how do we ever recover? Do we? Be cool to see some self-healing for documents lacking thumbnails as well but isn't super critical since there is a fallback.

apps/web/src/app/(dashboard)/documents/upload-document.tsx
```diff
@@ -77,6 +100,40 @@ export const UploadDocument = ({ className }: UploadDocumentProps) => {
     }
   };
 
+  const generateThumbnail = (page: number, canvas: HTMLCanvasElement | null) => {
```
Mythie, line 103: This method is gnarly but I understand why, happy with it the way it is!

apps/web/src/app/(dashboard)/documents/upload-document.tsx
```diff
@@ -119,6 +176,20 @@ export const UploadDocument = ({ className }: UploadDocumentProps) => {
           </div>
         </div>
       )}
+
+      {Boolean(docData) && (
```
Mythie, line 180: Typically prefer `!!` over `Boolean(x)` can't recall why

PR #740, opened 2023-12-07
refactor: download function to be reusable

dephraiim:


packages/ui/primitives/use-toast.ts
```diff
@@ -1,7 +1,8 @@
 // Inspired by react-hot-toast library
 import * as React from 'react';
 
-import { ToastActionElement, type ToastProps } from './toast';
+import type { ToastActionElement } from './toast';
+import { type ToastProps } from './toast';
```
dguyen, line 5: The auto lint seemed to mess this up, can you clean this by doing this?

```
import type { ToastActionElement, ToastProps } from './toast';
```

packages/ui/components/document/document-download-button.tsx
```diff
@@ -24,44 +23,18 @@ export const DocumentDownloadButton = ({
...
-
-      const bytes = await getFile(documentData);
-
-      const blob = new Blob([bytes], {
-        type: 'application/pdf',
-      });
-
-      const link = window.document.createElement('a');
-      const baseTitle = fileName?.includes('.pdf') ? fileName.split('.pdf')[0] : fileName;
+    setIsLoading(true);
```
dguyen, line 29: It looks like this is never set to false in the scenario that `documentData` is not available since `finally` no longer exists

You should be able to just shift this below the if statement.

packages/lib/client-only/download-pdf.ts
```diff
@@ -0,0 +1,37 @@
...
+import { toast } from '@documenso/ui/primitives/use-toast';
+
+import { getFile } from '../universal/upload/get-file';
+
+type DownloadPDFProps = {
+  documentData: DocumentData;
+  fileName?: string;
+};
+
+export const downloadFile = async ({ documentData, fileName }: DownloadPDFProps) => {
```
dguyen, line 11: It feels strange that if this errors out the parent that calls this isn't aware

It might be more practical to just throw the error up and have the parent handle toasting and catching errors

apps/web/src/app/(dashboard)/documents/data-table-action-button.tsx
```diff
@@ -39,47 +37,25 @@ export const DataTableActionButton = ({ row }: DataTableActionButtonProps) => {
...
-        type: 'application/pdf',
+    if (!recipient) {
+      document = await trpcClient.document.getDocumentById.query({
+        id: row.id,
       });
+    } else {
+      document = await trpcClient.document.getDocumentByToken.query({
+        token: recipient.token,
+      });
+    }
```
dguyen, line 50: There doesn't appear to be any error handling for this anymore

packages/lib/client-only/download-pdf.ts
Mythie: I'd also name this the same as the file, so `downloadFile` => `downloadPDF`

packages/ui/primitives/use-toast.ts
dephraiim: sure, I'll fix it.

packages/ui/components/document/document-download-button.tsx
dephraiim: 👍

dguyen (approved):
Looks good to me :)

packages/lib/client-only/download-pdf.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const downloadPDF = async ({ documentData, fileName }: DownloadPDFProps) => {
+  const bytes = await getFile(documentData);
+
+  const blob = new Blob([bytes], {
+    type: 'application/pdf',
+  });
+
+  const link = window.document.createElement('a');
+  const baseTitle = fileName?.includes('.pdf') ? fileName.split('.pdf')[0] : fileName;
```
Mythie, line 18: `const [baseTitle] = fileName?.includes('.pdf') : fileName.split('.pdf') : [fileName ?? 'document'];`

This way you can do `${baseTitle}_signed.pdf` without needing the ternary below 😄

Mythie (approved):

PR #745, opened 2023-12-09
refactor: end-to-end tests

G3root:
what this PR do ?

- Remove fixed seeded records for testing.
- Generate fresh fixture records prior to each test and perform cleanups upon test completion.


apps/web/src/app/(authenticated)/logout/page.tsx
```diff
@@ -0,0 +1,13 @@
+'use client';
+
+import { useEffect } from 'react';
+
+import { signOut } from 'next-auth/react';
+
+export default function Logout() {
```
G3root, line 7: a route to logout the current user used in e2e test

packages/lib/server-only/user/create-user.ts
```diff
@@ -32,6 +39,7 @@ export const createUser = async ({ name, email, password, signature }: CreateUse
       password: hashedPassword,
       signature,
       identityProvider: IdentityProvider.DOCUMENSO,
+      emailVerified,
```
G3root, line 42: added emailVerified arg because this function can be reuseable in user test fixture to create a verified email user

packages/lib/server-only/document/create-document.ts
```diff
@@ -1,19 +1,27 @@
...
 
 export type CreateDocumentOptions = {
   title: string;
   userId: number;
   documentDataId: string;
+  status?: DocumentStatus;
 };
 
-export const createDocument = async ({ userId, title, documentDataId }: CreateDocumentOptions) => {
+export const createDocument = async ({
```
G3root, line 13: added `status` arg because this function can be reuseable in doxument test fixture to create documents based on status

packages/app-tests/e2e/test-auth-flow.spec.ts
```diff
@@ -14,7 +16,7 @@ const username = 'Test User';
 const email = 'test-user@auth-flow.documenso.com';
 const password = 'Password123';
 
-test('user can sign up with email and password', async ({ page }: { page: Page }) => {
+test('user can sign up with email and password', async ({ page }) => {
```
G3root, line 19: no need to type as we are using the test function from the fixture

packages/app-tests/e2e/pr-718-add-stepper-component.spec.ts
```diff
@@ -71,5 +64,5 @@ test(`[PR-718]: should be able to create a document`, async ({ page }) => {
   await page.waitForURL('/documents');
 
   // Assert document was created
-  await expect(page.getByRole('link', { name: documentTitle })).toBeVisible();
+  await page.getByRole('link', { name: documentTitle }).click();
```
G3root, line 67: Successfully generate a document so it can be removed when the user is deleted after the test finishes

packages/app-tests/playwright.config.ts
```diff
@@ -18,7 +18,7 @@ export default defineConfig({
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 2 : undefined,
```
G3root, line 21: run tests parallely

PR #749, opened 2023-12-09
feat: disable upload document animation for unverified users

dephraiim:
CleanShot 2023-12-09 at 22 01 48@2x


.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 {
   "typescript.tsdk": "node_modules/typescript/lib",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
Mythie, line 4: 👀

packages/ui/primitives/document-dropzone.tsx
```diff
@@ -104,11 +104,12 @@ export const DocumentDropzone = ({
 
   return (
     <motion.div
-      className={cn('flex', className)}
+      className={cn('flex aria-disabled:cursor-none', className)}
```
Mythie, line 107: `cursor-not-allowed` would be better I think, good use of aria-disabled though I was going to ask for it but forgot to

packages/ui/primitives/toast.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from 'react';
```
Mythie, line 1: Old code 😇

dguyen:
Can we add an actual message in the dropdown indicating why it's disabled?

.vscode/settings.json
dephraiim: fixed

dguyen (approved):

PR #758, opened 2023-12-11
chore: fix workflows and update package.json file

adithyaakrishna:


package.json
```diff
@@ -43,7 +43,6 @@
     "turbo": "^1.9.3"
   },
   "name": "@documenso/root",
-  "packageManager": "npm@8.19.2",
```
documenso-inc, line 46: I feel like we had this for a reason but can't recall

Mythie (approved):

Issue #766, opened 2023-12-16
Improve rendering logic for buttons in NotFound Page


plxity:
### Improvement Description

Currently, we are checking `sessions` multiple times to see which component to render. We can leverage the ternary operator to do that. It would make the code cleaner

Current Implementation
```js
      {session && (
        <Button className="w-32" asChild>
          <Link href="/documents">Documents</Link>
        </Button>
      )}

      {!session && (
        <Button className="w-32" asChild>
          <Link href="/signin">Sign In</Link>
        </Button>
      )}
```

### Rationale

This improvement will make the code cleaner

### Proposed Solution

Using a ternary operator to render the button based on the conditions. 

Proposed Solution
```js
      {session ? (
        <Button className="w-32" asChild>
          <Link href="/documents">Documents</Link>
        </Button>
      ) : (
        <Button className="w-32" asChild>
          <Link href="/signin">Sign In</Link>
        </Button>
      )}
```

### Please check the boxes that apply to this improvement suggestion.

- [X] I have searched the existing issues and improvement suggestions to avoid duplication.
- [X] I have provided a clear description of the improvement being suggested.
- [X] I have explained the rationale behind this improvement.
- [X] I have included any relevant technical details or design suggestions.
- [X] I understand that this is a suggestion and that there is no guarantee of implementation.

catalinpit:
Hey @plxity, thanks for opening the issue and for the PR! Mostly, we try to stay away from the conditional ternary operator, since it can make the code harder to read & understand. As a result, we won't move forward with this PR. Thanks for understanding! 🙏

plxity:
@catalinpit :  Okay sure, but I believe ternary operators are way more readable than the approach already used. It becomes complex if it's nested, otherwise it's actually a clearer way to implement conditional rendering.

PR #767, opened 2023-12-16
fix: improved the rendering logic in NotFound component

plxity:
Fixes: #766


catalinpit:
Hey @plxity, thanks for the PR! Mostly, we try to stay away from the conditional ternary operator, since it can make the code harder to read & understand. As a result, we won't move forward with this PR. Thanks for understanding! 🙏

PR #772, opened 2023-12-19
feat: show document title for delete dialog - [DOC-387]

adithyaakrishna:
**Description:**

- Fixes https://linear.app/documenso/issue/DOC-387/delete-document-dialog-should-mention-the-filename

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
+++ b/apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
@@ -21,6 +21,7 @@ type DeleteDraftDocumentDialogProps = {
   open: boolean;
   onOpenChange: (_open: boolean) => void;
   status: DocumentStatus;
+  documentTitle: string;
 };
 
 export const DeleteDocumentDialog = ({
@@ -28,6 +29,7 @@ export const DeleteDocumentDialog = ({
   open,
   onOpenChange,
   status,
+  documentTitle,
 }: DeleteDraftDocumentDialogProps) => {
   const router = useRouter();
 
@@ -42,7 +44,7 @@ export const DeleteDocumentDialog = ({
 
       toast({
         title: 'Document deleted',
-        description: 'Your document has been successfully deleted.',
+        description: `Your document '${documentTitle} has been successfully deleted.`,
         duration: 5000,
       });
 
--- a/apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
+++ b/apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
@@ -41,10 +41,16 @@ export const DeleteDocumentDialog = ({
   const { mutateAsync: deleteDocument, isLoading } = trpcReact.document.deleteDocument.useMutation({
     onSuccess: () => {
       router.refresh();
+      const deletedFileToastDescription = (
+        <p>
+          Your document <span className="text-documenso">{documentTitle}</span> has been
+          successfully deleted.
+        </p>
+      );
 
       toast({
         title: 'Document deleted',
-        description: `Your document "${documentTitle}" has been successfully deleted.`,
+        description: deletedFileToastDescription,
         duration: 5000,
       });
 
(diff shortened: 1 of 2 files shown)


dguyen (approved):

Mythie:
We should also display the title within the dialog rather than just the resulting toast notification as well, otherwise you're going to learn that you deleted the wrong document after the fact.

adithyaakrishna:
@Mythie How abt if we put it in a code block for the file name? or should we put it in double quotes?

adithyaakrishna:
Updated the modal and the toast notification to show the documentTitle :)

**Confirmation Modal:**

Screenshot 2023-12-20 at 10 35 15

**Toast Notification:**

Screenshot 2023-12-20 at 10 35 22

Mythie:
I would use quotes and do without the extra colour since it makes it look like a link 😄

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
```diff
@@ -39,10 +41,16 @@ export const DeleteDocumentDialog = ({
   const { mutateAsync: deleteDocument, isLoading } = trpcReact.document.deleteDocument.useMutation({
     onSuccess: () => {
       router.refresh();
+      const deletedFileToastDescription = (
+        <p>
+          Your document <span className="text-documenso">{documentTitle}</span> has been
```
catalinpit, line 46: We should probably avoid using green text on a white background, according to the WCAG compliance results. What do you think?

CleanShot 2023-12-20 at 17 37 12

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
```diff
@@ -72,7 +80,10 @@ export const DeleteDocumentDialog = ({
     <Dialog open={open} onOpenChange={(value) => !isLoading && onOpenChange(value)}>
       <DialogContent>
         <DialogHeader>
-          <DialogTitle>Do you want to delete this document?</DialogTitle>
+          <DialogTitle>
+            Do you want to delete the <span className="text-documenso">{documentTitle}</span>{' '}
```
catalinpit, line 84: Same as above

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
```diff
@@ -81,7 +92,7 @@ export const DeleteDocumentDialog = ({
         </DialogHeader>
 
         {status !== DocumentStatus.DRAFT && (
-          <div className="mt-8">
+          <div className="mt-4">
```
catalinpit, line 95: After testing various top margins, it seems that this works well without any margin.

What do you think?

CleanShot 2023-12-20 at 17 41 37

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
adithyaakrishna: Yes Yes, I will update it to just be in quotes instead of the green text 👀

apps/web/src/app/(dashboard)/documents/delete-document-dialog.tsx
adithyaakrishna: I also felt the same, one of the previous PRs introduced the `mt-8` and I felt the margin was too huge as well

Mythie: That would be me, I like having the action buttons separated from any kind of input/form elements to assist with misclicking

catalinpit: In the example above, it doesn't seem like it would be easy to misclick. Or am I missing something? 👀

ElTimuro:
> Updated the modal and the toast notification to show the documentTitle :)
> 
> **Confirmation Modal:**
> 
> Screenshot 2023-12-20 at 10 35 15
> **Toast Notification:**
> 
> Screenshot 2023-12-20 at 10 35 22

@adithyaakrishna The phrasing is a bit bulky, I'd prefer sth. like "Do you want to delete XXXX" or "Do you want to delete the document XXXXX"

Mythie (approved):

PR #785, opened 2023-12-26
fix: fixed the title box overlapping issue

suryapts007:
The issue is fixed. Now the box is no more overlapping

Screenshot 2023-12-26 at 10 20 32 AM


catalinpit (approved):
thanks for the fix!

PR #787, opened 2023-12-27
feat: update disabled dropzone text

dguyen:
## Description

Update the dropzone so it will display the relevant disabled text based on the reason it is disabled.

image

## Checklist


catalinpit (approved):
Nice one!

PR #788, opened 2023-12-27
fix: resolve conflicting z-index values

dguyen:
## Description

Currently there are various z-index values that are causing:
- Toasts to be placed behind dialog blur background
- Menu being cropped off by header

## Changes Made

- Revert `z-[1000]` back to `z-50` for the header (not exactly sure why it was bumped)
- Refactor z-indexes over 9000 to start from 1000
- Ensure z-index of toast is higher than dialog

## Testing Performed

**Mobile toast is on top**

image

**Toast is in-front of blurred dialog background**
image

**Dropdown is no longer being cropped off**

image

## Checklist

### DIFF
--- a/apps/web/src/components/(dashboard)/layout/header.tsx
+++ b/apps/web/src/components/(dashboard)/layout/header.tsx
@@ -33,7 +33,7 @@ export const Header = ({ className, user, ...props }: HeaderProps) => {
   return (
     <header
       className={cn(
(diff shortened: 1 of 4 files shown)


dguyen:
Having the dialog at 1000 will probably cause issues with any sub elements that rely in z-index.

catalinpit (approved):

PR #789, opened 2023-12-27
chore: slugify team urls and add profanity filter

adithyaakrishna:


dguyen:
Was the slugify library not that good?

packages/trpc/package.json
```diff
@@ -17,10 +17,13 @@
     "@trpc/next": "^10.36.0",
     "@trpc/react-query": "^10.36.0",
     "@trpc/server": "^10.36.0",
+    "bad-words": "^3.0.4",
```
catalinpit, line 20: We had to do something similar at iCodeThis, but instead of URLs, we had to make sure the user comments didn't contain profanity.

We created an array with profane words, rather than installing a package. So, that made me think - do we need to install a library for this? Also, the words list seems really small, unless I'm missing something.

Just thinking out loud. 😃

dguyen: I just think it's weird to have profanity directly in our repository where it may show up in searches

catalinpit: Gotcha

adithyaakrishna:
@dguyen I found that `slugify` library would be a lot for out use case. The `generate-url-slug.ts` would be enough is what I felt. 

Also David, was this branch merged with `feat/teams`?

dguyen:
It wasn't merged, `feat/teams` was rebased on main and force pushed so it may have messed up some things

After discussing with Lucas, we decided to exclude profanity filtering for now in the first implementation and do the following to push the feature out faster:
- Only allow lower case, numbers and dashes
- Restrict names containing things like `documenso`, `support`, etc

I can do this since it's just some hardcode, thanks for looking into this issue though it'll be useful for the future 😄

Issue #793, opened 2023-12-28
Fix padding in footer


plxity:
### Issue Description

Currently, padding is not getting respected in the footer

### Steps to Reproduce

1. Visit the footer in the mobile view

### Expected Behavior

Check the screenshot below
<img width="390" alt="Screenshot 2023-12-28 at 11 11 04 PM" src="https://github.com/documenso/documenso/assets/28510494/4d1b544e-541a-4517-85a2-a3256ec89ca2">

### Current Behavior

Check the Screenshot below
<img width="401" alt="Screenshot 2023-12-28 at 11 06 41 PM" src="https://github.com/documenso/documenso/assets/28510494/4f432b8d-a9f7-46c5-aef3-cbd0ff91aa2d">
Shot below

### Please check the boxes that apply to this issue report.

- [X] I have searched the existing issues to make sure this is not a duplicate.
- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I have included any relevant screenshots.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [X] I want to work on creating a PR for this issue if approved

PR #794, opened 2023-12-28
fix: fixed padding in footer

plxity:
fixes: #793


Mythie:
Hey Apoorv, I commandeered your PR since the real issue is due to the footer not being responsive! You can hear more on the changes I made and such here: https://www.youtube.com/watch?v=LdGU14D6ROg

Unfortunately I opened the issue rather the PR in my Gitpod which meant I committed directly on `main` 🙈 

I'll still approve this to handle extra small devices as well 🎉

Mythie (approved):

PR #804, opened 2024-01-04
feat(ci): cache github workflow actions

G3root:


adithyaakrishna (dismissed):
@G3root This would be a good improvement to the CI, I was planning to cache the build artifact yesterday and I guess it's been implemented already

.github/workflows/clean-cache.yml
```diff
@@ -0,0 +1,29 @@
+name: cleanup caches by a branch
```
G3root, line 1: from https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries

.github/workflows/ci.yml
```diff
@@ -47,5 +37,31 @@ jobs:
         with:
           fetch-depth: 2
 
+      - name: Set up Docker Buildx
```
G3root, line 40: from https://docs.docker.com/build/ci/github-actions/cache/#local-cache

Mythie:
I thought that the npm cache was already satisfactory? Was I wrong in that assumption 🤔 

Work looks really good either way, although I wonder what the cost of this will look like 👀

adithyaakrishna:
If I am not wrong, I guess we would be using a built version of Documenso instead of rebuilding it again and again unless code stuff has changed. Along with this, I guess we would also be caching node_modules in a better way. Overall this may give us a slight advantage in terms of the CI timings

CC @G3root

G3root:
> If I am not wrong, I guess we would be using a built version of Documenso instead of rebuilding it again and again unless code stuff has changed. Along with this, I guess we would also be caching node_modules in a better way. Overall this may give us a slight advantage in terms of the CI timings
> 
> CC @G3root

yeah, we can go with that approach, but I believe it carries some risk because Changes in environment variables could affect the build output. Or, we can try experimenting with TurboRepo to manage the cache. https://turbo.build/repo/docs/ci/github-actions#caching-with-github-actionscache

Mythie (approved):

Mythie:
Sorry for the delay, approving but delaying merging for the moment since I'd like to limit some blast radius.

/remindme 1 week

Mythie:
/tip 150

The work is much appreciated by the way 😄

PR #739, opened 2023-12-07
feat: signing auto scroll with tips

Hiperultimate:
resolves: #710 

Auto-scroll to open field is now being triggered on pressing complete button.

Also fixes an issue where field tool tips were not being displayed 

https://github.com/documenso/documenso/assets/32923170/7d689207-ec96-4097-afdc-4f0eeb0b00a1

image

PR #730, opened 2023-12-05
feat: fix jump to sign field

ketanMehtaa:
resolves https://github.com/documenso/documenso/issues/710.

PR #729, opened 2023-12-05
feat: auto-scroll to next open-field while signing document

ksushant6566:
resolves #710.

The complete button now takes the user to the next open-field. When all fields are done, opens confirmation dialog as usual.
attaching a loom below.

Edit: added a conditional on complete button to say `Skip to next field` when there are open fields and `Complete` otherwise.

https://www.loom.com/share/bb554acaa2b947139cd125e482910dd7?sid=85e37cd8-7f9a-4ce1-8a67-d3a82f5ea14b

PR #805, opened 2024-01-05
feat: jump to next field

### LINKED ISSUE #710: [DOC-376] No jump to sign field
* When the user already has a signature in their account, the complete button is disabled, forcing them to scroll through the doc and click each field manually.
* When clicked on complete, the auto-scroll to open field should also be triggered.

<sub>From [SyncLinear.com](https://synclinear.com) | [DOC-376](https://linear.app/documenso/issue/DOC-376/no-jump-to-sign-field)</sub>

catalinpit:
Fixes #710

When the fields are not filled, the button will have the `outline` variant and it'll say "Fill fields". Clicking on the button takes you to the unfilled field.

CleanShot 2024-01-05 at 12 58 24

Once all fields are filled, the button will be green and it'll say "Complete".

CleanShot 2024-01-05 at 12 58 30

Created the PR to move faster and ship the changes asap.

PR based on these PRs:
- https://github.com/documenso/documenso/pull/739 by @Hiperultimate
- https://github.com/documenso/documenso/pull/729 by @ksushant6566
- https://github.com/documenso/documenso/pull/730 by @ketanMehtaa

Thank you for the work, folks!

### DIFF
--- a/apps/web/src/app/(signing)/sign/[token]/form.tsx
+++ b/apps/web/src/app/(signing)/sign/[token]/form.tsx
@@ -49,6 +49,11 @@ export const SigningForm = ({ document, recipient, fields }: SigningFormProps) =
     return sortFieldsByPosition(fields.filter((field) => !field.inserted));
   }, [fields]);
 
+  const fieldsValidated = () => {
+    setValidateUninsertedFields(true);
+    validateFieldsInserted(fields);
+  };
+
   const onFormSubmit = async () => {
     setValidateUninsertedFields(true);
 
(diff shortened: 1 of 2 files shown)


dguyen:
I think the button shouldn't be called `Fill fields` since it's not 'filling' anything. `Next Field` or something else would be more applicable here

Also the button should probably stay as the default variant in my opinion

Mythie:
+1 for both pieces of feedback 😄

catalinpit:
Thanks for the feedback! 

I implemented both things, although, I must say it feels a bit weird (personally) to have the button green before you can complete the signing. But if this is the consensus, I'll abide it. 😆

dguyen (approved):

PR #828, opened 2024-01-11
feat: sign document with text signature

ephraimduncan:


apps/web/src/app/(signing)/sign/[token]/form.tsx
```diff
@@ -128,15 +156,71 @@ export const SigningForm = ({ document, recipient, fields }: SigningFormProps) =
...
+
+                    <div
+                      className="absolute inset-x-0 bottom-0 flex cursor-auto items-end justify-between px-4 pb-1 pt-2"
+                      onClick={(e) => e.stopPropagation()}
+                      onKeyDown={(e) => e.stopPropagation()}
+                    >
+                      <Input
+                        id="signatureText"
+                        className="text-foreground placeholder:text-muted-foreground border-0 border-none bg-transparent p-0 text-sm focus-visible:ring-transparent"
+                        placeholder="Draw or type name here"
```
Comment (adithyaakrishna) [inline apps/web/src/app/(signing)/sign/[token]/form.tsx:189]:
```suggestion
                        placeholder="Draw or type your name here"
```

packages/lib/server-only/document/seal-document.ts
```diff
@@ -69,6 +69,8 @@ export const sealDocument = async ({ documentId, sendEmail = true }: SealDocumen
 
   const doc = await PDFDocument.load(pdfData);
 
+  console.log('Fields to insert into PDF: ', fields);
```
adithyaakrishna, line 72: @dephraiim Could you please remove this line?

apps/marketing/src/components/(marketing)/widget.tsx
```diff
@@ -392,10 +393,11 @@ export const Widget = ({ className, children, ...props }: WidgetProps) => {
...
           <DialogDescription>
-            By signing you signal your support of Documenso's mission in a <br></br>
-            <strong>non-legally binding, but heartfelt way</strong>. <br></br>
-            <br></br>You also unlock the option to purchase the early supporter plan including
-            everything we build this year for fixed price.
+            By signing you signal your support of Documenso's mission in a <br />
+            <strong>non-legally binding, but heartfelt way</strong>. <br />
+            <br />
+            You also unlock the option to purchase the early supporter plan including everything we
+            build this year for fixed price.
```
adithyaakrishna, line 400: Can we rephrase this line into something that is much friendlier? 😶  

Eg: 
`Join the exclusive club! Choose the early supporter plan to snag all our creations this year at a fixed, pocket-friendly rate.` 
or
`Your key to a year of excitement: the early supporter plan! Grab all our new releases at a fantastic, fixed price`

adithyaakrishna: @catalinpit @dguyen Would love to know your opinion on this

dguyen: I think the current is fine with maybe a few adjustments if we really want

packages/lib/server-only/pdf/insert-field-in-pdf.ts
```diff
@@ -12,7 +12,10 @@ import { isSignatureFieldType } from '@documenso/prisma/guards/is-signature-fiel
 import type { FieldWithSignature } from '@documenso/prisma/types/field-with-signature';
 
 export const insertFieldInPDF = async (pdf: PDFDocument, field: FieldWithSignature) => {
-  const fontCaveat = await fetch(process.env.FONT_CAVEAT_URI).then(async (res) =>
+  const fontDancingScriptURL =
```
Mythie, line 15: Ideally we do the same trick we did for Caveat since going to an external source is costly

Mythie (approved):

Mythie (changes requested):
Misclick soz

packages/lib/server-only/pdf/insert-field-in-pdf.ts
dephraiim: okay sure, I'll pull dancing script locally and we can use it

apps/marketing/src/components/(marketing)/widget.tsx
dephraiim: For the time being, I think it would be best to maintain the current line. We can discuss the changes in a separate pull request.

dguyen:
@dephraiim I tested overflowing your signature changes (after looking at #877), and it seemed to break a bit

image

ElTimuro:
@dguyen @dguyen be advised that when we release it, it has to be together with an option to force the user to use handwritten as part of 21 CFR part 11

PR #835, opened 2024-01-16
feat: restrict app access for unverified users

catalinpit:


apps/web/src/components/forms/signup.tsx
```diff
@@ -61,10 +63,12 @@ export const SignUpForm = ({ className }: SignUpFormProps) => {
     try {
       await signup({ name, email, password, signature });
 
-      await signIn('credentials', {
-        email,
-        password,
-        callbackUrl: '/',
+      router.push('/signin');
```
catalinpit, line 66: Replaced the `signIn` function with a redirect to the `/signin` page, since the newly-registered users can't sign in until they confirm their email address.

apps/web/src/components/forms/signin.tsx
```diff
@@ -122,6 +130,15 @@ export const SignInForm = ({ className }: SignInFormProps) => {
 
         const errorMessage = ERROR_MESSAGES[result.error];
 
+        if (result.error === ErrorCode.UNVERIFIED_EMAIL) {
+          const user = await getUser({ email });
+          const token = user?.VerificationToken[user.VerificationToken.length - 1].token;
```
catalinpit, line 135: Doing this workaround to avoid exposing the email address in the URL. The assumption is that a newly-registered user will always have at least 1 verification token. We'll use the last verification token and identify the user on the `/unverified-account` page, whenever the user clicks the "resend confirmation email" button.

If you can think of a better solutions, it is more than welcome!

apps/web/src/app/(unauthenticated)/unverified-account/page.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const { toast } = useToast();
+
+  const token = searchParams?.get('t') ?? '';
+
+  const { data: { email } = {} } = trpc.profile.getUserFromVerificationToken.useQuery({ token });
+
+  const { mutateAsync: sendConfirmationEmail } = trpc.profile.sendConfirmationEmail.useMutation();
+
+  const onResendConfirmationEmail = async () => {
+    if (!email) {
```
catalinpit, line 28: Avoid pressing the button before the email is fetched.

apps/web/src/components/forms/signin.tsx
Gautam-Hegde, line 135: Do we need a `user.VerificationToken.length > 0` check or is it implicit?

catalinpit: The newly registered user will always have at least one token. Unless, it's manually deleted from the db.

packages/trpc/server/profile-router/router.ts
```diff
@@ -31,6 +35,36 @@ export const profileRouter = router({
...
+        const { email } = input;
+
+        return await getUserByEmail({ email });
+      } catch (err) {
+        throw new TRPCError({
+          code: 'BAD_REQUEST',
+          message: 'We were unable to retrieve the specified account. Please try again.',
+        });
+      }
+    }),
```
dguyen, line 51: Looks like anyone can hit this endpoint and retrieve the whole user object based on any email?

dguyen (changes requested):

dguyen:
After clicking sign up, the user should probably be sent to `/unverified-account`?

Since currently nothing happens when they click sign up.

packages/lib/next-auth/auth-options.ts
```diff
@@ -69,6 +70,17 @@ export const NEXT_AUTH_OPTIONS: AuthOptions = {
...
+        const createdWithinLast72Hours = userCreationDate > new Date(Date.now() - ONE_DAY * 3);
+
+        /*
+          avoid messing with the users who signed up before the email verification requirement
+          the error is thrown only if the user doesn't have a verified email and the account was created within the last 72 hours
+        */
+        if (!user.emailVerified && createdWithinLast72Hours) {
+          throw new Error(ErrorCode.UNVERIFIED_EMAIL);
+        }
+
```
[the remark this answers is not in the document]
Mythie: I will :)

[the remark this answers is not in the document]
dguyen: I think it's good to stop users from accessing an unverified account regardless if they previously registered.

If they haven't confirmed their email they wouldn't be able to do much on the application anyway.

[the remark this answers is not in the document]
catalinpit: Yeah, that was my thinking too initially. But after the discussions, I did it this way.

packages/trpc/server/profile-router/router.ts
catalinpit: 🤦‍♂️ thanks for the catch!

apps/web/src/components/forms/signin.tsx
dguyen: This is going to cause an issue because you're exposing the token to the user.

For example, the user signs in with an unconfirmed email and gets redirected to:

http://localhost:3000/unverified-account?t=72cc56a3c55dbb5a3c60e3d8203f6c20f345fb2b

They can take the token in the search param, and go directly to `/verify-email` and have the email be verified

http://localhost:3000/verify-email/72cc56a3c55dbb5a3c60e3d8203f6c20f345fb2b

A better solution would be to encrypt the email into the search param and decrypt it

There's a pending PR for this now (if you could please review it if you have time): 

https://github.com/documenso/documenso/pull/863

apps/web/src/app/(unauthenticated)/unverified-account/page.tsx
```diff
@@ -0,0 +1,68 @@
...
+      </div>
+      <div className="">
+        <h2 className="text-2xl font-bold md:text-4xl">Confirm email</h2>
+
+        <p className="text-muted-foreground mt-4">
+          To gain full access to your account and unlock all its features, please confirm your email
+          address by clicking on the link sent to your email address.
+        </p>
+
+        <Button className="mt-4" disabled={isButtonDisabled} onClick={onResendConfirmationEmail}>
```
dguyen, line 62: Can we use the loading prop here?

apps/web/src/app/(unauthenticated)/unverified-account/page.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+      await sendConfirmationEmail({ encryptedEmail });
+
+      toast({
+        title: 'Success',
+        description: 'Verification email sent successfully.',
+        duration: 5000,
+      });
+
+      setTimeout(() => setIsButtonDisabled(false), RESEND_CONFIRMATION_EMAIL_TIMEOUT);
```
dguyen, line 37: I think it would be more intuitive to the user if we do something like this instead

- Store a time of when the last email was sent
- Allow the button to always be clickable unless we are in the process of sending an email

Then at the start of this function, if the cooldown period is still active we can toast saying they need to wait X amount

-------------

Alternatively, we can display something in the UI saying please wait X amount of time

I'm unsure which is better, or if there are other solutions but imo there should be some sort of indicator to the user

apps/web/src/app/(unauthenticated)/unverified-account/page.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Mails } from 'lucide-react';
+
+import { ONE_SECOND } from '@documenso/lib/constants/time';
+import { trpc } from '@documenso/trpc/react';
+import { Button } from '@documenso/ui/primitives/button';
+import { useToast } from '@documenso/ui/primitives/use-toast';
+
+const RESEND_CONFIRMATION_EMAIL_TIMEOUT = 20 * ONE_SECOND;
+
+export default function UnverifiedAccount() {
```
dguyen, line 16: I think it may be more optimal to have this page not be a client page, so we can immediately decrypt the search param on the server side, then check if it's an email and pass it to a client component which handles the rest. 

It'll let us:
- Display the email (if required)
- Handle invalid emails directly (error state)
- Frees up the restriction we have on the sendConfirmation endpoint where the value has to be encrypted

Thoughts? Just a suggestion

catalinpit: Thanks for the review! Sounds good, but I have the following questions:
1 - when would we display the email on the `unverified page`?
2 - wouldn't invalid emails be handled in the registration process? The user is redirected directly to the `unverified page` after signing up/logging. So the user can't manually provide an invalid address. Or am I missing something?
3 - does that make any difference?

All in all, sounds good. Just the above questions for my understanding.

dguyen: #### 1 - when would we display the email on the unverified page?

It could potentially be used in the template, or in a toast when the email is resent

Example, the current template doesn't make too much sense since there's no partial access anymore

```html
<p className="text-muted-foreground mt-4">
  To gain full access to your account and unlock all its features, please confirm your email
  address by clicking on the link sent to your email address.
</p>
```

We could do something like this, which would be more informative

> An email has been sent to your email address something@something.com. Please confirm...




####  2 - wouldn't invalid emails be handled in the registration process? The user is redirected directly to the unverified page after signing up/logging. So the user can't manually provide an invalid address. Or am I missing something?

The search param is editable, so this could theoretically be changed to anything (highly unlikely, and no real reason to)

Having it done server side would allow us to check if there is even a pending user for this email, so we can display an appropriate message for that scenario as well if we wanted to

#### 3 - does that make any difference?

Nope, since I don't think this endpoint would be used elsewhere (?), it's just cleaner

catalinpit: 1 - hmmm, ok. I understand.

2 - it is editable indeed, but isn't it encrypted?

3 - for now, yes.

Ok then, let's make this a server component.

packages/lib/server-only/user/get-user-by-email.ts
```diff
@@ -9,5 +9,8 @@ export const getUserByEmail = async ({ email }: GetUserByEmailOptions) => {
     where: {
       email: email.toLowerCase(),
     },
+    include: {
+      VerificationToken: true,
```
Mythie, line 13: Should we do this or fetch it in a separate method for usage in auth-options?

e.g. `getRecentVerificationToken({ userId }) => Promise<VerificationToken | null>`

packages/lib/next-auth/auth-options.ts
```diff
@@ -90,6 +91,19 @@ export const NEXT_AUTH_OPTIONS: AuthOptions = {
           }
         }
 
+        if (!user.emailVerified) {
+          const totalUserVerificationTokens = user.VerificationToken.length;
+          const lastUserVerificationToken = user.VerificationToken[totalUserVerificationTokens - 1];
```
Mythie, line 96: As per my other comment it might be nice to get the token as a dedicated item, then we can check if it's expired or was sent > X minutes ago at which point we would resend 😄

apps/web/src/components/forms/send-confirmation-email.tsx
```diff
@@ -0,0 +1,93 @@
...
+                </FormItem>
+              )}
+            />
+          </fieldset>
+          <Button size="lg" type="submit">
+            Send confirmation email
+          </Button>
+        </form>
+      </Form>
+    </div>
```
dguyen, line 91: - I think outer div is redundant
- `fieldset` should wrap around the button
- You should add `<FormMessage />` to handle invalid emails
- Probably throw a `isSubmitting` to the button `loading` prop

dguyen (changes requested):

packages/lib/next-auth/auth-options.ts
```diff
@@ -90,6 +92,26 @@ export const NEXT_AUTH_OPTIONS: AuthOptions = {
...
         }
 
+        if (!user.emailVerified) {
+          const [lastUserVerificationToken] = await getLastVerificationToken({ userId: user.id });
+
+          if (!lastUserVerificationToken) {
+            await sendConfirmationToken({ email });
+            throw new Error(ErrorCode.UNVERIFIED_EMAIL);
+          }
+
```
catalinpit, line 102: should the `expiredToken`, `lastSentToken` and `sentWithinLastHour` be extracted into a separate function? or is it fine like this?

Mythie (approved):

Mythie:
lgtm 🔥

PR #839, opened 2024-01-17
fix: fix bug for completed document shortcut

catalinpit:
When you're in the `/documents` page in the dashboard, if you hover over a draft and a completed document, you'll see different URLs.

At the moment, the shortcut tries to go to the following URL for a completed document `/documents/{doc-id}`. 

However, that's the wrong URL, since the URL for a completed doc is `/sign/{token}` when the user is the recipient, not the one that sent the document for signing.

If it's the document owner & the document is completed, the URL is fine as `/documents/{doc-id}`.


dguyen (commented):
It looks like there might be something wrong with your dev environment.

I think things like `session.user.email` should be causing a build error since session can be undefined

Likewise with these functions since no type is provided as a param

```
  const isAccountOwnerRecipient = (document) => {
```

Mythie (approved):

Issue #810, opened 2024-01-08
Field should be dispaled while form is getting submitted


plxity:
### Issue Description

Currently, the input fields don't get disabled while the form is getting submitted, which allows users to update values while the request is being processed

### Steps to Reproduce

1. Login to the application
2. Visit `settings/profile`
3. Click on `Update Profile` CTA and change the value in between

### Expected Behavior

The input fields should be disabled while the request is ongoing

### Current Behavior

The input fields are not disabled while the request is ongoing

### Screenshots (optional)

https://github.com/documenso/documenso/assets/28510494/2384e915-0af0-445c-99c6-bfd2f9d06ed5

### Operating System [e.g., Windows 10]

MacOS

### Browser [e.g., Chrome, Firefox]

Chrome

### Version [e.g., 2.0.1]

120.0.6

### Please check the boxes that apply to this issue report.

- [X] I have searched the existing issues to make sure this is not a duplicate.
- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I have included any relevant screenshots.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [X] I want to work on creating a PR for this issue if approved

plxity:
@catalinpit Would like to know your opinion on this.

catalinpit:
Hey @plxity, thanks for opening this issue! It's a good catch. I agree that the signature pad should be disabled while the form is submitting.

hiteshwadhwani:
Is anyone working on this issue?

plxity:
@hiteshwadhwani Yes, I’m.

hiteshwadhwani:
ok @plxity

Mythie:
This just needs a `<fieldset>` wrapped over it, you can find examples in our other forms 💪

plxity:
@Mythie Thanks for the help. Apologies for delaying it, I was involved with some personal work.

plxity:
@Mythie `fieldset` approach didn't work, I have used a different approach.

roae:
@plxity @Mythie adding `pointer-events-none` class in containerClassName prop could work

plxity:
@roae Agreed, this is a cleaner approach. Thanks for the suggestion. @Mythie Can you please review the changes - #842

PR #842, opened 2024-01-19
fix: disabled signing pad when submitting form

plxity:
fixes : #810


apps/web/src/components/forms/profile.tsx
```diff
@@ -122,7 +121,10 @@ export const ProfileForm = ({ className, user }: ProfileFormProps) => {
                 <FormControl>
                   <SignaturePad
                     className="h-44 w-full"
-                    containerClassName="rounded-lg border bg-background"
+                    containerClassName={cn(
+                      'rounded-lg border bg-background',
+                      isSubmitting ? 'pointer-events-none opacity-50' : null,
```
Mythie, line 126: This can just be `isSubmitting && 'pointer-events-none opacity-50` 😄

dguyen: IMO we should pass a `disabled` prop to the `SignatuePad` which handles it in there for reusability purposes instead of just having it for the form profile

Similar to how it was done before in this pr

plxity: @Mythie Using `cn` makes the code cleaner and more readable I guess. 
@dguyen I agreed tho, @Mythie Should we have both of the methods integrated? What do you suggest?

dguyen: What do you mean both methods? Wouldn't there only be one?

plxity: I meant, reducing the opacity and disabling the pointer events vis CSS and programmatically disabling javascript events to be on the safer side.

dguyen: It should be fine to just do it via CSS since this isn't really critical

plxity: sure, then I belive this PR looks good to me.

dguyen (approved):

PR #846, opened 2024-01-21
feat: account deletion

dephraiim:


packages/lib/server-only/2fa/verify-2fa-token.ts
```diff
@@ -17,6 +17,7 @@ export const verifyTwoFactorAuthenticationToken = async ({
   user,
   totpCode,
 }: VerifyTwoFactorAuthenticationTokenOptions) => {
+  // TODO: This is undefined and I can't figure out why.
```
dephraiim, line 20: @Mythie

dephraiim:
@coderabbitai review

packages/ee/server-only/stripe/delete-customer.ts
```diff
@@ -0,0 +1,10 @@
+import { stripe } from '@documenso/lib/server-only/stripe';
+import type { User } from '@documenso/prisma/client';
+
+export const deleteStripeCustomer = async (user: User) => {
+  if (!user.customerId) {
+    return null;
+  }
+
+  return await stripe.customers.del(user.customerId);
```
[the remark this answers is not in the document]
Mythie: I would say we probably don't need to delete the stripe customer since we may still require it for potential accounting/fraud purposes.

packages/lib/server-only/2fa/verify-2fa-token.ts
Mythie: This will be a build time thing, latest changes on main should deal with it I believe.

packages/lib/server-only/user/delete-user.ts
```diff
@@ -13,10 +13,33 @@ export const deleteUser = async ({ email }: DeleteUserOptions) => {
     },
   });
 
+  const defaultDeleteUser = await prisma.user.findFirst({
```
Mythie, line 16: probably make this `deletedServiceAccount` or similar. We could probably also store this as a method in our lib or something.

So then we would do something like `const serviceAccount = await deletedServiceAccount();` 

Could be nice 🤔

packages/prisma/migrations/20240205120648_create_delete_account/migration.sql
```diff
@@ -0,0 +1,25 @@
...
+  (
+    'deleted@documenso.com',
+    '2024-02-05 11:58:39.668 UTC',
+    NULL,
+    '2024-02-05 11:58:39.670 UTC',
+    '2024-02-05 11:58:39.670 UTC',
+    '2024-02-05 11:58:39.670 UTC',
+    ARRAY['USER'::TEXT]::"public"."Role" [],
+    CAST('GOOGLE'::TEXT AS "public"."IdentityProvider"),
+    FALSE
```
Mythie, line 24: Spicy but looks correct. We can probably use `NOW()` though instead of an explicit datetime.

packages/prisma/migrations/20240205120648_create_delete_account/migration.sql
```diff
@@ -0,0 +1,25 @@
...
+  )
```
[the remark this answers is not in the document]
Mythie: Also relevant I guess

packages/trpc/server/auth-router/schema.ts
```diff
@@ -26,3 +26,9 @@ export const ZSignUpMutationSchema = z.object({
 export type TSignUpMutationSchema = z.infer<typeof ZSignUpMutationSchema>;
 
 export const ZVerifyPasswordMutationSchema = ZSignUpMutationSchema.pick({ password: true });
+
+export const ZDeleteAccountMutationSchema = z.object({
+  email: z.string().email(),
```
Mythie, line 31: Would we need this? Presumably it can be argless mutation and we get the user from the context value, this means there's no chance of accidentally deleting someone elses account.

We would probably also have a version in the admin router later on for us to offboard users who don't wish to delete their own account.

Mythie (changes requested):

packages/prisma/migrations/20240205120648_create_delete_account/migration.sql
dephraiim: Will take care of that

packages/trpc/server/auth-router/schema.ts
dephraiim: Right, yeah, I'll go ahead without this

packages/lib/server-only/2fa/verify-2fa-token.ts
dephraiim: Merged the latest changes on main and it is still undefined. It is wierd

apps/web/src/components/forms/profile.tsx
```diff
@@ -85,6 +113,61 @@ export const ProfileForm = ({ className, user }: ProfileFormProps) => {
...
+      });
+
+      await deleteAccount();
+
+      toast({
+        title: 'Account deleted',
+        description: 'Your account has been deleted successfully.',
+        duration: 5000,
+      });
+
```
Mythie, line 150: Seems like this could be flipped around since it repeats itself?

packages/lib/server-only/2fa/verify-2fa-token.ts
Mythie: :S

packages/lib/server-only/user/delete-user.ts
```diff
@@ -13,10 +13,33 @@ export const deleteUser = async ({ email }: DeleteUserOptions) => {
...
 
+  if (!defaultDeleteUser) {
+    throw new Error(`Default delete account not found`);
+  }
+
+  await prisma.document.updateMany({
+    where: {
+      userId: user.id,
+      status: {
+        in: ['PENDING', 'COMPLETED'],
```
Mythie, line 34: Prefer to use the enums here

Mythie (approved):

PR #709, opened 2023-12-01
fix: visibility of update password & security fields using identityprovider

anikdhabal:
fixes #690 
Screenshot 2023-12-01 233708

PR #856, opened 2024-01-23
fix: auto verify google sso emails

dguyen:
## Description

Currently logging in via Google SSO will still require you to verify your email. This shouldn't happen since going through Google SSO implies you have access to that email.

## Changes Made

This change resolves that by setting the `emailVerified` attribute in the database when you access the application via Google SSO.

This also updates the identity provider for users who log in through Google to have it correctly set as `GOOGLE`.

## Testing Performed

- Logged in via SSO and no email verification was requested
- Logged in via non SSO and email verification was requested

## Checklist

### DIFF
--- a/packages/lib/next-auth/auth-options.ts
+++ b/packages/lib/next-auth/auth-options.ts
@@ -9,6 +9,7 @@ import type { GoogleProfile } from 'next-auth/providers/google';
 import GoogleProvider from 'next-auth/providers/google';
 
 import { prisma } from '@documenso/prisma';
+import { IdentityProvider } from '@documenso/prisma/client';
 
 import { isTwoFactorAuthenticationEnabled } from '../server-only/2fa/is-2fa-availble';
 import { validateTwoFactorAuthentication } from '../server-only/2fa/validate-2fa';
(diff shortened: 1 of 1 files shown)


packages/lib/next-auth/auth-options.ts
```diff
@@ -138,6 +138,23 @@ export const NEXT_AUTH_OPTIONS: AuthOptions = {
         merged.emailVerified = user.emailVerified?.toISOString() ?? null;
       }
 
+      if (
+        (trigger === 'signIn' || trigger === 'signUp') &&
+        merged.emailVerified === null &&
+        account?.provider === 'google'
```
Mythie, line 144: Does this work? If it does that's amazing ❤️

dguyen: Pho sho 🙏

dguyen: I just realized that we're not using setting `identityProvider` on the user table to `GOOGLE` 

Should I just slide that in here? We need it set correctly for this PR: #709

dguyen:
Applied an update so that whenever the user signs in or signs up via Google, we will update the user identity provider to be `GOOGLE`.

We can't do this conditionally, so this will happen every time during sign in or sign up via Google SSO.

packages/lib/next-auth/auth-options.ts
```diff
@@ -138,6 +139,22 @@ export const NEXT_AUTH_OPTIONS: AuthOptions = {
         merged.emailVerified = user.emailVerified?.toISOString() ?? null;
       }
 
+      if ((trigger === 'signIn' || trigger === 'signUp') && account?.provider === 'google') {
```
dguyen, line 142: It's not optimal to run this every sign up or sign in via Google, but we don't have access to `identityProvider` to see if it's correctly set.

Mythie: We could easily skip out if we know the user is already verified no?

dguyen: It would, but we don't know if the provider is correctly set on the user table

Mythie: Slap it in the token while we're here, this saves tripping to the db all the time 😄

Mythie: Actually don't worry about it, we trip enough as is and this is only on signIn or signUp

Mythie (approved):

Issue #840, opened 2024-01-17
Improved Document Dropzone UI


sumitbishti:
### Improvement Description

Currently the document dropzone ui looks like this on hover.

<img width="942" alt="Screenshot 2024-01-17 234237" src="https://github.com/documenso/documenso/assets/75713174/0090e741-1e1c-4ccf-913f-eef1d9602814">

### Rationale

It doesn't look good on hover

### Proposed Solution

<img width="929" alt="Screenshot 2024-01-17 235850" src="https://github.com/documenso/documenso/assets/75713174/91f782da-927d-4036-ad1e-8db007aff5fd">

### Please check the boxes that apply to this improvement suggestion.

- [x] I have searched the existing issues and improvement suggestions to avoid duplication.
- [x] I have provided a clear description of the improvement being suggested.
- [x] I have explained the rationale behind this improvement.
- [x] I have included any relevant technical details or design suggestions.
- [x] I understand that this is a suggestion and that there is no guarantee of implementation.

sumitbishti:
I would like to take this up please.

Mythie:
This is dependent on screen size and other aspects, if we're going to fix this we should consider how we avoid a bandaid fix so it works across a variety of resolutions.

If you're happy to do this I'm alright with assigning it to you 🙌

sumitbishti:
Yes please, I'll do it. 
What I am going to do is check it for various resolutions and screen sizes and maybe browsers. 
And target the ones which it is not working for.

Anything else that I might be missing?

PR #857, opened 2024-01-23
fix: improved document-dropzone ui for small vertical screens

sumitbishti:
improved document-dropzone ui for small vertical screens (screens less than 800px vertically)
Although it can still become congested on really small vertical screens, but possibility is really low.

fixes: #840

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/upload-document.tsx
+++ b/apps/web/src/app/(dashboard)/documents/upload-document.tsx
@@ -96,10 +96,12 @@ export const UploadDocument = ({ className }: UploadDocumentProps) => {
     }
   };
 
+  const isSmallVerticalScreen = typeof window !== 'undefined' && window.innerHeight < 800;
+
   return (
     <div className={cn('relative', className)}>
       <DocumentDropzone
-        className="min-h-[40vh]"
+        className={`${isSmallVerticalScreen ? 'min-h-[50vh]' : 'min-h-[40vh]'}`}
         disabled={remaining.documents === 0 || !session?.user.emailVerified}
         disabledMessage={disabledMessage}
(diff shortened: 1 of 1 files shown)


apps/web/src/app/(dashboard)/documents/upload-document.tsx
```diff
@@ -96,10 +96,12 @@ export const UploadDocument = ({ className }: UploadDocumentProps) => {
     }
   };
 
+  const isSmallVerticalScreen = typeof window !== 'undefined' && window.innerHeight < 800;
+
   return (
     <div className={cn('relative', className)}>
       <DocumentDropzone
-        className="min-h-[40vh]"
+        className={`${isSmallVerticalScreen ? 'min-h-[50vh]' : 'min-h-[40vh]'}`}
```
Mythie, line 104: I reckon we could something closer to `min-h-[max(400px,40vh)` or similar to achieve the desired effect without javascript 😄

sumitbishti: Actually i want to set the min height to 50% of the vh in smaller vertical screens, and not a constant number like 400px here. Plus i tried doing ```min-h-[max(400px,40vh)]```, it didn't worked!

Other thing we could do is set the min height to ```50vh``` instead of ```40vh``` for all screen sizes.

Mythie: You were absolutely right, it should have been a `min` rather than `max`. I've updated and tested across a variety of screen sizes!

This has the added benefit of not making the upload massive on vertical monitors which is nice 😄

Mythie (approved):

Issue #621, opened 2023-11-02
[IMPROVEMENT]: more specific instructions for adding a document


Dun-sin:
### Improvement Description

I think the following should be added:
- the type of file which can be uploaded
- the limit of the file to be uploaded
![image](https://github.com/documenso/documenso/assets/78784850/48fe0107-905c-4be0-b0a5-c676f2bc995b)

### Rationale

I tried using the app for the first time and the first time and immediately thought to upload msword to check as it ended with .doc extension, so specifying what file format/extension would be accepted.

Adding the limit it has would also help with users avoiding the large file and know beforehand what is required.

### Please check the boxes that apply to this improvement suggestion.

- [X] I have searched the existing issues and improvement suggestions to avoid duplication.
- [X] I have provided a clear description of the improvement being suggested.
- [X] I have explained the rationale behind this improvement.
- [X] I have included any relevant technical details or design suggestions.
- [X] I understand that this is a suggestion and that there is no guarantee of implementation.

Dun-sin:
i would like to work on this if this is a accepted change and any way you want me to do that.

Dun-sin:
> Hey @Dun-sin, thanks for the suggestions! I agree that adding the maximum document limit would be helpful. Regarding the doc type, do people usually sign non-pdfs?

no, they don't i know that, but when you go to image converter site for instance, they still indicate to upload an image, i mean we both know users are a mystery, it can be skipped btw doesn't really matter, the most important in my opinion is the file limit

sumitbishti:
is anyone working on this issue?

dguyen:
Apologies to those involved for the delay, this one seemed to have gotten lost.

@Dun-sin are you still interested in working on this feature? If not I can assign it to @sumitbishti 

To summarise some things:

> I tried using the app for the first time and the first time and immediately thought to upload msword to check as it ended with .doc extension, so specifying what file format/extension would be accepted.

I think changing the text to `Drag & drop your PDF document here.` should suffice.

> Adding the limit it has would also help with users avoiding the large file and know beforehand what is required.

Agreed, I believe that the current upload limit is 50MBs for the S3 upload transport we implemented. Currently if you select a file that exceeds the limit, it will silently error.

Since users would rarely upload a 50mb document, we can keep the UI clean by displaying a toast when they select a file that's too large. The toast should contain the information on the current current file size limit.

This would involve displaying a toast error saying the current file limit size when a file limit size error occurs via `onDropRejected` in `useDropzone`.

[...]

PR #867, opened 2024-01-24
perf: mentioned type and size of the doc to be uploaded

sumitbishti:
explicitly mentioned "PDF" to upload, and added a toast if pdf size is greater than 50mb

fixes: #621


apps/web/src/app/(dashboard)/documents/upload-document.tsx
```diff
@@ -96,13 +103,23 @@ export const UploadDocument = ({ className }: UploadDocumentProps) => {
     }
   };
 
+  const onFileDropRejected = () => {
+    toast({
+      title: 'File is larger than 50mb',
```
dguyen, line 108: Since this limit actually depends on the provider you use, could you create the following environment variable so users can dynamically change it for their use case?

```
# OPTIONAL: Displays the maximum document upload limit to the user in MBs
NEXT_PUBLIC_DOCUMENT_UPLOAD_LIMIT=5
```

Also could you please reword it to be `File cannot be larger than ...`

dguyen (changes requested):

apps/web/src/app/(dashboard)/documents/upload-document.tsx
sumitbishti: Should i also use this env variable in the codebase in place of the limit 50mb? Or just create the env var and leave it like that?

dguyen: Yeah that'll be a good idea

Could you put it as a constant in `lib/constants/app.ts`?

Something like this but it also parses the env into number if possible

```
export const APP_DOCUMENT_UPLOAD_SIZE_LIMIT = process.env.NEXT_PUBLIC_DOCUMENT_SIZE_UPLOAD_LIMIT ?? 50;
```

Mythie: That'll make it `string | number`, we would also need to use whatever value in the `next.config.js` for the body size limits this may also apply to certain API routes.

packages/ui/primitives/document-dropzone.tsx
```diff
@@ -175,10 +187,10 @@ export const DocumentDropzone = ({
           </p>
 
           <p className="text-muted-foreground/80 mt-1 text-sm">
-            {disabled ? disabledMessage : 'Drag & drop your document here.'}
+            {disabled ? disabledMessage : 'Drag & drop your PDF here.'}
```
dephraiim, line 190: Do we need to specify a pdf? Document is fine imo

dguyen: It's done to clarify that we only take in PDFs

apps/web/src/app/(dashboard)/documents/upload-document.tsx
dguyen: Yeah that's why I said it would require parsing

> we would also need to use whatever value in the next.config.js 

Not necessarily since this is purely for document upload limits

sumitbishti: do i need to use the document_size_limit env variable inside config in [trpc].ts files?
```
export const config = {
  maxDuration: 60,
  api: {
    bodyParser: {
      sizeLimit: '50mb',
    },
  },
};
```

dguyen: Currently the file upload transport doesn't go through TRPC, so I don't think it's relevant here

Mythie: It goes through server actions which have a similar limit configured. We also have a maxSize set on the dropzone primitive.

I think the best course of action would be to hardcode this but extract that current mix of magic numbers into a constant somewhere.

packages/lib/constants/app.ts
```diff
@@ -6,3 +6,6 @@ export const APP_FOLDER = IS_APP_MARKETING ? 'marketing' : 'web';
 export const APP_BASE_URL = IS_APP_WEB
   ? process.env.NEXT_PUBLIC_WEBAPP_URL
   : process.env.NEXT_PUBLIC_MARKETING_URL;
+
+export const APP_DOCUMENT_UPLOAD_SIZE_LIMIT =
+  Number(process.env.NEXT_PUBLIC_DOCUMENT_SIZE_UPLOAD_LIMIT) ?? 50;
```
dguyen, line 11: I don't think this will work as expected

```
Number(undefined) ?? 50 // Becomes NaN
```

I think you meant something like this?

```
Number(undefined) || 50 // Becomes 50
```

packages/ui/primitives/document-dropzone.tsx
```diff
@@ -10,6 +10,8 @@ import { megabytesToBytes } from '@documenso/lib/universal/unit-convertions';
 import { cn } from '../lib/utils';
 import { Card, CardContent } from './card';
 
+import {APP_DOCUMENT_UPLOAD_SIZE_LIMIT} from'@documenso/lib/constants/app';
```
dguyen, line 13: This should be causing linting issues, is your workspace configured correctly?

I also noticed some files in this PR are having final whitespace is being removed, which shouldn't happen if the environment is correctly configured

dguyen (changes requested):

packages/ui/primitives/document-dropzone.tsx
sumitbishti: I need help in this issue!

dguyen: Try restart vscode first if you haven't yet done so

When you go to this file, it should show up as linting errors

image

Have a look in the vscode `problems` tab to see if it's working correctly

Also have a look bottom right of vscode to see if prettier is working, it should show up and have a check next to it

image

apps/web/src/app/(dashboard)/documents/upload-document.tsx
```diff
@@ -96,13 +98,23 @@ export const UploadDocument = ({ className }: UploadDocumentProps) => {
     }
   };
 
+  const onFileDropRejected = () => {
+    toast({
+      title: `File cannot be larger than ${APP_DOCUMENT_UPLOAD_SIZE_LIMIT}mb`,
```
catalinpit, line 103: I'd swap the "title" message with the "description" message.

packages/lib/constants/app.ts
```diff
@@ -6,3 +6,6 @@ export const APP_FOLDER = IS_APP_MARKETING ? 'marketing' : 'web';
 export const APP_BASE_URL = IS_APP_WEB
   ? process.env.NEXT_PUBLIC_WEBAPP_URL
   : process.env.NEXT_PUBLIC_MARKETING_URL;
+
+export const APP_DOCUMENT_UPLOAD_SIZE_LIMIT =
+  Number(process.env.NEXT_PUBLIC_DOCUMENT_SIZE_UPLOAD_LIMIT) || 50;
```
catalinpit, line 11: I don't think the empty lines from the end should be deleted

apps/web/src/app/(dashboard)/documents/upload-document.tsx
catalinpit: Also, it should probably be `MB`

catalinpit (changes requested):
Looks good, thanks for the work!

catalinpit (approved):

dguyen (approved):

Issue #864, opened 2024-01-24
Undo option on signing pad not working correctly


plxity:
### Issue Description

On clicking undo icon, it clears the whole canvas instead of removing the recent changes

### Steps to Reproduce

1. https://app.documenso.com/settings/profile
2. Change signature
3. Click on Undo

### Expected Behavior

It should only undo the recent changes

### Current Behavior

It clears the complete canvas

### Screenshots (optional)

https://github.com/documenso/documenso/assets/28510494/c8afad61-bde1-4be4-b81b-b4bba25799a7

### Please check the boxes that apply to this issue report.

- [X] I have searched the existing issues to make sure this is not a duplicate.
- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I have included any relevant screenshots.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [X] I want to work on creating a PR for this issue if approved

plxity:
@Mythie @catalinpit let me know what you folks think about this issue.

Gautam-Hegde:
I think, undo functions normally when a signature stroke happens inside the sign-canvas area unless the user tries to cross over the canvas it behaves like this.

Gautam-Hegde:
> @Gautam-Hegde I don't think so, the UX pattern involved with undo functionality remains the same -- Undo the previous event that happened, so if a stoke happens, undo button should just reverse it when click


https://github.com/documenso/documenso/assets/85569489/b03c3f6f-19f3-4833-82ec-f910b4c0b30e

plxity:
> > @Gautam-Hegde I don't think so, the UX pattern involved with undo functionality remains the same -- Undo the previous event that happened, so if a stoke happens, undo button should just reverse it when click
> 
>  issue.864.mp4

Please watch the video I've attached to the issue. When I log in and navigate to `https://app.documenso.com/settings/profile`, add a stroke to the canvas and click on undo, it shouldn't clear the whole signature on the canvas. Instead, it should just remove the stroke that I just added.

Gautam-Hegde:
> > > @Gautam-Hegde I don't think so, the UX pattern involved with undo functionality remains the same -- Undo the previous event that happened, so if a stoke happens, undo button should just reverse it when click
> > 
> > 
> > issue.864.mp4
> 
[...]

PR #868, opened 2024-01-25
fix: undo operation on signature pad

plxity:
fixes: #864


plxity:
@Mythie @dguyen Please review the changes.

adithyaakrishna (approved):
LGTM 🌮

plxity:
@Mythie Bumping this up 😅

catalinpit:
I'm on the `main` branch, and the undo functionality works properly. Am I missing something?

https://github.com/documenso/documenso/assets/25515812/b1a94c93-39fa-452c-a08b-6fd7f590a088

plxity:
@catalinpit Please watch the video I have attached and the steps to reproduce.

catalinpit:
> @catalinpit Please watch the video I have attached and the steps to reproduce.

Yeah, I've seen those. I thought it happens in all scenarios, but the issue actually occurs when you already have a signature and try to update it.



Thanks for the heads up! Will test the changes and review the PR.

catalinpit:
Ok, I just tested it. It works fine on the profile page, but for some reason doesn't work when signing a document. I attached a video.

Does it work for you?

https://github.com/documenso/documenso/assets/25515812/4bbb2142-e5f9-4696-bc8f-b4a575a94f72

plxity:
Yup, it seems to be working for me. Can you restart server or clear browser cache and try again?

catalinpit:
> Yup, it seems to be working for me. Can you restart server or clear browser cache and try again?

I restarted the server, cleared browser cache, and even ran `npm run reset:hard`, and it still doesn't work unfortunately...

plxity:
@catalinpit My bad, I have fixed it. In `useEffect` deps value, there is no need to add `defaultValue`. Because we want that effect to only run once when the component loads, I've fixed it and have tested the changes at all the places. Please review it again.

packages/ui/primitives/signature-pad/signature-pad.tsx
```diff
@@ -183,8 +184,11 @@ export const SignaturePad = ({
     // Clear the canvas
     if ($el.current) {
       const ctx = $el.current.getContext('2d');
+      const { width, height } = $el.current;
```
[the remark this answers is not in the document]
plxity: Fixed it, I think we can add eslint error for an unused variable😅

packages/ui/primitives/signature-pad/signature-pad.tsx
```diff
@@ -183,8 +184,11 @@ export const SignaturePad = ({
     // Clear the canvas
     if ($el.current) {
```
catalinpit, line 185: Another thing I observed is that when you undo the first change to an already existing signature, the signature is not updated. See the video to better understand what I mean.

https://github.com/documenso/documenso/assets/25515812/281b776d-8366-4116-8bb9-0df5d6f1cfdb

I tried calling the `onChange` function after clearing the canvas to update the signature URL and it seems to work.

```js
if ($el.current) {
      const ctx = $el.current.getContext('2d');
      const { width, height } = $el.current;
      ctx?.clearRect(0, 0, width, height);
      if (typeof defaultValue === 'string' && defaultImageRef.current) {
        ctx?.putImageData(defaultImageRef.current, 0, 0);
      }
      newLines.forEach((line) => {
        const pathData = new Path2D(getSvgPathFromStroke(getStroke(line, perfectFreehandOptions)));
        ctx?.fill(pathData);
      });

      /*
         here
      */
      onChange?.($el.current.toDataURL());
    }
```

Please double check and see if it works and/or if there is a better way of doing it.

Here's how it works with the new addition.

https://github.com/documenso/documenso/assets/25515812/10a95739-e0b5-4889-894b-5d57df96c136

catalinpit (changes requested):

packages/ui/primitives/signature-pad/signature-pad.tsx
plxity: @catalinpit Thanks for pointing out. This makes sense, I've added an onChange handler. This is a good method to work with it :)

plxity: @catalinpit Let me know if the changes are fine :)

catalinpit (dismissed):
thanks for the work!

Mythie (approved):

Mythie:
Great work 😄

PR #876, opened 2024-01-26
fix: fix header layout shift

### LINKED ISSUE #875: Layout Shift between views
### Issue Description

- The main content area layout jumps when switching a bit between tabs (Documents, Templates)

### Steps to Reproduce

- Just log in and switch between the tabs
- Watch the Logo

### Expected Behavior

_No response_

### Current Behavior

- Switching between the tabs should only change the content, the header bar should remain static
- The content should be the same within both views

### Screenshots (optional)

https://github.com/documenso/documenso/assets/1309312/7e2d3dd5-aa73-44ad-968b-7a2591c9a330

MacOS Ventura 13.3.1

Chrome

### Version [e.g., 2.0.1]

20.0.6099.234 (Official Build) (arm64)

- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.

ksushant6566:
fix #875 

The problem was due to the scroll bar. Solution is to use ``vw`` unit for header as they are not affected by scroll-bars.

Loom: https://www.loom.com/share/057755740fca48c18bccb08912b65f5c?sid=3b3b674c-3991-45bb-871d-f7795a145525

### DIFF
--- a/apps/web/src/components/(dashboard)/layout/header.tsx
+++ b/apps/web/src/components/(dashboard)/layout/header.tsx
@@ -33,13 +33,13 @@ export const Header = ({ className, user, ...props }: HeaderProps) => {
   return (
     <header
       className={cn(
-        'supports-backdrop-blur:bg-background/60 bg-background/95 sticky top-0 z-[50] flex h-16 w-[calc(100vw-8rem)] items-center border-b border-b-transparent backdrop-blur duration-200',
+        'supports-backdrop-blur:bg-background/60 bg-background/95 sticky top-0 z-[50] flex h-16 w-[100vw] items-center border-b border-b-transparent backdrop-blur duration-200 lg:w-[calc(100vw-2rem)] lg:pl-4',
         scrollY > 5 && 'border-b-border',
         className,
       )}
       {...props}
     >
-      <div className="mx-auto flex w-full max-w-screen-xl items-center justify-between gap-x-4 px-4 md:justify-normal md:px-8">
+      <div className="mx-auto flex w-full max-w-screen-xl items-center justify-between gap-x-4 px-4 md:px-8">
         <Link
           href="/"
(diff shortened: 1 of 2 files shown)


ksushant6566:
https://www.loom.com/share/3a7c18a8f9364c81a9f08a0f9d96e097?sid=a3b1eef5-f802-4553-9c42-7ad9778b21cf

The header breaks the layout for Ipad screen, fixing this also.

dguyen:
This change seems to cause issues where the header is no longer aligned with the content

image

Looking at this article it seems like `scrollbar-gutter: stable` works good in firefox/chrome on macos (doesn't work in safari)

https://dev.to/rashidshamloo/preventing-the-layout-shift-caused-by-scrollbars-2flp

I haven't tried 3rd or 4th options in that article though, which may be better.

apps/web/src/components/(dashboard)/layout/header.tsx
```diff
@@ -33,13 +33,13 @@ export const Header = ({ className, user, ...props }: HeaderProps) => {
   return (
     <header
       className={cn(
-        'supports-backdrop-blur:bg-background/60 bg-background/95 sticky top-0 z-[50] flex h-16 w-full items-center border-b border-b-transparent backdrop-blur duration-200',
+        'supports-backdrop-blur:bg-background/60 bg-background/95 sticky top-0 z-[50] flex h-16 w-[100vw] items-center border-b border-b-transparent backdrop-blur duration-200 lg:w-[calc(100vw-2rem)] lg:pl-4',
```
Mythie, line 36: This seems like it'll potentially cause a horizontal scrollbar due to the fixed sizing, the `lg` breakpoint class also feels like a hack which would be best to avoid 😬

dguyen:
Going to close this due to inactivity for now

PR #879, opened 2024-01-28
fix: sign dialog completed title color in dark mode

dephraiim:


apps/web/src/app/(signing)/sign/[token]/sign-dialog.tsx
```diff
@@ -45,7 +45,9 @@ export const SignDialog = ({
       </DialogTrigger>
       <DialogContent>
         <div className="text-center">
-          <div className="text-xl font-semibold text-neutral-800">Sign Document</div>
+          <div className="dark:text-muted-foreground text-xl font-semibold text-neutral-800 ">
```
Comment (dguyen) [inline apps/web/src/app/(signing)/sign/[token]/sign-dialog.tsx:48]: Can you replace all text colours so it's just `text-foreground` for this element?

dguyen (approved):

PR #881, opened 2024-01-29
fix: show fields on every step while editing documents

dephraiim:
CleanShot 2024-01-29 at 00 51 31@2x

CleanShot 2024-01-29 at 00 51 39@2x


Mythie:
Looks like SubjectFieldItem is fairly generic, I'd rename it as such and then also show this on the other steps that don't involve editing fields 😄

dephraiim:
@Mythie I included it for all the other sections except the title section. This line throws an error everytime on the title section with error `document not found1

https://github.com/documenso/documenso/blob/f8125aec54fe864b9afe166dca95cf76dc52bdf0/packages/ui/primitives/document-flow/show-field-item.tsx#L89

Mythie:
lgtm

Issue #891, opened 2024-01-31
Disable signing pad while submitting form (Related to #810)


plxity:
### Issue Description

This issue is related to #810. I just realised we are using a signing pad in the form at a couple of places and I missed adding a disable state there. 

### Steps to Reproduce

Same as #810 

### Expected Behavior

The signing pad should be disabled while submitting form 

### Current Behavior

Signing pad is not disabled

### Please check the boxes that apply to this issue report.

- [X] I have searched the existing issues to make sure this is not a duplicate.
- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I have included any relevant screenshots.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [X] I want to work on creating a PR for this issue if approved

PR #892, opened 2024-01-31
fix: Added signing pad disable state while submitting form

plxity:
Fixes : #891

### DIFF
--- a/apps/marketing/src/components/(marketing)/widget.tsx
+++ b/apps/marketing/src/components/(marketing)/widget.tsx
@@ -399,6 +399,7 @@ export const Widget = ({ className, children, ...props }: WidgetProps) => {
           </DialogDescription>
 
           <SignaturePad
+            containerClassName={isSubmitting ? 'pointer-events-none opacity-50' : undefined}
             className="aspect-video w-full rounded-md border"
             defaultValue={signatureDataUrl || ''}
             onChange={setDraftSignatureDataUrl}
--- a/apps/marketing/src/components/(marketing)/widget.tsx
+++ b/apps/marketing/src/components/(marketing)/widget.tsx
@@ -399,7 +399,7 @@ export const Widget = ({ className, children, ...props }: WidgetProps) => {
           </DialogDescription>
 
           <SignaturePad
-            isDisabled={isSubmitting}
+            disabled={isSubmitting}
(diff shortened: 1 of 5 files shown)


apps/marketing/src/components/(marketing)/widget.tsx
```diff
@@ -399,6 +399,7 @@ export const Widget = ({ className, children, ...props }: WidgetProps) => {
           </DialogDescription>
 
           <SignaturePad
+            containerClassName={isSubmitting ? 'pointer-events-none opacity-50' : undefined}
```
dguyen, line 402: Please create a `disabled` prop on the SignaturePad and update the logic there.

plxity: @dguyen I've added `isDisabled` props on the signing pad and have fixed the usage at all the places. Please review.

dguyen: Can you please use `disabled` so it matches normal input elements

packages/ui/primitives/signature-pad/signature-pad.tsx
```diff
@@ -214,7 +216,13 @@ export const SignaturePad = ({
...
   return (
-    <div className={cn('relative block', containerClassName)}>
+    <div
+      className={cn(
+        'relative block',
+        containerClassName,
+        isDisabled ? 'pointer-events-none opacity-50' : null,
+      )}
+    >
       <canvas
```
dguyen, line 226: You can do something like this to make it neater

```
      className={cn('relative block', containerClassName, {
        'pointer-events-none opacity-50': disabled,
      })}
```

plxity: @dguyen Done

dguyen: Cheers!

Did you see the other note about using `disabled` instead of `isDisabled`?

dguyen: `disabled` matches how we disable inputs, buttons, fields.

`isSubmitting` is simply a variable we pass around.

plxity: Sure, made the changes. Please review

Issue #890, opened 2024-01-31
Clicking "Documenso logo" or "Documents" at the top left doesn't change the active tab to "ALL"


sumitbishti:
### Issue Description

While you are at a certain section other than "ALL", click Documenso logo or Documents at the top left you will see all the documents listed below, but the active tab remains the same what was previously selected.

Another instance of this is occurring when you click on the search bar and click on the sections present there, active tab is not getting updated.

### Steps to Reproduce

While you are at a certain section other than "ALL", click Documenso logo or Documents at the top left you will see all the documents listed below, but the active tab remains the same what was previously selected.

Search documents thorough the search bar and click on the already mentioned sections.

### Expected Behavior

The active tab should change to "ALL".

### Current Behavior

Active tab remains the same what was previously selected.

### Screenshots (optional)

All the documents are listed, but the active tab hasn't changed.

<img width="926" alt="Screenshot 2024-01-31 193402" src="https://github.com/documenso/documenso/assets/75713174/ac89d871-152a-4e9f-bef7-181e59a2cf0b">

### Operating System [e.g., Windows 10]

Windows 11

### Browser [e.g., Chrome, Firefox]

Chrome

### Please check the boxes that apply to this issue report.

- [X] I have searched the existing issues to make sure this is not a duplicate.
- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I have included any relevant screenshots.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [X] I want to work on creating a PR for this issue if approved

sumitbishti:
Would like to fix this.

PR #896, opened 2024-02-01
fix: active-tab correctly changes to the currently active documents section

sumitbishti:
fixes: #890


packages/eslint-config/index.cjs
sumitbishti, line 1: Oops! These changes shouldn't be here...

PR #898, opened 2024-02-01
fix: Prevent users from bypassing document limitations via templates

### LINKED ISSUE #883: Free document sending limitation bypass
### Issue Description

I exactly don't know whether it's bug or a feature. The application fails to enforce the restriction on sending only 5 free documents per month effectively. **Users can bypass this limitation by creating new templates and sending documents through them**, without encountering any restrictions regarding their usage quota.

**Despite reaching the 5-document limit, the application allows the user to continue sending additional documents throw templates without any restriction.**

### Steps to Reproduce

1. Log in to the application
2. Crose the limitation of sending 5 documents in the free tire.
3. Then go to the templates page.
4. Create a new template.
5. Send the template document to signers.
6. Then if you want you can delete the template
7. Notice that the application does not display any warnings, errors, or notifications regarding exceeding the free document limit.

### Expected Behavior

The application should enforce the limitation on sending only 5 free documents including template documents also per month per user.

### Current Behavior

The application fails to enforce the free document limitation effectively. Users can easily bypass the restriction by creating new templates and sending documents through them without encountering any restrictions regarding their usage quota.

### Screenshots (optional)

https://github.com/documenso/documenso/assets/87828904/2840abb6-534a-452e-a185-39a4306669c5

Windows 11

Chrome, Edge

### Version [e.g., 2.0.1]

Latest

- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.

ashrafchowdury:
## Description

**Fixed document limitation bypassing issues through templates.** Previously, users could bypass document restrictions by utilizing templates even after reaching their limitations. This fix ensures that templates will no longer function as a workaround when users reach their document limits.

## Changes
1. imported `useLimits` hook on `data-table-templates.tsx`
2. Disabled the 'Use Template' button when the user reaches their limit.
3. Added an Alert Component on top of the templates page to notify users that they can't use templates anymore because they have reached their limit.
4. Used `getServerLimits` hook on `template-router` to a condition on the server.

## Example
image

## Issue
Closes #883


packages/trpc/server/template-router/router.ts
```diff
@@ -41,6 +42,13 @@ export const templateRouter = router({
       try {
         const { templateId } = input;
 
+        const limits = await getServerLimits({ email: ctx.user.email });
+
+        if (limits.remaining.documents === 0) {
+          throw new Error('You have reached your document limit.');
+          return;
```
dguyen, line 49: I'm assuming the reason you did 

```
      const documentFromTemplate = await createDocumentFromTemplate({
        templateId,
        templateId,
      });
```

In the prior code is since it could be undefined?

The undefined is being caused by this redundant return statement because of the throw error.

If you remove this return, you can revert the old code back to use `{ id }`

dguyen (changes requested):

packages/trpc/server/template-router/router.ts
ashrafchowdury: Ohh, I see.

dguyen (approved):
Looks good to me, thanks :)

PR #899, opened 2024-02-02
feat: add transparent background to fields

dephraiim:


packages/ui/primitives/document-flow/add-fields.tsx
```diff
@@ -316,7 +316,7 @@ export const AddFieldsFormPartial = ({
           {selectedField && (
             <Card
               className={cn(
-                'bg-background pointer-events-none fixed z-50 cursor-pointer transition-opacity',
+                'bg-documenso pointer-events-none fixed z-50 cursor-pointer border-2 bg-opacity-20 backdrop-blur-[1.2px]',
```
catalinpit, line 319: I find it quite difficult to see the fields properly with these values. What do you think?

CleanShot 2024-02-02 at 14 46 39

I played a bit with the values, and managed to get to this version. What do you think?

```
bg-documenso h-full w-full bg-opacity-80 backdrop-blur-[1px]
```

CleanShot 2024-02-02 at 14 51 38

Or, here's another example (maybe closer to the design spec?).

```
bg-documenso/60 h-full w-full bg-opacity-80 backdrop-blur-[1px]
```

CleanShot 2024-02-02 at 14 55 40

Anyway, just my thoughts. Feel free to discard them. 😆

dguyen: I agree with Catalin, it's hard to see

IMO transparent when dragging and not transparent when dropped would be optimal

adithyaakrishna: Idk why but the design one seems to be way better. I feel the background values would be good in the ranges of 30-40 😶

catalinpit: Yup, the design one looks the best. Couldn't replicate it 1-1 though.

At the moment, it's a bit difficult to see the field text when it's placed on top of the component. Any ideas on how we can overcome this?

Other than that, it looks good to me.

CleanShot 2024-02-19 at 09 36 43@2x

Mythie (approved):

PR #918, opened 2024-02-07
fix: recipients with CC role not being editable

dguyen:
## Description

Fixed issue where setting a recipient role as CC will prevent any further changes as it is considered as "sent" and "signed".

## Other changes

- Prevent editing document after completed
- Removed CC and Viewers from the field recipient list since they will never be filled
- Minor UI issues

## Checklist

- [X] I have added/updated tests that prove the effectiveness of these changes.

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
+++ b/apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
@@ -114,7 +114,7 @@ export const DataTableActionDropdown = ({ row, team }: DataTableActionDropdownPr
       <DropdownMenuContent className="w-52" align="start" forceMount>
         <DropdownMenuLabel>Action</DropdownMenuLabel>
 
-        {recipient?.role !== RecipientRole.CC && (
+        {recipient && recipient?.role !== RecipientRole.CC && (
           <DropdownMenuItem disabled={!recipient || isComplete} asChild>
             <Link href={`/sign/${recipient?.token}`}>
               {recipient?.role === RecipientRole.VIEWER && (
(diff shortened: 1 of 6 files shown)


packages/ui/primitives/document-flow/add-fields.tsx
```diff
@@ -304,6 +304,13 @@ export const AddFieldsFormPartial = ({
     return recipientsByRole;
   }, [recipients]);
 
+  const recipientsByRoleToDisplay = useMemo(() => {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    return (Object.entries(recipientsByRole) as [RecipientRole, Recipient[]][]).filter(
+      ([role]) => role !== RecipientRole.CC && role !== RecipientRole.VIEWER,
+    );
+  }, [recipientsByRole]);
```
dguyen, line 312: Done to pull logic out and provide types here instead of in template

apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
```diff
@@ -114,7 +114,7 @@ export const DataTableActionDropdown = ({ row, team }: DataTableActionDropdownPr
       <DropdownMenuContent className="w-52" align="start" forceMount>
         <DropdownMenuLabel>Action</DropdownMenuLabel>
 
-        {recipient?.role !== RecipientRole.CC && (
+        {recipient && recipient?.role !== RecipientRole.CC && (
```
Mythie, line 117: 🤔

Mythie (commented):
Silly question, could we not just remove the current ternaries when setting the recipients for the document and instead just add an `update` call in the `sendDocument` method that marks all CC'd as sent and signed?

apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
dguyen: Prior to this change, this will render a blank area if the user owns the document but is none of the recipients

PR #921, opened 2024-02-08
fix: table layout shift while changing tabs

plxity:
fixes: #875 

https://github.com/documenso/documenso/assets/28510494/083fd87a-ef62-40e6-9696-9c04b4411502


plxity:
@dguyen Please review this PR

dguyen:
Please don't ping maintainers directly unless the PR is stale, we actively monitor the issues and PRs and will get around to them when possible

plxity:
Noted.

packages/ui/primitives/data-table.tsx
```diff
@@ -115,7 +115,12 @@ export function DataTable<TData, TValue>({
               table.getRowModel().rows.map((row) => (
                 <TableRow key={row.id} data-state={row.getIsSelected() && 'selected'}>
                   {row.getVisibleCells().map((cell) => (
-                    <TableCell key={cell.id}>
+                    <TableCell
+                      key={cell.id}
+                      style={{
+                        width: cell.column.getSize(),
+                      }}
```
Mythie, line 122: I would have to assume that this is covered by specifying the size already, is that not the case?

plxity: I didn't get your question, can you elaborate?

Mythie: I assume that tanstack table is already setting the width of the column using `cell.column.getSize()` so doing so in the style attribute here should be redundant.

Instead we should only need the manually specified size change that you had in the other part of this change.

plxity: Okay, So are you suggesting to hardcode the width in this file?

plxity:
@Mythie Can you review this one?

Mythie (approved):

Mythie:
Cheers Apoorv! 😄

PR #925, opened 2024-02-12
fix: redirect URL preventing document flow

dguyen:
## Description

Currently the document redirect URL feature is preventing documents from being created unless a redirect URL is provided.

During the document edit flow, the redirect URL is hidden in an advanced tab with the value of an empty string, which will always fail the current Zod validation since `optional` requires undefined to pass.

There are multiple ways to fix this, but I think this is the easiest method where we can assume an empty string is valid.

## Testing Performed

Will require retesting the feature since I'm just assuming an empty string won't break anything downstream for the redirect URL feature.

Pinging @adithyaakrishna to retest the feature.

### DIFF
--- a/packages/trpc/server/document-router/schema.ts
+++ b/packages/trpc/server/document-router/schema.ts
@@ -1,6 +1,5 @@
 import { z } from 'zod';
 
-import { URL_REGEX } from '@documenso/lib/constants/url-regex';
 import { DocumentStatus, FieldType, RecipientRole } from '@documenso/prisma/client';
 
 export const ZGetDocumentByIdQuerySchema = z.object({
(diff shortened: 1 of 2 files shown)


packages/trpc/server/document-router/schema.ts
```diff
@@ -74,12 +73,7 @@ export const ZSendDocumentMutationSchema = z.object({
     message: z.string(),
     timezone: z.string(),
     dateFormat: z.string(),
-    redirectUrl: z
```
Mythie, line 77: zod url's have issues apparently, it's probably easier to just update the refine to allow an empty string.

dguyen: Any references?

Mythie: On the OG PR

dguyen: Got it 👍

Mythie (approved):

adithyaakrishna:
@dguyen Confirmed everything works 🚀

Issue #920, opened 2024-02-07
[Bug] Incorrect highlighting of items in search bar, theme change, and while adding the fields in a doc


sumitbishti:
### Issue Description

UPDATE: Bug is with the incorrect highlighting of the current email, and not the one mentioned below.
UPDATE 2: This bug is also present in the search bar, and while changing theme.

Current theme is dark but Light Mode is selected (by default).
Even if you select other theme, it doesn't get highlighted.
<img width="494" alt="Screenshot 2024-02-11 120007" src="https://github.com/documenso/documenso/assets/75713174/31bd8ff0-0413-4550-b7a4-709e199051c7">

Highlighted email is different than currently selected email.
<img width="604" alt="image" src="https://github.com/documenso/documenso/assets/75713174/f9d91f39-5c76-4f3f-a06f-b5fff4bf21e9">

Highlighted item is always the first one, irrespective of the selection.
<img width="547" alt="image" src="https://github.com/documenso/documenso/assets/75713174/f5f3460f-79f3-4dc0-ab0e-9bc08d9ff2ee">

### Please check the boxes that apply to this issue report.

- [X] I have searched the existing issues to make sure this is not a duplicate.
- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I have included any relevant screenshots.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [X] I want to work on creating a PR for this issue if approved

dguyen:
I can't replicate the issue.

Although it may be an UI issue where `sumit7edu` may look like it's selected, but it's actually just highlighted

The actual selected recipient is the `tmp7pmt` where you can see the check next to it

You can push a fix for this to highlight the correct item when the dropdown is open if you want

sumitbishti:
@dguyen I have updated the bug-reproduce steps. You can see for yourself now

PR #929, opened 2024-02-13
fix: fixed highlight issue in recipient selection, search bar, theme change

sumitbishti:
fixes: #920

### DIFF
--- a/apps/web/src/components/(dashboard)/common/command-menu.tsx
+++ b/apps/web/src/components/(dashboard)/common/command-menu.tsx
@@ -17,6 +17,7 @@ import {
 import type { Document, Recipient } from '@documenso/prisma/client';
 import { trpc as trpcReact } from '@documenso/trpc/react';
 import {
+  Command,
   CommandDialog,
   CommandEmpty,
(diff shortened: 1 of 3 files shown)


Mythie:
Could you please attach a set of images or video demo'ing the changes 😇

dguyen:
Hey @sumitbishti, we're fine with how the standard command menu highlights the first item. It lets users use arrow keys to select what they want.

The main issue here is that the for the recipient roles dropdown, the highlighted item should be the currently selected recipient, not the first item.

sumitbishti:
@dguyen okay i will just keep the changes related to recipient dropdown then. 
But what about the change theme dialog box? It wrongly shows the first (Light theme) item, when dark/system themes are currently applied.

dguyen:
> But what about the change theme dialog box? It wrongly shows the first (Light theme) item, when dark/system themes are currently applied.

It's not wrong, because the global search command feature isn't intended for that, it's meant to show what's available.

Whereas the recipient dropdown is specifically meant to show the current selected recipient.

PR #930, opened 2024-02-14
feat: Ability to send team invitation in bulk

### LINKED ISSUE #923: feat: Ability to bulk import (team members)
### Feature Description

It would be better if we provide the functionality to invite members. We can provide two options for this: 1) A TextField where users can input email addresses, with each email separated by commas. 2) The ability to upload a .csv file from the local device for bulk import.

Now, for inviting members to the team, users only need to input email addresses separated by commas into the text field.

### Use Case

_No response_

### Proposed Solution

_No response_

### Alternatives (optional)

_No response_

### Additional Context

_No response_

- [ ] I have searched the existing feature requests to make sure this is not a duplicate.
- [ ] I have provided a detailed description of the requested feature.
- [ ] I have explained the use case or scenario for this feature.
- [ ] I have included any relevant technical details or design suggestions.
- [ ] I understand that this is a suggestion and that there is no guarantee of implementation.

anikdhabal:
fixes #923 

https://github.com/documenso/documenso/assets/81948346/9f7cf419-91ec-4f43-b2c7-6fd3d0c13bfe


ElTimuro:
- @anikdhabal neat
- What happends with invalid email?
- A toast informing how many were successfully invite would be good, so the user does not have to count the resulting row in the memebers table

anikdhabal:
> * @anikdhabal neat
> * What happends with invalid email?
> * A toast informing how many were successfully invite would be good, so the user does not have to count the resulting row in the memebers table

Thanks for your review and suggestion.

Yeah, the code handles invalid email addresses. It simply shows the error message 'invalid email' in both cases.

apps/web/src/components/(teams)/dialogs/invite-team-member-dialog.tsx
```diff
@@ -70,41 +74,61 @@ export const InviteTeamMembersDialog = ({
...
-    });
+  const handleFileUpload = (e: React.ChangeEvent<HTMLInputElement>) => {
+    if (!e.target.files?.length) {
+      return;
+    }
+    const file = e.target.files[0];
+
+    if (file) {
+      const reader = new FileReader();
+      const emailRegex = /^([A-Z0-9_+-]+\.?)*[A-Z0-9_+-]@([A-Z0-9][A-Z0-9-]*\.)+[A-Z]{2,}$/i;
```
Mythie, line 100: Could we not use zod here?

apps/web/src/components/(teams)/dialogs/invite-team-member-dialog.tsx
```diff
@@ -70,41 +74,61 @@ export const InviteTeamMembersDialog = ({
...
+      reader.onload = (e) => {
```
Mythie, line 101: This is a bit scary, things like papaparse and such exist for handling the reading of CSV files in a safe manner since the format itself is prone to weirdness.

apps/web/src/components/(teams)/dialogs/invite-team-member-dialog.tsx
```diff
@@ -70,41 +74,61 @@ export const InviteTeamMembersDialog = ({
...
   const form = useForm<TInviteTeamMembersFormSchema>({
     resolver: zodResolver(ZInviteTeamMembersFormSchema),
     defaultValues: {
-      invitations: [
-        {
-          email: '',
-          role: TeamMemberRole.MEMBER,
-        },
-      ],
+      email: '' || [''],
```
Mythie, line 85: Seems it would be easier to always make this an array of emails so we don't have to deal with unions. I assume in the UI you would have an option to "Add another member" within the form.

apps/web/src/components/(teams)/dialogs/invite-team-member-dialog.tsx
anikdhabal: Yeah, we are also using Zod here. We can remove this.

apps/web/src/components/(teams)/dialogs/invite-team-member-dialog.tsx
anikdhabal: Thanks for the suggestion. Papaparse would be good for handling the csv file. Let me quickly add that.

Mythie: Consider the bundle size implications as well there may be smaller and just as good options 😄 Papaparse was the most popular last I messed with CSV's but I'm unsure as to what current recommendations are.

https://bundlephobia.com/ is a fantastic website for seeing how big something might be!

anikdhabal: Okay😄. Let me check this out.

dguyen:
Some thoughts just by looking at the loom (haven't looked at the code)

1. The default tab should be individual
2. Keep the old functionality prior to this PR where multiple users can be invited as the 'individual' tab
3. Instead of directly inviting bulk bulk team members, I think we should also allow the user to "Review" it (optional) by loading it into the individual section so they can review and change selected roles

anikdhabal:
> Some thoughts just by looking at the loom (haven't looked at the code)
> 
> 1. The default tab should be individual
> 2. Keep the old functionality prior to this PR where multiple users can be invited as the 'individual' tab
> 3. Instead of directly inviting bulk bulk team members, I think we should also allow the user to "Review" it (optional) by loading it into the individual section so they can review and change selected roles

Hello David, thank you for the suggestion. The idea of reviewing things sounds cool.

anikdhabal:
https://github.com/documenso/documenso/assets/81948346/3474c753-3c5a-442c-94c2-b092087df713

Thoughts on this? Instead of sending invites from the bulk import section, we use it for importing emails and then redirect to the invite member page, where we can change the selected role before sending the invite.

Or,

Users can send invitations from both sides, just like the current changes.

apps/web/src/components/(teams)/dialogs/invite-team-member-dialog.tsx
```diff
@@ -151,80 +175,123 @@ export const InviteTeamMembersDialog = ({
             An email containing an invitation will be sent to each member.
           </DialogDescription>
         </DialogHeader>
-
+        <div className="border-input ring-offset-background flex w-full justify-around rounded-md border">
```
harkiratsm, line 178: We could have used the Tabs component here. :thought_balloon:

Mythie:
> Documenso.and.8.more.pages.-.Personal.-.Microsoft_.Edge.2024-02-16.02-50-11.mp4 
> Thoughts on this? Instead of sending invites from the bulk import section, we use it for importing emails and then redirect to the invite member page, where we can change the selected role before sending the invite.
> 
> Or,
> 
> Users can send invitations from both sides, just like the current changes.

If we're doing this I'd avoid the text input and just have it only accept CSV where it'll be in `email,role` format so we can toss a bunch of ui elements.

We should also supply a template csv which would look like:

```
email,role
john@example.com,SIGNER
```

Seems crazy but in the past many clients have wanted the template to work off 😄

anikdhabal:
Screenshot 2024-04-04 010604

dguyen (approved):
Looks good to me, thanks!

PR #934, opened 2024-02-15
chore: minor lint fixes

dephraiim:


Mythie:
What does prettier think of this?

dephraiim:
> What does prettier think of this?

Can you clarify? `think` in what sense exactly?

Mythie:
I noticed that you used oxc for this, so if we merge is prettier going to tell me that it needs to change back to the prior form?

dephraiim:
Nope

Mythie (approved):

PR #935, opened 2024-02-15
feat: add roles to templates recipients

dephraiim:


packages/ui/primitives/recipient-role-icons.tsx
```diff
@@ -0,0 +1,10 @@
+import { BadgeCheck, Copy, Eye, PencilLine } from 'lucide-react';
+
+import type { RecipientRole } from '.prisma/client';
+
+export const ROLE_ICONS: Record<RecipientRole, JSX.Element> = {
+  SIGNER: <PencilLine className="h-4 w-4" />,
+  APPROVER: <BadgeCheck className="h-4 w-4" />,
+  CC: <Copy className="h-4 w-4" />,
+  VIEWER: <Eye className="h-4 w-4" />,
+};
```
dguyen, line 10: I think instead of extracting this, you can create a new input component for the role dropdown

Mythie: Lets deal with this later on 😄

Mythie (approved):

PR #945, opened 2024-02-17
feat: custom text field

dephraiim:
https://github.com/documenso/documenso/assets/55143799/c0f296ce-1e91-4b49-9c19-1bdffbd3163f


ElTimuro:
@dephraiim pretty awesome

- One line input would be more fitting since this doesn't support multiline I suspect? (would be something else as well)
- Dialog Button should read "Enter"

Enhancements (could be done now or later:
- It should be possible to label the field with a title, placeholder and description text when creating
- Field should be able to be marked as optional

Mythie (approved):

dephraiim:
🫡

PR #956, opened 2024-02-22
feat: custom banner

dephraiim:


Mythie:
Hello hello, I comandeered your PR to use a SiteSettings table instead. Not because the PR was wrong but I feel like we will want more of these toggleable items in the future where a feature flag does not make sense 😄

Mythie (approved):

apps/web/src/components/(dashboard)/layout/banner.tsx
```diff
@@ -0,0 +1,29 @@
...
+              <span dangerouslySetInnerHTML={{ __html: banner.data.content }} />
+            </div>
+          </div>
+        </div>
+      )}
+    </>
+  );
+};
+
+// Banner
```
dephraiim, line 27: 😬

PR #957, opened 2024-02-23
feat: signature certificate download

paolo:
This PR adds a new feature that appends a signature certificate to the pdf when user is downloading the pdf.
1. add new helper method to generate the cerficates(appendCertificates) 
2. connect appendCertificates method with onClickDownload button of two entry points
Screenshot 2024-03-01 at 12 05 37 PM


Mythie:
Nice start!

apps/web/package.json
```diff
@@ -22,6 +22,7 @@
     "@documenso/trpc": "*",
     "@documenso/ui": "*",
     "@hookform/resolvers": "^3.1.0",
+    "@react-pdf/renderer": "^3.3.8",
```
dephraiim, line 25: we have `react-pdf` already, any reason for this?

Mythie: This is a different react-pdf 😄 This one is for creating PDF's in the declarative fashion

paolo: precisely, this PR will be essentially creating PDFs documents.
this particular library is actually recommended by `react-pdf`: https://github.com/wojtekmaj/react-pdf?tab=readme-ov-file#lost

paolo:
Hi @Mythie , This PR should be ready for review

Mythie:
Looking at these now!

IgorHorta:
> Looking at these now!

Hey @Mythie, any updates here ?

PR #971, opened 2024-02-28
fix: return the recipient as an array to match other formats from zapier

catalinpit:
Return the recipient as an array to match the other formats for Zapier. Otherwise, Zaps with the "DOCUMENT_OPENED" hooks won't work.

All the other webhooks return the "Recipient" field as an array.


Mythie (approved):

Mythie:
Looks like ya build is failing though 😛

Mythie (approved):

PR #985, opened 2024-03-01
feat: save document logs as pdf

paolo:
In order to be CFR P11 compliant, document logs file should be created and stored within Documenso's servers.

This branch is using a similar strategy from the documents themselves to store "additional data", namely, logs and signature certificates (for the time being).

A new table has been created: `DocumentAdditionalData`, which is similar to `DocumentData`, thanks to this similarity, I was able to re-use some of the functions implemented to store and read files from the configured storage.

Regarding the PDF creation, `@react-pdf/renderer` was used to easily shape the document using their provided components. here's a sample of the resulting pdf document.

[demo-logs.pdf](https://github.com/documenso/documenso/files/14464750/demo-logs.pdf)

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -25,6 +25,7 @@ const config = {
     serverActions: {
       bodySizeLimit: '50mb',
     },
+    serverComponentsExternalPackages: ['@react-pdf/renderer'],
   },
   reactStrictMode: true,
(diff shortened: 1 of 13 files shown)


paolo:
Hi @Mythie , This PR should be ready for review

Mythie:
Looking at these now!

PR #987, opened 2024-03-04
feat: add offline development support

dguyen:
## Description

Add support to develop without network access since TRPC by default will prevent network requests when offline.

https://tanstack.com/query/v4/docs/framework/react/guides/network-mode#network-mode-always

## Changes Made

- Added `NEXT_PUBLIC_OFFLINE_DEVELOPMENT` environment variable to allow offline development
- Removed teams feature flag


Mythie (dismissed):

Mythie:
Approved, I wonder if it would be smarter to do something like the below rather than an env var

`const isDevelopingOffline = window.location.origin === 'localhost' && !window.navigator.onLine`

Mythie (approved):

PR #990, opened 2024-03-06
fix: remove duplicate neon pooler

dguyen:
## Description

Fixes the issue with Vercel preview deployments failing.

It appears that the old `PGHOST` environment variable injected by neon was:

`ep-snowy-snowflake-a2vc5pa2.eu-central-1.aws.neon.tech`

It is now:

`ep-snowy-snowflake-a2vc5pa2.eu-central-1-pooler.aws.neon.tech`

Notice the `-pooler` being attached automatically to the `PGHOST`.

## References

> The following changes were made to the [Neon Vercel Integration](https://vercel.com/integrations/neon):
>
>To ensure that users accessing a Neon database from a serverless environment have enough connections, the DATABASE_URL and PGHOST environment variables added to a Vercel project by the Neon integration are now set to a pooled Neon connection string by default. Pooled connections support up to 10,000 simultaneous connections. Previously, these variables were set to an unpooled connection string supporting fewer concurrent connections.

https://neon.tech/docs/changelog

https://neon.tech/docs/guides/vercel#manage-vercel-environment-variables


Mythie:
de nada

scripts/vercel.sh
```diff
@@ -77,10 +76,7 @@ function remap_database_integration() {
   if [[ "$NEXT_PRIVATE_DATABASE_URL" == *"neon.tech"* ]]; then
     log "Remapping for Neon integration"
 
-    PROJECT_ID="$(echo "$PGHOST" | cut -d'.' -f1)"
-    PGBOUNCER_HOST="$(echo "$PGHOST" | sed "s/${PROJECT_ID}/${PROJECT_ID}-pooler/")"
-
-    export NEXT_PRIVATE_DATABASE_URL="postgres://${PGUSER}:${PGPASSWORD}@${PGBOUNCER_HOST}/${PGDATABASE}?pgbouncer=true"
+    export NEXT_PRIVATE_DATABASE_URL="postgres://${PGUSER}:${PGPASSWORD}@${PGHOST}/${PGDATABASE}?pgbouncer=true"
```
Mythie, line 79: Now we need to account for the direct database url so prisma migrations can run 🥲

Mythie: Also we have a repeat of the logic in `packages/prisma/helper.ts`

dguyen: > Now we need to account for the direct database url so prisma migrations can run 🥲

Apparently if we upgrade prisma to 5.10 it lets us migrate using a connection pool?

https://neon.tech/docs/guides/prisma#connection-pooling-with-prisma-migrate

> Also we have a repeat of the logic in packages/prisma/helper.ts

The logic in the helper checks whether pooler is already attached, want me to try clean it up anyway?

```
  // Support for neon.tech (Neon Database)
  if (url.hostname.endsWith('neon.tech')) {
    const [projectId, ...rest] = url.hostname.split('.');

    if (!projectId.endsWith('-pooler')) {
      url.hostname = `${projectId}-pooler.${rest.join('.')}`;
    }

    url.searchParams.set('pgbouncer', 'true');

    process.env.NEXT_PRIVATE_DATABASE_URL = url.toString().replace('https://', 'postgres://');
  }
```

dguyen: Also what was the reason we can't use the direct database URL for neon for migrations?

dguyen:
@Mythie could you please have a look when possible

- Updated the Neon/Vercel integration to pass in `DATABASE_URL_UNPOOLED`
- Updated scripts to use this when required
- Catch all seed errors since they should not stop a migration/build from running (maybe we should actually fix all the invalid seeds when they run twice?)

Created example PR with a migration, since this variable is only created for new branches:

https://github.com/documenso/documenso/pull/992

Not sure if this will conflict with any thing on production since this should only be for previews

Mythie (approved):
lgtm, but also too big brain for me to fully understand anymore

PR #994, opened 2024-03-07
docs: add article on removing server actions

catalinpit:


Mythie (approved):

Mythie:
lgtm

michaelhays:
Hey @Mythie, really enjoyed this article. I have an application using tRPC for the API layer, and have been considering the switch to server actions for a while. But like many, I've had a handful of reservations, especially around stability/production-readiness.

Curious, are there any issues you're following with React or Next.js to follow along with the progress of server actions? And on a similar note, are you planning on sticking with tRPC for querying, or looking to migrate to RSC for data retrieval at some point?

Mythie:
Hey Michael! We're not really following any issues but instead just generally try gauge sentiment via Twitter and Newsletters.

We do use server components for quite a bit of data fetching, although realistically we will be rolling back a lot of this as well since client side requests should offer a snappier feel with less concern for waterfalls when using `router.refresh()`

michaelhays:
Hah same, wisdom of crowds right?

Noted, dang that's disappointing! Feels like we're at least a year off from this being truly ready for prime-time. Appreciate the insight!

PR #1006, opened 2024-03-08
feat: improve lint-staged performance 

G3root:
Right now, the eslint command runs separately for each staged file. This PR aims to change that by running just one eslint command for all the files that have been changed.


packages/eslint-config/index.cjs
```diff
@@ -4,29 +4,20 @@ module.exports = {
     'turbo',
     'eslint:recommended',
     'plugin:@typescript-eslint/recommended',
-    'plugin:prettier/recommended',
```
G3root, line 7: removed prettier eslint plugin 

https://www.joshuakgoldberg.com/blog/you-probably-dont-need-eslint-config-prettier-or-eslint-plugin-prettier/

Mythie: While I can agree with the sentiment of the article with having tools perform what they were designed for the editor integration leaves me worried.

For this change to work we'll also have to update our VSCode settings for the project and hope that others on non-VSCode editors have no difficulties with auto-formatting/linting.

I've also had issue with saveOnFormat + eslint causing an infinite loop of changes 😆 

That said it was some time ago and if we feel confident in this change we could do it!

lint-staged.config.cjs
```diff
@@ -1,7 +1,15 @@
+const path = require('path');
+
+const buildEslintCommand = (filenames) =>
```
Mythie, line 3: I'd probably just call these `eslint(...)` and `prettier(...)` or `runEslint`

Mythie (approved):
Approving and happy to merge for now, prettier changes will mostly be on a trial basis but lets see how we go!

PR #1007, opened 2024-03-08
feat: add website cta

catalinpit:
hover
notover

### DIFF
--- a/apps/marketing/src/app/(marketing)/blog/[post]/page.tsx
+++ b/apps/marketing/src/app/(marketing)/blog/[post]/page.tsx
@@ -7,7 +7,7 @@ import { ChevronLeft } from 'lucide-react';
 import type { MDXComponents } from 'mdx/types';
 import { useMDXComponent } from 'next-contentlayer/hooks';
 
-import CTA from '~/components/(marketing)/CTA';
+import { CallToAction } from '~/components/(marketing)/call-to-action';
 
 export const dynamic = 'force-dynamic';
 
@@ -93,7 +93,8 @@ export default function BlogPostPage({ params }: { params: { post: string } }) {
           Back to all posts
         </Link>
       </article>
-      {post.cta && <CTA />}
+
+      {post.cta && <CallToAction className="mt-8" utmSource={`blog__${params.post}`} />}
     </div>
   );
 }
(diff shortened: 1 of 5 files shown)


apps/marketing/src/components/(marketing)/CTA.tsx
```diff
@@ -0,0 +1,24 @@
...
+        Join the Open Document Signing Movement
+      </h2>
+
+      <p className="text-muted-foreground max-w-[55ch] text-center text-lg leading-normal">
+        Create your account and start using state-of-the-art document signing. Open and beautiful
+        signing is within your grasp.
+      </p>
+
+      <Button className="mt-4 rounded-full no-underline" size="sm" asChild>
+        <Link href="https://app.documenso.com/signup?utm_source=cta" target="_blank">
```
catalinpit, line 18: The `utm_source` should be customizable, I think

Mythie: should be yes, I'd also call this `call-to-action.tsx` with the function being `CallToAction` which is mostly just due to taste.

You can take an optional utmSource prop with a default of cta

`export default function CallToAction({ utmSource = 'cta }: CallToActionProps) { ... }`

apps/marketing/src/components/(marketing)/CTA.tsx
```diff
@@ -0,0 +1,24 @@
+import Link from 'next/link';
+
+import { Button } from '@documenso/ui/primitives/button';
+
+export default function CTA() {
+  return (
+    <div className="bg-background shadow-foreground/5 border-documenso mt-12 flex flex-col items-center justify-center rounded-lg border-4 border-double p-8 shadow-lg hover:shadow-md">
```
Mythie, line 7: also take classname as a prop, we never want to have `mt-12` or similar in the component itself since it's usually dependent on the layout/view/page its in

apps/marketing/src/app/(marketing)/blog/[post]/page.tsx
```diff
@@ -89,6 +91,8 @@ export default function BlogPostPage({ params }: { params: { post: string } }) {
         <ChevronLeft className="mr-2 h-6 w-6" />
         Back to all posts
       </Link>
+
+      {post.cta && <CTA />}
```
Comment (Mythie) [inline apps/marketing/src/app/(marketing)/blog/[post]/page.tsx:95]: <CallToAction className="mt-12" utmSource={`blog_${params.post}`} />

Mythie (approved):

Issue #1002, opened 2024-03-08
Speed up ESLint precommit checks


Mythie:
### Describe the improvement you are suggesting in detail

Currently it takes quite a while for ESLint to run during our precommit checks, we've attempted doing this with lint-staged with the following:

- The entire repository being linted
- Just the staged files being linted in one command e.g. `eslint --fix a.ts b.ts c.ts`
- Just the staged files being linted in multiple commands e.g. `eslint --fix a.ts && eslint --fix b.ts`

### Additional Information & Alternatives (optional)

This super long precommit time can scare off contributors and makes general development painful since you can be stuck on the linting stage for seconds up to minutes depending on the size of your commit (yes I know keep commits small, I'm sorry!).

Improving this will be a net benefit to DX for both contributors and the Documenso team.

### Do you want to work on this improvement?

No

### Please check the boxes that apply to this improvement suggestion.

- [X] I have searched the existing issues and improvement suggestions to avoid duplication.
- [X] I have provided a clear description of the improvement being suggested.
- [X] I have explained the rationale behind this improvement.
- [X] I have included any relevant technical details or design suggestions.
- [X] I understand that this is a suggestion and that there is no guarantee of implementation.

rohit-saluja:
can you assign me this issue.

Mythie:
Assigned, best of luck! We aren't expecting it to be incredibly fast but right now it's painstakingly slow.

I'd suggest creating like 30-50 dummy `.ts` files with content to test the slowdown during precommit 😄

rohit-saluja:
created a [pr](https://github.com/documenso/documenso/pull/1008) for this

PR #1008, opened 2024-03-08
fix: eslint config file parseOptions.project path is updated

rohit-saluja:
I found out that the problem of slow down is the use of parseOption.project with multiple tsconfig files which increases usage of memory consumption quite a lot.

Here is the reference I found to resolve this issue:

[reference link](https://github.com/typescript-eslint/typescript-eslint/issues/1192)
[Doc link](https://typescript-eslint.io/getting-started/typed-linting/monorepos/)

Based on this reference  I created the solution and tested out locally it does reduce the eslint time.

#1002

### DIFF
--- a/packages/eslint-config/index.cjs
+++ b/packages/eslint-config/index.cjs
@@ -20,7 +20,7 @@ module.exports = {
 
   parserOptions: {
     tsconfigRootDir: __dirname,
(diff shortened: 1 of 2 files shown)


Mythie (approved):

Mythie:
Awesome stuff, thanks Rohit! 🎉 

captain-america-salute

PR #1022, opened 2024-03-13
feat: added custom dark mode styling for swagger ui

adithyaakrishna:
**Description:**

Updated the OpenAPI doc dark mode styling here https://app.documenso.com/api/v1/openapi

**Before:**

Screenshot 2024-03-13 at 09 48 36

**After:**

Screenshot 2024-03-13 at 12 40 08


adithyaakrishna:
The light mode hasnt been changed and is kept the same :)

Screenshot 2024-03-13 at 13 29 20

catalinpit (commented):
Does it work for you locally?

I tried it on GitPod, but it still doesn't work for me. I see the old style.

CleanShot 2024-03-13 at 10 02 44

packages/api/v1/api-documentation.tsx
```diff
@@ -1,11 +1,30 @@
...
+import { useEffect } from 'react';
+
+import { useTheme } from 'next-themes';
 import SwaggerUI from 'swagger-ui-react';
 import 'swagger-ui-react/swagger-ui.css';
 
 import { OpenAPIV1 } from '@documenso/api/v1/openapi';
 
 export const OpenApiDocsPage = () => {
+  const { theme } = useTheme();
```
catalinpit, line 12: I think we should use `resolvedTheme` in this case.

> resolvedTheme: If enableSystem is true and the active theme is "system", this returns whether the system preference resolved to "dark" or "light". Otherwise, identical to theme
*[source](https://github.com/pacocoursey/next-themes#:~:text=resolvedTheme%3A%20If%20enableSystem%20is%20true%20and%20the%20active%20theme%20is%20%22system%22%2C%20this%20returns%20whether%20the%20system%20preference%20resolved%20to%20%22dark%22%20or%20%22light%22.%20Otherwise%2C%20identical%20to%20theme)*

Or am I wrong? What do you think?

catalinpit (changes requested):

adithyaakrishna:
@catalinpit Im not sure about why thats happening on gitpod, but locally it seems to work perfectly fine 😶 

[Swagger.webm](https://github.com/documenso/documenso/assets/23498248/ac58fb42-78b0-4a35-be79-e63df7b2c533)

packages/api/v1/api-documentation.tsx
adithyaakrishna: Makes sense, I have sent a commit below to update it :)

adithyaakrishna:
The 00c71fd commit fixes the focus ring issue as well :)

catalinpit (approved):

Mythie (approved):

Issue #1020, opened 2024-03-12
Admin Panel: Document super delete


Mythie:
Currently a document can't be deleted from Documenso permanently, this is done intentionally since even if an owner deletes a document the recipients still have a partial stake in it and thus should always be able to view it on Documenso.

That said bad actors may use the system and send documents that are spam or designed to trick users, self-hosters may also have also have sent a slew of testing documents which they now want to clean up as they move to production.

Instead of doing this manually in the database which is high risk and prone to error we should offer a way to delete the document within the current admin panel. This should require a confirmation before deletion can take place.

Requirements:
- Add a button to the `admin/documents/[id]` page to super delete the document
- Have the button trigger a dialog prompting for confirmation of deletion
- **Optional:** Email the document owner letting them know that their document was removed (maybe state a reason?)

rohit-saluja:
can I work on this

rohit-saluja:
created a [pr](https://github.com/documenso/documenso/pull/1023) for this.

PR #1023, opened 2024-03-13
feat: document super delete

rohit-saluja:
Added a dialog button at the bottom of the admin/documents/[id] page with confirmation popup.

Confirmation popup have validation for reason to input.

On confirmation document is deleted, and an email is triggred to the owner of document with the reason stated.

Let me know if there is any more requirement or correction is needed in this pr. :) #1020


Mythie:
Was this really two weeks ago :O 

I'm so sorry I'll get onto this next!

apps/web/src/app/(dashboard)/admin/documents/[id]/delete-document-dialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+  DialogTrigger,
+} from '@documenso/ui/primitives/dialog';
+import { Input } from '@documenso/ui/primitives/input';
+import { useToast } from '@documenso/ui/primitives/use-toast';
+
+export type DeleteDocumentDialogProps = {
+  document: Document;
+};
+
+export const DeleteDocumentDialog = ({ document }: DeleteDocumentDialogProps) => {
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/admin/documents/[id]/delete-document-dialog.tsx:28]: Lets call this `SuperDeleteDocumentDialog` and call the file `super-delete-document-dialog.tsx`, this will make it easier to find the file later on while working in the codebase 😄

apps/web/src/app/(dashboard)/admin/documents/[id]/delete-document-dialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const router = useRouter();
+  const { toast } = useToast();
+  const [reason, setReason] = useState('');
+  const { mutateAsync: deleteDocument, isLoading: isDeletingDocument } =
+    trpc.admin.deleteDocument.useMutation();
+
+  const handleDeleteDocument = async () => {
+    try {
+      await deleteDocument({ id: document.id, userId: document.userId, reason });
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/admin/documents/[id]/delete-document-dialog.tsx:37]: We should also check that reason is provided in this handler just incase!

apps/web/src/app/(dashboard)/admin/documents/[id]/page.tsx
```diff
@@ -81,6 +82,10 @@ export default async function AdminDocumentDetailsPage({ params }: AdminDocument
           ))}
         </Accordion>
       </div>
+
+      <hr className="my-4" />
+
+      {document && !document.deletedAt && <DeleteDocumentDialog document={document} />}
```
Comment (Mythie) [inline apps/web/src/app/(dashboard)/admin/documents/[id]/page.tsx:88]: We don't need to worry if the user has already deleted the document, we as admins would want the authoritative ability to delete any document. Otherwise a user may upload a malicious document and then delete it after hooking the target which stops us from removing it from the platform.

packages/email/template-components/template-document-delete.tsx
```diff
@@ -0,0 +1,33 @@
...
+  documentName,
+  assetBaseUrl,
+}: TemplateDocumentDeleteProps) => {
+  return (
+    <>
+      <TemplateDocumentImage className="mt-6" assetBaseUrl={assetBaseUrl} />
+
+      <Section>
+        <Text className="text-primary mx-auto mb-0 max-w-[80%] text-center text-lg font-semibold">
+          Your document has been deleted
```
Mythie, line 21: Your document has been deleted by the admins, the reason provided for the deletion is:

<reason>

This document can not be recovered, if you would like to dispute the reason for future documents please contact support.

packages/trpc/server/admin-router/router.ts
```diff
@@ -118,4 +121,20 @@ export const adminRouter = router({
...
     }
   }),
+  deleteDocument: adminProcedure
+    .input(ZAdminDeleteDocumentMutationSchema)
+    .mutation(async ({ input }) => {
+      const { id, userId, reason } = input;
+      try {
+        await deleteDocument({ id, userId });
+        await sendDeleteEmail({ documentId: id, reason });
+        return;
```
Mythie, line 131: We might make this a `{ success: true }` kind of return, while we can have tRPC return `undefined` as it would here we should provide something a little more enhanced.

packages/trpc/server/admin-router/router.ts
```diff
@@ -3,13 +3,16 @@ import { TRPCError } from '@trpc/server';
 import { findDocuments } from '@documenso/lib/server-only/admin/get-all-documents';
 import { updateRecipient } from '@documenso/lib/server-only/admin/update-recipient';
 import { updateUser } from '@documenso/lib/server-only/admin/update-user';
+import { deleteDocument } from '@documenso/lib/server-only/document/delete-document';
```
Mythie, line 6: We will want to either accept an optional arg for the super delete (permanent: boolean) or provide a new method to do so.

The intention with the super delete is that if we encounter spam on the platform we want to remove it for good as we also remove the user account. We would never want to soft delete someones documents as that's for them to manage.

rohit-saluja:
> Was this really two weeks ago :O
> 
> I'm so sorry I'll get onto this next!

No problem. will update this by the end of the day.

packages/trpc/server/admin-router/router.ts
rohit-saluja: A new function is been added for super-delete-document
It will always hard-delete the document.
For the case of pending status, it will send canceled emails to users.

rohit-saluja:
@Mythie  all the suggested changes are pushed to this pull request

Mythie (dismissed):

Mythie:
/tip 50

LGTM! 🕺🏻

cat-button-explosion

Mythie (approved):

PR #1033, opened 2024-03-17
fix: add missing theme switcher menu

anikdhabal:


Mythie:
G'day, the decision to remove the theme switcher from the menu and instead rely on the one in the command bar was intentional. 

We're mindful of having the avatar menu grow too large particularly as we've moved to including teams and such.

We are also aware that having it in the command bar makes it partially inaccessible however since we support system theme and have it set by default we consider it to be a minor issue for now.

Going to close this now, sorry 💚

anikdhabal:
> G'day, the decision to remove the theme switcher from the menu and instead rely on the one in the command bar was intentional.
> 
> We're mindful of having the avatar menu grow too large particularly as we've moved to including teams and such.
> 
> We are also aware that having it in the command bar makes it partially inaccessible however since we support system theme and have it set by default we consider it to be a minor issue for now.
> 
> Going to close this now, sorry 💚

Hey @Mythie thanks for the review. Could we not just create a different section in settings called 'Appearance' where users can change the theme, instead of adding that in the avatar menu? In the future, we can add more features to that section.

eg:-
Screenshot 2024-03-19 000514

PR #1035, opened 2024-03-18
fix: disable edit signer inputs

dguyen:
## Description

Update the add signer form to disable the signers when required.

I assume the actual issue is that `{...field}` was spreading a disabled prop which was overriding our one.

## Changes Made

- Use fieldset to disable inputs
- Manually disable select since fieldset doesn't work for that select for some reason

### DIFF
--- a/packages/ui/primitives/document-flow/add-signers.tsx
+++ b/packages/ui/primitives/document-flow/add-signers.tsx
@@ -151,9 +151,10 @@ export const AddSignersFormPartial = ({
 
           <AnimatePresence>
             {signers.map((signer, index) => (
-              <motion.div
+              <motion.fieldset
                 key={signer.id}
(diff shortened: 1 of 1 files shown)


Mythie:
Same situation for the add signers page:
image

All three signers are pending so they should be disabled :)

dguyen:
I don't understand what you're referring to

Are you saying the fix doesn't work?

Mythie:
The fix proposed works but we should also cover the `add-signers` step which also suffers from the same kind of regression likely due to the same reasons (`{...field}` spreading).

dguyen:
These changes are for the add signers?

Mythie:
Jesus 😢

Mythie (approved):

PR #1037, opened 2024-03-19
fix: swagger documentation authentication

catalinpit:
## Summary by CodeRabbit

- **Refactor**
	- Enhanced the API specification generation process to include operation IDs, security schemes, and security definitions more efficiently.


Mythie:
This seems to have mostly solved the issue but now we use the authorization value from the top of the swagger ui and ignore the ones passed in the "try it out" section despite it being marked as required.

Not sure how to solve but we should try just a little bit extra to see if we can make this just a tiny bit smoother.

catalinpit:
> This seems to have mostly solved the issue but now we use the authorization value from the top of the swagger ui and ignore the ones passed in the "try it out" section despite it being marked as required.
> 
> Not sure how to solve but we should try just a little bit extra to see if we can make this just a tiny bit smoother.

Yeah... I realised that, but I couldn't think of a better solution. I can remove the `authorization` param from the "Try it out" section, but I don't think that's a good solution. This way, it might be confusing to users. Similarly, I can make it non-required, but again, I don't think it's a good idea.

Will think of a better solution.

Mythie (approved):

PR #1039, opened 2024-03-20
fix: update document flow fetch logic

### LINKED ISSUE #470: BUG Signer fields cannot be edited/removed after modifying the signer.

dguyen:
## Description

**Fixes issues with mismatching state between document steps.**

For example, editing a recipient and proceeding to the next step may not display the updated recipient. And going back will display the old recipient instead of the updated values.

**This PR also improves mutation and query speeds by adding logic to bypass query invalidation.**

```ts
export const trpc = createTRPCReact<AppRouter>({
  unstable_overrides: {
    useMutation: {
      async onSuccess(opts) {
        await opts.originalFn();

        // This forces mutations to wait for all the queries on the page to reload, and in
        // this case one of the queries is `searchDocument` for the command overlay, which
        // on average takes ~500ms. This means that every single mutation must wait for this.
        await opts.queryClient.invalidateQueries(); 
      },
    },
  },
});
```

I've added workarounds to allow us to bypass things such as batching and invalidating queries. But I think we should instead remove this and update all the mutations where a query is required for a more optimised system.

## Example benchmarks

Using stg-app vs this preview there's an average 50% speed increase across mutations.

**Set signer step:**
Average old speed: ~1100ms
Average new speed: ~550ms

**Set recipient step:**
Average old speed: ~1200ms
Average new speed: ~600ms

**Set fields step:**
Average old speed: ~1200ms
Average new speed: ~600ms

## Related Issue

This will resolve #470

## Changes Made

- Added ability to skip batch queries
- Added a state to store the required document data.
- Refetch the data between steps if/when required
- Optimise mutations and queries

## Checklist

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx
+++ b/apps/web/src/app/(dashboard)/documents/[id]/edit-document.tsx
@@ -6,6 +6,10 @@ import { useRouter, useSearchParams } from 'next/navigation';
 
 import { useQueryClient } from '@tanstack/react-query';
 
+import {
+  DO_NOT_INVALIDATE_QUERY_ON_MUTATION,
+  SKIP_QUERY_BATCH_META,
+} from '@documenso/lib/constants/trpc';
 import { DocumentStatus } from '@documenso/prisma/client';
 import type { DocumentWithDetails } from '@documenso/prisma/types/document';
 import { trpc } from '@documenso/trpc/react';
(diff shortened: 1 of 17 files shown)


Mythie:
Dumb question and I get why you've done it, but instead of maintaining the `useStates` you could instead just change up the queryClient cache to update the `getWithDetails` query.

This would yield the same kind of result but without us managing it as directly.

dguyen:
It's done this way so we can use the results from the TPRC mutations to update the relevant data, which reduces loading times

Unless I'm misunderstanding what you're suggesting

The query is just to ensure we refresh every now and then.

Mythie:
Apologies, I seem to have a hard time expressing myself this week.

My point is rather than doing this state locally you can use [utils.setData](https://trpc.io/docs/client/react/useUtils#helpers) to update the queryClient's data for the `getDocumentWithDetails` query.

This would avoid the busy work of maintaining state ourself while also keeping everything updated without refetching.

dguyen:
Ah I see I'll have a look.

Pretty annoying it seems like we have to leave router refresh in since we need to clear the router cache for documents page 😩

Tempted to use trpc to render the documents page for client side at a later time

dguyen:
Applied the changes and the PR is now ready for review

Mythie:
I assume we will need to do the same for templates as well 😬

dguyen:
I retested your changes and they work, anything else to do for this?

Mythie:
> I retested your changes and they work, anything else to do for this?

Lets do templates in a separate PR

Mythie (approved):

PR #1043, opened 2024-03-21
chore: minor update to open page

dephraiim:


ElTimuro:
- Added some minor text changes
- Otherwise good to go

PR #1044, opened 2024-03-21
chore: add status widget

dephraiim:


ElTimuro:
image

There should be a pulsing, green dot
image

apps/marketing/src/app/(marketing)/layout.tsx
```diff
@@ -1,78 +1,14 @@
...
-                  Claim Now
-                </a>
-              </div>
-            </div>
-          </div>
-        )}
-
-        <Header className="mx-auto h-16 max-w-screen-xl px-4 md:h-20 lg:px-8" />
-      </div>
+    <div className="relative flex min-h-[100vh] max-w-[100vw] flex-col overflow-y-auto overflow-x-hidden pt-20 md:pt-28">
```
Mythie, line 10: The conditionals in the old code are load bearing, I appreciate the refactoring but it's probably easier to leave it as it was and keep the PR scoped to just the status widget 😄

dephraiim: The initial marketing layout was a Client Component at the root, which made it impossible to access the status information from the Server Component status widget so I had to refactor the layout and I moved all the client-side components within the layout, ensuring that the root of the layout remains a Server Component.

ElTimuro:
@dephraiim set this status: 
image

but not showing up
image

is this on our side or openstatus?

dephraiim:
@ElTimuro, I think it is from OpenStatus

https://api.openstatus.dev/public/status/documenso returns `operational` and that is where I'm getting the value

apps/marketing/src/components/(marketing)/status-widget.tsx
mxkaske, line 1: Hey guys! So if you wanna use the StatusWidget and keep SSG (and avoid React Server Components) you can make it a Client Component instead and wrap it around `Suspense` to return a fallback until the status is loaded client side.
```js
export function StatusWidgetContainer() {
  const res = use(getStatus('documenso'))
  return (
    <Suspense fallback={<StatusWidgetFallback />}>
      <StatusWidge status={res.status} />
    </Suspense>
  )
}

// fallback example pls make it yours
function StatusWidgetFallback() {
  return (
    <div className="border-border flex max-w-fit items-center gap-2 rounded-md border px-3 py-1 text-sm">
      <span className="bg-muted h-5 w-20 animate-pulse rounded-md" />
      <span className="bg-muted relative inline-flex h-2 w-2 rounded-full" />
    </div>
  );
}
```

Props where props are due: [tweet](https://twitter.com/gregberge_/status/1761979422360649897).

It's not ideal, and we gonna provide a client side solution. But until now, that should do the work pretty much. It depends on what strat you wanna go.

mxkaske: Because it's not the correct slug of the page. Here you go: `documenso-status`.

dephraiim: makes sense, thank you @mxkaske

dephraiim:
Trying to revert to the initial layout has become a mess. Closing this in favour of this #1068

PR #1051, opened 2024-03-22
fix: mask recipient token

catalinpit:
The searchDocuments function is used for the shortcuts commands, afaik. The function returns the documents that match the user query (if any), alongside all their recipients.

The reason for that is so it can build the path for the document. E.g. if you're the document owner, the document path will be `..../documents/{id}`. But if you're a signer for example, the document path (link) will be `..../sign/{token}`.

So instead of doing that on the frontend, I moved it to the backend.

At least that's what I understood. If I'm wrong, please correct me.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `CommandMenu` component to simplify search result generation and improve document link management based on user roles.
- **Refactor**
	- Updated document search logic to include recipient token masking and refined document mapping.
- **Style**
	- Minor formatting improvement in document routing code.


packages/trpc/server/document-router/router.ts
```diff
@@ -281,7 +282,12 @@ export const documentRouter = router({
           query,
           userId: ctx.user.id,
         });
-        return documents;
+
+        const maskedDocuments = documents.map((document) =>
+          maskRecipientTokensForDocument({ document, user: ctx.user }),
+        );
```
dguyen, line 288: What's this for? We're already masking the token inside the function

The owner of the document gets to see all the tokens

dguyen (approved):

dguyen (approved):

ElTimuro:
@dguyen could you TLDR this change for my?

dguyen:
> @dguyen could you TLDR this change for me?

As part of the command search, documents are searched which returns recipients tokens if you are the owner of the document

This PR just masks it when it is not required to be visible

Ultimately it doesn't matter since the owner can always access the recipient tokens on the `/documents` page

ElTimuro:
> > @dguyen could you TLDR this change for me?
> 
> As part of the command search, documents are searched which returns recipients tokens if you are the owner of the document
> 
> This PR just masks it when it is not required to be visible
> 
> Ultimately it doesn't matter since the owner can always access the recipient tokens on the `/documents` page

got it, thanks.

PR #1054, opened 2024-03-23
fix: render fields on document load

dguyen:
## Description

Currently if you try to load the document edit page when fields need to be rendered, you will not be able to see the fields until you proceed to the next step.

This is because the fields require the document PDF to be loaded prior to rendering them.

This PR resolves that issue by only rendering the fields after the PDF is loaded.

## Changes Made

- Add a state to track whether the PDF is loaded
- Render the fields only after the PDF is loaded

## Testing Performed

Tested document flow manually and the fields are rendered correctly on load.

## Checklist

- [X] I have updated the documentation to reflect these changes, if applicable.


Mythie:
I thought we handled this with an element visible component which renders the children once it can find a PDF page? 🤔

dguyen:
It's something to do with the window.document not being available immediately. So an error is thrown in the createPortal when trying to access document.body & also in the coordinate calculation.

Initially I fixed it by waiting for the window to load, but then the fields were rendered with the incorrect height and there were some weird bugs, so I just defaulted to wait for the document PDF to be loaded instead.

We can try to resolve properly if you want, I just think this is a fast solution

Mythie:
Fast is fine for now

Mythie (dismissed):

PR #1067, opened 2024-03-28
fix: passkey login

dguyen:
## Description

Fixed issue where passkeys do not work on https deployments.

### DIFF
--- a/packages/trpc/server/auth-router/router.ts
+++ b/packages/trpc/server/auth-router/router.ts
@@ -132,7 +132,10 @@ export const authRouter = router({
   }),
 
   createPasskeySigninOptions: procedure.mutation(async ({ ctx }) => {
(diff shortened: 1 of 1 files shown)


packages/trpc/server/auth-router/router.ts
```diff
@@ -132,7 +132,10 @@ export const authRouter = router({
   }),
 
   createPasskeySigninOptions: procedure.mutation(async ({ ctx }) => {
-    const sessionIdToken = parse(ctx.req.headers.cookie ?? '')['next-auth.csrf-token'];
+    const cookies = parse(ctx.req.headers.cookie ?? '');
+
+    const sessionIdToken =
+      cookies['__Host-next-auth.csrf-token'] || cookies['next-auth.csrf-token'];
```
Mythie, line 138: Checks out, noting that if we move upgrade to Auth.js this will chance once again 💀

Mythie (approved):

PR #1078, opened 2024-04-04
feat: download completed docs via api

catalinpit:
## Description

Allow users to download a completed document via API.

## Testing Performed

Tested the code locally by trying to download both draft and completed docs. Works as expected.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Implemented functionality to download signed documents directly from the app.


packages/api/v1/implementation.ts
```diff
@@ -83,6 +86,63 @@ export const ApiContractV1Implementation = createNextRoute(ApiContractV1, {
...
+      const document = await getDocumentById({
+        id: Number(documentId),
+        userId: user.id,
+        teamId: team?.id,
+      });
+
+      if (
+        !document ||
+        !document.documentDataId ||
+        DocumentDataType.S3_PATH !== document.documentData.type
```
Mythie, line 111: This is a 400 error instead since the document exists but we don't support downloading them via the api at the moment.
 We should extract this specific condition into a different return

packages/api/v1/implementation.ts
```diff
@@ -83,6 +86,63 @@ export const ApiContractV1Implementation = createNextRoute(ApiContractV1, {
...
+          status: 404,
+          body: {
+            message: 'Document not found',
+          },
+        };
+      }
+
+      if (document.status !== DocumentStatus.COMPLETED) {
+        return {
+          status: 404,
```
Mythie, line 123: 400 error

packages/api/v1/implementation.ts
```diff
@@ -83,6 +86,68 @@ export const ApiContractV1Implementation = createNextRoute(ApiContractV1, {
...
+
+      const { url } = await getPresignGetUrl(document.documentData.data);
+
+      return {
+        status: 200,
+        body: { downloadUrl: url },
+      };
+    } catch (err) {
+      return {
+        status: 404,
```
dguyen, line 143: I think this should be a 500 error?

catalinpit: done, thx.

PR #1092, opened 2024-04-10
feat: add certificate and audit log pdfs

Mythie:


apps/web/src/app/(internal)/%5F%5Fhtmltopdf/certificate/page.tsx
```diff
@@ -0,0 +1,313 @@
...
+    if (
+      documentAuthOptions.globalActionAuth === 'PASSKEY' ||
+      recipientAuthOptions.actionAuth === 'PASSKEY'
+    ) {
+      authLevel = 'Passkey Re-Authentication';
+    }
+
+    if (recipientAuthOptions.actionAuth === 'EXPLICIT_NONE') {
+      authLevel = 'Email';
+    }
```
dguyen, line 131: This seems like this might be prone to bugs where different globalAccessAuth vs accessAuth would display the wrong auth used

Example a recipient has ACCOUNT auth set, but the global is PASSKEY

Seems like it would choose PASSKEY instead of the actual method used, which is ACCOUNT

Can you use the `extractDocumentAuthMethods` util function which handles extracting the auth expected?

apps/web/src/app/(internal)/%5F%5Fhtmltopdf/certificate/page.tsx
```diff
@@ -0,0 +1,313 @@
...
+                        </p>
+                      </div>
+                    </TableCell>
+
+                    {/* <TableCell truncate={false} className="w-[min-content] align-top">
+                      <p className="text-sm text-muted-foreground print:text-xs">
+                        Authentication: {'<authentication>'}
+                      </p>
+                      <p className="text-sm text-muted-foreground print:text-xs">IP: {'<ip>'}</p>
+                    </TableCell> */}
```
dguyen, line 293: Old stuff?

packages/ui/primitives/table.tsx
```diff
@@ -2,13 +2,16 @@ import * as React from 'react';
...
-    </div>
-  ),
-);
+const Table = React.forwardRef<
+  HTMLTableElement,
+  React.HTMLAttributes<HTMLTableElement> & {
+    overflowHidden?: boolean;
+  }
+>(({ className, overflowHidden, ...props }, ref) => (
+  <div className={cn('w-full', overflowHidden ? 'overflow-hidden' : 'overflow-auto')}>
```
dguyen, line 11: Would it be better to pass parentClassName or something instead so we can use it in the future?

Sames with TableCell

apps/web/src/app/(internal)/%5F%5Fhtmltopdf/certificate/page.tsx
dguyen, line 293: Old stuff?

dguyen (changes requested):
Also build is failing, and can't run as dev 

Module not found: Can't resolve 'electron'

apps/web/src/app/(dashboard)/documents/[id]/logs/download-certificate-button.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const DownloadCertificateButton = ({
+  className,
+  documentId,
+}: DownloadCertificateButtonProps) => {
+  const { mutateAsync: downloadCertificate, isLoading } =
+    trpc.document.downloadCertificate.useMutation();
+
+  const onDownloadCertificatesClick = async () => {
+    const { url } = await downloadCertificate({ documentId });
```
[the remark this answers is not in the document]
Comment (Mythie) [reply apps/web/src/app/(dashboard)/documents/[id]/logs/download-certificate-button.tsx:22]: :(

apps/web/src/app/(internal)/%5F%5Fhtmltopdf/audit-log/data-table.tsx
```diff
@@ -0,0 +1,89 @@
+'use client';
```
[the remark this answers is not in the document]
Mythie: It's an internal route, trying to make it obvious that we shouldn't be navigating to it in normal situations

apps/web/src/app/(internal)/%5F%5Fhtmltopdf/certificate/page.tsx
Mythie: I'll change that, did a minor look for some code that would do that but didn't find it initially so I did this monstrosity

apps/web/src/app/(internal)/%5F%5Fhtmltopdf/certificate/page.tsx
Mythie: ya

packages/ui/primitives/table.tsx
Mythie: Honestly the defaults are good, this is more specialised since I really want to avoid a responsive table

apps/web/src/app/(internal)/%5F%5Fhtmltopdf/certificate/page.tsx
dguyen: Fair, just the path name looks a little cooked %5F%5F

apps/web/src/app/(internal)/%5F%5Fhtmltopdf/certificate/page.tsx
```diff
@@ -93,40 +90,30 @@ export default async function SigningCertificate({ searchParams }: SigningCertif
...
-      recipientAuthOptions.actionAuth === 'PASSKEY'
-    ) {
+    if (extractedAuthMethods.derivedRecipientActionAuth === 'PASSKEY') {
       authLevel = 'Passkey Re-Authentication';
     }
 
-    if (recipientAuthOptions.actionAuth === 'EXPLICIT_NONE') {
+    if (extractedAuthMethods.derivedRecipientActionAuth === 'EXPLICIT_NONE') {
       authLevel = 'Email';
     }
```
dguyen, line 118: Is it possible to throw a tsmatch in or something so when we add new action auth methods we can cause build errors?

dguyen (approved):

apps/web/src/app/(internal)/%5F%5Fhtmltopdf/certificate/page.tsx
Mythie: Yeah that should work, for some reason I thought it wouldn't but it should do

Mythie: Although it wouldn't be exhaustive since we consider access and action auth 🤔

dguyen: You could create a function which has tspattern inside that takes in the actionAuth and returns a string or null right

If it's null you use the default authLevel email/account in this outer function

Mythie: Aight

dguyen (approved):
fanks

PR #1099, opened 2024-04-11
feat: add document log page link

dguyen:
## Description

Adds a link from the document page view to the document page log view

image


ElTimuro:
Planning to PR "Audit Log" as label, but that's not a critical change :D

PR #1106, opened 2024-04-16
chore: added filename extension check

adithyaakrishna:
**Description:**

This PR adds a check for filename title and if the title ends with `.pdf` then the extension isnt added or else its added

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced email attachment handling to ensure PDF files are correctly identified with a ".pdf" extension.

### DIFF
--- a/packages/lib/server-only/document/send-completed-email.ts
+++ b/packages/lib/server-only/document/send-completed-email.ts
@@ -130,7 +130,7 @@ export const sendCompletedEmail = async ({ documentId, requestMetadata }: SendDo
         text: render(template, { plainText: true }),
         attachments: [
           {
-            filename: document.title,
+            filename: document.title.endsWith('.pdf') ? document.title : document.title + '.pdf',
             content: Buffer.from(completedDocument),
           },
         ],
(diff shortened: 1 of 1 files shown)


ElTimuro:
@adithyaakrishna   any idea why this happens in the first place? This fix is good for now, but i would like to know why

adithyaakrishna:
@ElTimuro This usually happens if the user uploads a PDF file `For example, Adi.pdf` which they renamed it to `Adi` without the `.pdf` extension at last. Our app shows the preview but the sendEmail functionality which has the filename didnt have a check for the extension. So added the same in this PR

Please Lmk I wasnt clear or if you have any other doubts and I'll add a short video for reproducing this issue :)

ElTimuro:
> @ElTimuro This usually happens if the user uploads a PDF file `For example, Adi.pdf` which they renamed it to `Adi` without the `.pdf` extension at last. Our app shows the preview but the sendEmail functionality which has the filename didnt have a check for the extension and if a file sent without the extension in its filename, Gmail cant determine the filetype so it showed the blue file icon on the users received email. To fix that, I added the filename extension check in this PR
> 
> Please Lmk I wasnt clear or if you have any other doubts and I'll add a short video for reproducing this issue :)

- no pretty clear
- Though weird that people would do that (user do crazy stuff but stil)
- I always assumed it was sth. related to "." in the filename breaking the parsing or sth.
- Anyhow, will test and merge this

catalinpit (approved):

PR #1109, opened 2024-04-18
feat: add visible completed fields

dguyen:
## Description

Added the ability for recipients to see fields from other recipients who have completed the document when they are signing the document

Added the ability for the document owner to see fields from recipients who have completed the field on the document page view (only visible when the document is pending)

## 🚨🚨 Migrations🚨🚨

- Drop all `Fields` that do not have a `Recipient` set (not sure how it was possible in the first place)
- Remove optional `Recipient` field on `Field` which doesn't make sense 

## Summary by CodeRabbit

- **New Features**
	- Enhanced document viewing by adding read-only fields based on document status.
	- Improved signing page by fetching and displaying completed fields for tokens.
	- Updated avatar component to show recipient status with tooltips for better user interaction.

- **Bug Fixes**
	- Made `recipientId` a required field in the database to ensure data consistency.

- **Refactor**
	- Optimized popover functionality in UI components for better performance and user experience.

- **Documentation**
	- Added detailed component and function descriptions for new features in the system.


ElTimuro:
@dguyen should we keep the recipientless fields a bit (or at least download them) if we need the data later for debugging?

dguyen:
> @dguyen should we keep the recipientless fields a bit (or at least download them) if we need the data later for debugging?

We can't run the migration without them gone

I don't think the data is really useful, but I'll try to get a copy anyway

Mythie (approved):
im trustin ya on this one but we should revisit it later

PR #1111, opened 2024-04-19
fix: prevent signing draft documents

dguyen:
## Description

Currently users can sign and complete draft documents, which will result in a completed document in an invalid state.

## Changes Made

- Prevent recipients from inserting or uninserting fields for draft documents
- Prevent recipients from completing draft documents 
- Remove ability to copy signing tokens unless document is pending

## Summary by CodeRabbit

- **New Features**
	- Enhanced document status visibility and control across various components in the application. Users can now see and interact with document statuses more dynamically in views like `DocumentPageView`, `DocumentEditPageView`, and `DocumentsDataTable`.
	- Improved document signing process with updated status checks, ensuring actions like signing, completing, and removing fields are only available under appropriate document statuses.

- **Bug Fixes**
	- Adjusted document status validation logic in server-side operations to prevent actions on incorrectly stated documents, enhancing the overall security and functionality of document processing.

### DIFF
--- a/packages/lib/server-only/document/complete-document-with-token.ts
+++ b/packages/lib/server-only/document/complete-document-with-token.ts
@@ -49,8 +49,8 @@ export const completeDocumentWithToken = async ({
 
   const document = await getDocument({ token, documentId });
 
-  if (document.status === DocumentStatus.COMPLETED) {
-    throw new Error(`Document ${document.id} has already been completed`);
+  if (document.status !== DocumentStatus.PENDING) {
+    throw new Error(`Document ${document.id} must be pending`);
   }
 
   if (document.Recipient.length === 0) {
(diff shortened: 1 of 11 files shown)


packages/lib/server-only/document/complete-document-with-token.ts
```diff
@@ -49,8 +49,8 @@ export const completeDocumentWithToken = async ({
 
   const document = await getDocument({ token, documentId });
 
-  if (document.status === DocumentStatus.COMPLETED) {
```
catalinpit, line 52: Isn't this check needed anywhere anymore?

catalinpit (commented):
This "Sign" option should probably be disabled for draft docs, like "void", "download" and the others are.

CleanShot 2024-04-19 at 11 15 58

packages/lib/server-only/document/complete-document-with-token.ts
dguyen: What do you mean? The check is still there indirectly via `status !== PENDING`

dguyen:
> This "Sign" option should probably be disabled for draft docs, like "void", "download" and the others are.

Nice catch, thanks!

packages/lib/server-only/document/complete-document-with-token.ts
catalinpit: Got confused with the statuses. You right.

catalinpit (approved):
Nice work!

PR #1112, opened 2024-04-22
fix: refactor use template

dguyen:
## Description

Refactor the "use template" flow

These changes include backwards compatible logic for `/api/v1` which makes it a bit messy.

## Changes Made

- Add placeholders for recipients
- Add audit log when document is created
- Trigger DOCUMENT_CREATED webhook when document is created
- Remove role field when using template
- Remove flaky logic when associating template recipients with form recipients
- Refactor to use `Form` 

### Using template when document has no recipients

image

### Using template with recipients 

image

### Using template with the send option selected

image


ElTimuro:
@dguyen is this direct template link prep or just cleanup?

dguyen:
> @dguyen is this direct template link prep or just cleanup?

It's just a little bit of both

Was doing mocking up stuff to get a general outline/idea of how direct template links would fit in and found some general issues which require fixing

PR #1115, opened 2024-04-24
fix: hide account action reauth

dguyen:
## Description

Hide the account reauth option

image

image


ElTimuro (commented):
- Should we do commented out code?
- I would propose never commenting out but always deleting

Mythie (approved):

PR #1125, opened 2024-04-29
chore: send document onUpdate

adithyaakrishna:
**Description:**

This PR updates the condition to send email on subject or message update


packages/lib/server-only/document/send-document.tsx
```diff
@@ -213,31 +212,6 @@ export const sendDocument = async ({
     }),
   );
 
-  const allRecipientsHaveNoActionToTake = document.Recipient.every(
```
ElTimuro, line 216: @adithyaakrishna not sure I get it, won't removing this not break the "every signer is a BCC" case?

adithyaakrishna: Closing this PR and opening another one, this has some changes not done by me :/

PR #1126, opened 2024-04-29
chore: update logic for resending emails

adithyaakrishna:
**Description:**

This PR updates the email sending logic to send emails on subject and message being updated and if the recipient signed status is not signed, then email is sent and if signed, email isnt sent


Mythie:
Closing, this seems like it would unintentionally spam recipients if a given owner made a few updates within a quick timespan. 

Resending should be explicit via the resend dialog or implicit on a fixed schedule (which is yet to come).

PR #1127, opened 2024-04-29
feat: remove the existing empty signer if its the only one

adithyaakrishna:
**Description:**

This PR removes the existing empty signer if its the only one


packages/ui/primitives/document-flow/add-signers.tsx
```diff
@@ -174,6 +164,20 @@ export const AddSignersFormPartial = ({
     removeSigner(index);
   };
 
+  const onAddSelfSigner = () => {
```
dguyen, line 167: Did we specifically move this function for a reason? Makes the diff unnecessarily larger

Also add some spacing between these lines in the function for readability

packages/ui/primitives/document-flow/add-signers.tsx
```diff
@@ -174,6 +164,20 @@ export const AddSignersFormPartial = ({
...
+    const lastSignerIndex = signers.length - 1;
+    if(!signers[lastSignerIndex].name || !signers[lastSignerIndex].email){
```
dguyen, line 169: Your spacing is messed up, I think it might be your extensions? No space after `if` and no space before `{`

This if statement is incorrect, if the last signer has an email but no name it gets replaced?

dguyen (changes requested):
I'd expect this to fill the first empty field found

https://github.com/documenso/documenso/assets/20962767/2352bc97-675a-485a-a021-e443fe0b3f81

Also some bugs


https://github.com/documenso/documenso/assets/20962767/81a84d30-998e-4608-81fc-9678e1cb1654

packages/ui/primitives/document-flow/add-signers.tsx
adithyaakrishna: Yes, it was on purpose as I wanted to onRemoveSigner() after its declaration

adithyaakrishna:
@dguyen Regarding the bug, we only save the information of the signer if we click `Continue` or `Add Signer`. If the user realizes he wants to add himself after typing the email, he can just click the `Add Myself` button and his details will get added. 

But if its needed to have a condition to check if the fields are filled and if so, add a new entry, I would be glad to change it to that one 👀

dguyen:
> @dguyen Regarding the bug, we only save the information of the signer if we click `Continue` or `Add Signer`. If the user realizes he wants to add himself after typing the email, he can just click the `Add Myself` button and his details will get added.

I'm not sure what you mean, adding yourself obviously should not override other recipients.

packages/ui/primitives/document-flow/add-signers.tsx
dguyen: This isn't resolved, did you read the whole message?

Look below, if a previous SAVED recipient has no name, adding yourself removes them. 

https://github.com/documenso/documenso/assets/20962767/731c3e70-28f3-421d-bc75-680ff47502f7

adithyaakrishna:
@dguyen Gotcha, I will update the conditions to reflect the same

Mythie (approved):

PR #1130, opened 2024-04-30
fix: disable encrypted pdfs

dguyen:
## Description

Currently if you complete a pending encrypted document, it will prevent the document from being sealed due to the systems inability to decrypt it.

This PR disables uploading any documents that cannot be loaded as a temporary measure.

**Note**
This is a client side only check

## Changes Made

- Disable uploading documents that cannot be parsed
- Refactor putFile to putDocumentFile
- Add a flag as a backup incase something goes wrong

### DIFF
--- a/packages/lib/universal/upload/put-file.ts
+++ b/packages/lib/universal/upload/put-file.ts
@@ -19,7 +19,7 @@ type File = {
  * Uploads a document file to the appropriate storage location and creates
  * a document data record.
  */
-export const putDocumentFile = async (file: File) => {
+export const putPdfFile = async (file: File) => {
   const isEncryptedDocumentsAllowed = await getFlag('app_allow_encrypted_documents');
 
   // This will prevent uploading encrypted PDFs or anything that can't be opened.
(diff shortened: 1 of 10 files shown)


packages/lib/universal/upload/put-file.ts
```diff
@@ -12,14 +15,36 @@ type File = {
...
+    await PDFDocument.load(await file.arrayBuffer()).catch((e) => {
+      console.error(`PDF upload parse error: ${e.message}`);
+
+      throw new AppError('INVALID_DOCUMENT_FILE');
+    });
+  }
+
+  const { type, data } = await putFileInDatabase(file);
+
+  return await createDocumentData({ type, data });
```
Mythie, line 36: I'd instead have this `return await putFile(...)` and remove the other changes

I'd probably also rename this to `putPdfFile`

Mythie (changes requested):

packages/lib/universal/upload/put-file.ts
dguyen: Opps yeah `putFileInDatabase` was supposed to be `putFile`

I'm not sure what you mean by remove the rest though?

PR #1131, opened 2024-04-30
fix: enforce users to have stripe account

dguyen:
## Description

Currently users who sign in via Google SSO do not get assigned a Stripe customer account.

This enforces the Stripe customer requirement on sign in.

There might be a better place to put this so it's open to any suggestions.


Mythie (approved):

Mythie:
lgtm

PR #1139, opened 2024-05-02
feat: force signature fields for document signers

catalinpit:
## Description

Show a dialog when the document has signers with no signature fields placed.

## Changes Made

Created a new dialog that'll be triggered when the document owner tries to send a document to the signers without placing signature fields. The document owners can't proceed to the next step unless they add signature fields.

## Checklist

https://github.com/documenso/documenso/assets/25515812/f1b5c34e-2ce0-40e3-804c-f05d23045710

## Summary by CodeRabbit

- **New Features**
  - Introduced "Direct Links" for async signing, allowing users to create documents from templates using public links.
  - Added `MissingSignatureFieldDialog` component to ensure users don't miss adding signature fields.

- **Enhancements**
  - Updated blog content to provide guidance on contract management and announce new pricing plans.

- **Bug Fixes**
  - Improved async signing process for better efficiency and control.

- **Refactor**
  - Improved internal code structure and import order for stripe-related functionality.

- **Tests**
  - Enhanced e2e tests to verify signature presence before document creation and updated test flows for document approval.


packages/ui/primitives/document-flow/missing-signature-field-dialog.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <DialogTitle>No signature field found</DialogTitle>
+          <DialogDescription>
+            <p className="mt-2">
+              Some signers have not been assigned a signature field. Please assign a signature field
+              to each signer before proceeding.
+            </p>
+          </DialogDescription>
+        </DialogHeader>
+        <DialogFooter>
+          <div className="flex w-full flex-nowrap gap-4">
```
dguyen, line 35: I think you should remove this `div` so we can use the normal layout for footers to stay consistent

catalinpit: Ok, thanks. I did it, although the buttons on the webhook modal is also like this (buttons on the left side). I guess I should update that as well?

dguyen: Yeah just use the defaults which is right side

dguyen:
Can you add some server side checks as well?

packages/trpc/server/field-router/router.ts
```diff
@@ -22,6 +23,27 @@ export const fieldRouter = router({
...
+            fields.some(
+              (field) => field.type === 'SIGNATURE' && field.signerEmail === recipient.email,
+            ),
+        );
+
+        if (!everySignerHasSignature) {
+          throw new TRPCError({
+            code: 'BAD_REQUEST',
+            message:
+              'Some signers have not been assigned a signature field. Please assign a signature field to each signer before proceeding.',
```
ElTimuro: "At least 1 signature field"

packages/ui/primitives/document-flow/missing-signature-field-dialog.tsx
```diff
@@ -0,0 +1,45 @@
...
+              to each signer before proceeding.
+            </p>
+          </DialogDescription>
+        </DialogHeader>
+        <DialogFooter>
+          <Button type="button" variant="secondary" onClick={onOpenChange}>
+            Cancel
+          </Button>
+          <Button type="submit" onClick={onOpenChange}>
+            Proceed
```
ElTimuro, line 39: Do we still need both options if you can't proceed? Why the 2 buttons?

catalinpit: True that. I kept it like this for consistency. Should I just go ahead and remove the proceed button?

ElTimuro: - Since this is not just a hint why you can't contine it can be a alert actually

packages/lib/server-only/document/send-document.tsx
```diff
@@ -120,6 +121,29 @@ export const sendDocument = async ({
...
+    (recipient) =>
+      recipient.role !== RecipientRole.SIGNER ||
+      fieldsWithSignerEmail.some(
+        (field) => field.type === 'SIGNATURE' && field.signerEmail === recipient.email,
+      ),
+  );
+
+  if (!everySignerHasSignature) {
+    throw new Error('Some signers have not been assigned a signature field.');
+  }
```
Mythie, line 145: This will mean API users will also have to follow these conventions, we should consider if that's something we want @ElTimuro

packages/ui/primitives/document-flow/add-fields.tsx
```diff
@@ -596,9 +615,16 @@ export const AddFieldsFormPartial = ({
             previousStep();
             remove();
           }}
-          onGoNextClick={() => void onFormSubmit()}
+          onGoNextClick={handleGoNextClick}
         />
       </DocumentFlowFormContainerFooter>
+
+      {!everySignerHasSignature && (
```
Mythie, line 622: Any reason to conditionally render here? I get why it's there but it seems easier to just let it always render since we toggle visibility anyways.

dguyen (dismissed):

dguyen (approved):

PR #1144, opened 2024-05-03
fix: profile claim name length

ElTimuro:
fixes the caim name length on the profile claim popup

### DIFF
--- a/packages/trpc/server/profile-router/router.ts
+++ b/packages/trpc/server/profile-router/router.ts
@@ -90,7 +90,7 @@ export const profileRouter = router({
       try {
         const { url } = input;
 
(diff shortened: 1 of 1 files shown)


packages/trpc/server/profile-router/router.ts
```diff
@@ -90,7 +90,7 @@ export const profileRouter = router({
       try {
         const { url } = input;
 
-        if (IS_BILLING_ENABLED() && url.length <= 6) {
+        if (IS_BILLING_ENABLED() && url.length < 6) {
```
adithyaakrishna, line 93: LGTM

adithyaakrishna (approved):

PR #1161, opened 2024-05-15
chore: change default sender name to match prod

ElTimuro:
change the default sender to sth. nicer


ElTimuro:
- Changed the ENV Variable on App Staging and Prod
- Reployed Staging
- Prod will update on next deploy

dguyen (approved):

PR #1165, opened 2024-05-18
feat: add direct templates links

dguyen:
## Description

Direct templates links is a feature that provides template owners the ability to allow users to create documents based of their templates.

## General outline

This works by allowing the template owner to configure a "direct recipient" in the template.

When a user opens the direct link to the template, it will create a flow where they sign the fields configured by the template owner for the direct recipient. After these fields are signed the following will occur:

- A document will be created where the owner is the template owner
- The direct recipient fields will be signed
- The document will be sent to any other recipients configured in the template
- If there are none the document will be immediately completed

## Notes

There's a custom prisma migration to migrate all documents to have 'DOCUMENT' as the source, then sets the column to required.


packages/prisma/migrations/20240522085944_add_template_direct_links/migration.sql
```diff
@@ -0,0 +1,41 @@
...
+CREATE TYPE "DocumentSource" AS ENUM ('DOCUMENT', 'TEMPLATE', 'TEMPLATE_DIRECT_LINK');
+
+-- AlterTable
+ALTER TABLE "Document" ADD COLUMN     "source" "DocumentSource";
+
+-- Custom: UpdateTable
+UPDATE "Document" SET "source" = 'DOCUMENT' WHERE "source" IS NULL;
+
+-- Custom: AlterColumn
+ALTER TABLE "Document" ALTER COLUMN     "source" SET NOT NULL;
```
dguyen, line 17: This was migration was altered to make `document.source` not null, but at the same time set all current sources to `document`

Mythie (dismissed):
Approving for now but we really will need to come back to this mess (in a good way) 😿

PR #1195, opened 2024-06-24
fix: limits syncing issue

### LINKED ISSUE #1192: Update limit/doc-table when deleting documents or when drafting documents.
### Describe the improvement you are suggesting in detail

Currently when a user creates a document and does not complete it (draft), the limit does not update until a full page refresh. Similarly, when a user deletes a document. 

Screenshot 2024-06-21 at 1 05 55 PM

### Useful information
```
const refreshLimits = async () => {
    const newLimits = await getLimits({ teamId });

    setLimits((oldLimits) => {
      if (equals(oldLimits, newLimits)) {
        return oldLimits;
      }

      return newLimits;
    });
  };
```
This method refreshes the limit to see if it was updated. 
It's current use case is only in 2 ```useEffect()``` to be called when the Limits provider component mounts.

This inaccuracy happens in 2 scenarios.
1. Uploading document which directs you to the 'edit' page.
      - When a user  goes back to the 'documents' page, since the provider does not need to re-mount the limits value is not updated.

2. When deleting a document, we call the callback method ```onSuccess```, which refreshes the page. This refresh only serves purpose for server pages and refetches the new data for the table. It does not un-mount provider, which makes the limits value stale.

### Proposal

We can create a new ```triggerRefetchLimits()``` function from the Limits context and add it to the consumable values. Or we can use ```refetchLimits()```. 

This can be used when deleted a document inside ```onSuccess``` or it can be used when uploading a document before redirection to the editing page.

This is my first contribution, so I would love some feedback!

### Do you want to work on this improvement?

Yes

### Please check the boxes that apply to this improvement suggestion.

- [X] I have searched the existing issues and improvement suggestions to avoid duplication.
- [X] I have provided a clear description of the improvement being suggested.
- [x] I have explained the rationale behind this improvement.

emmpaz:
---
name: Pull Request
---

## Description

Expose ```refreshLimit()``` to be able to keep the limit in sync when deleting/creating a document.

## Related Issue

Fixes #1192 

## Changes Made

Adding the ```refreshLimits()``` as a context value. We can only keep it as a optional parameter, because requring it means to put it in the initial props which means it will become part of the limits state, where a function cannot be within a state.

Inside ```useLimits()``` I create a type and runtime safe function that makes sure ```refreshLimits()``` is a function within the context. 

- Change 1
  Inside ```delete-document-dialog.tsx```, I simply add ```void refreshLimits()``` after we refresh the page.

- Change 2
   Inside ```upload-document.tsx```, we can do the same as above. After successfully uploading a document, we can refresh the limit.

## Testing Performed

I have uploading a document and keeping it as a draft by redirecting back the '/documents' page and can confirm the limit updates.

I have deleted a document from my list and can confirm the limit updates accordingly.

## Checklist

## Additional Notes

As I am exposing this function to the context and new components consume this context, I had to make sure that there was not any un-necessary re-renders.

As ```router.refresh()``` only refreshes the server data, the client component is cached so there is no re-render for the client that consumes this data. 

## Summary by CodeRabbit

- **New Features**
  - Added functionality to refresh usage limits after deleting a document.
  - Added functionality to refresh usage limits after uploading a document.

- **Improvements**
  - Enhanced limits management by incorporating a `refreshLimits` function in context, ensuring up-to-date limit information.


ephraimduncan (approved):

Mythie (approved):

Mythie:
Thanks for this 😸 
thanks-pusheen

PR #1206, opened 2024-06-29
fix: use native URL parser instead of wrong regex

aeris:
Current way to validate URL from regex is bugged
- no support for auth part `URL_REGEX.test("http://auth@example.org")`
- no support for explicit port `URL_REGEX.test("http://example.org:5000")`
- no support for fragment `URL_REGEX.test("http://example.org#foo")`
- no support for query `URL_REGEX.test("http://example.org?foo=bar")`
- valid domain rejection `URL_REGEX.test("http://localhost")`

This patch use the native URL node parsing to test correctly for URL validity

## Summary by CodeRabbit

- **Refactor**
  - Improved URL validation by changing the method from a regular expression pattern to a function that tests the validity of a URL. This should enhance accuracy in URL validation.


Mythie:
Thanks for this one! I think we can rename the constant to isValidURL or isValidRedirectURL (since this is only used for that) as per the rabbit suggestion and just turn it into a function updating the current call sites.

We will also want to make sure that we're only allow http/https protocol links since using the URL constructor we could potentially have other protocols added that would be a hassle to deal with later on.

Think you could be up for doing that? If not I'm happy to make the required changes to get this across the line 😸

Mythie:
Took a while but we got there 🙌🏻

Mythie (approved):

PR #1238, opened 2024-07-23
fix: dateformat api bug

catalinpit:
The `dateFormat` value wasn't processed correctly for documents created via the API.

Now, it checks that the `dateFormat` value is valid and throws an error if it is not.

## Summary by CodeRabbit

- **New Features**
  - Enhanced date format and time zone validation for API requests to improve error handling.
  - Updated handling of date format and time zone settings in the UI for dynamic assignment based on predefined options.
  - Integrated OpenAPI documentation features into the API schema for better clarity on expected input formats.

- **Bug Fixes**
  - Fixed an issue where invalid time zones could be accepted in settings.


Mythie:
If I recall there is also a way to add extra information to the generated OpenAPI spec using TS-REST which is where we could also list the formats that we accept.

catalinpit:
> If I recall there is also a way to add extra information to the generated OpenAPI spec using TS-REST which is where we could also list the formats that we accept.

Great catch, thanks!

CleanShot 2024-07-29 at 12 05 27@2x

Mythie (approved):

PR #1313, opened 2024-08-28
feat: add TypedSQL to Documenso

jharrell:
---
name: TypedSQL in Documenso
about: Add new TypedSQL feature from Prisma ORM 5.19.0 to Documenso
---

## Description

Hello again! Now that we're released, I'd like to see if we can get TypedSQL into Documenso proper. I realize this PR is a bit more dramatic than what you're probably looking for, I'm happy to scale back changes but just wanted to get this ported over sooner so more eyes can get a look.

## Changes Made

- moved calls from `prisma generate` to `prisma generate --sql`
- transitioned raw queries and query builder usages to TypedSQL
- ...

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced new SQL generation commands to enhance database management.
	- Added SQL queries for monthly document completion, user sign-up analysis, and user growth tracking.
- **Improvements**
	- Streamlined the build process with updated scripts and dependency management.
- **Localization**
	- Updated German localization strings to improve user experience for German-speaking users.
	- Modified marketing localization for better clarity in German.


Mythie:
Yo had a new look at this one, the hard dependency on the database being available feels like it'll be tough when packaging into a container.

Feels like it could be a fun opportunity to support pglite or something since I understand you need to do some introspection on the actual database to understand the type for the SQL. Happy to chat over a call or something as well 😄

Mythie:
> Yo had a new look at this one, the hard dependency on the database being available feels like it'll be tough when packaging into a container.
> 
> Feels like it could be a fun opportunity to support pglite or something since I understand you need to do some introspection on the actual database to understand the type for the SQL. Happy to chat over a call or something as well 😄

Following on, just realised this is due to using SQLx on the Rust side of things which is a fantastic choice! Let me see what we can do to facilitate this and make the building aspect of things not dreadful!

jharrell:
> Following on, just realised this is due to using SQLx on the Rust side of things which is a fantastic choice! Let me see what we can do to facilitate this and make the building aspect of things not dreadful!

Standing on the shoulders of giants 😆
We're really happy that SQLx helped us here. As for PGLite, there's a community [driver adapter](https://github.com/lucasthevenet/pglite-utils/tree/main/packages/prisma-adapter) but unfortunately I don't think that TypedSQL works with driver adapters. So we'd have to come up with a different path.

Issue #1178, opened 2024-06-05
GET Document Fields


julian-tourii:
### Feature Description

Please add an endpoint to list the fields on a document

Swagger: https://app.documenso.com/api/v1/openapi

### Use Case

Updating a document requires updating the fields associated with the document. To update the fields of a document or to delete fields from a document, the existing fields must be queried.

### Proposed Solution

Add a swagger endpoint to GET the document fields `/api/v1/documents/{id}/fields`

### Please check the boxes that apply to this feature request.

- [X] I have searched the existing feature requests to make sure this is not a duplicate.
- [X] I have provided a detailed description of the requested feature.
- [X] I have explained the use case or scenario for this feature.
- [X] I have included any relevant technical details or design suggestions.
- [X] I understand that this is a suggestion and that there is no guarantee of implementation.
- [x] I want to work on creating a PR for this issue if approved

dguyen:
Hello! Thanks for bringing this up, I've assigned this to you

AnshulKahar2729:
Hey @julian-tourii, are you working on it, if not I would like to contribute in it,

PR #1317, opened 2024-09-02
feat: return fields in GET /documents/:id endpoint

Etrenak:
## Description

To be able to use the PATCH `/api/v1/documents/{id}/fields/{fieldId}` endpoint, we need to know the fields ID of a particular document. The issue #1178 suggested to create a new endpoint for this. To be consistent with the `/templates` endpoint, I propose in this PR to directly add the `fields` field to the `/documents/:id` endpoint.

## Related Issue

This is related to #1178 but it might still be interesting to create a new endpoint which could also return the `fieldMeta`.

## Changes Made

Added `fields` to `ZSuccessfulGetDocumentResponseSchema` in the API contract, and changed the implementation accordingly.

## Testing Performed

- Tested on a document with and without fields.
- Full production build (`npm run build`)

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced API response to include additional field data associated with documents.
	- Introduced a new schema for field objects, improving data validation and structure in API responses.

- **Chores**
	- Restructured schema definitions for better organization and readability.


Mythie:

https://documen.so/cla

Mythie:
This one looks good, I'll just need to grab a moment to run through the testing on this one manually!

Etrenak:
For my use case, I would like to be able to also get fieldMeta returned by this endpoint. However, when I tried to also return the fieldMeta, I encountered some typing issues.

That is, I simply added fieldMeta in the schema like this :
```typescript
export const ZSuccessfulGetDocumentResponseSchema = ZSuccessfulDocumentResponseSchema.extend({
  recipients: z.lazy(() => z.array(ZSuccessfulRecipientResponseSchema)),
  fields: ZFieldSchema.pick({
    id: true,
    recipientId: true,
    type: true,
    page: true,
    positionX: true,
    positionY: true,
    width: true,
    height: true,
  }).extend({
    fieldMeta: ZFieldMetaSchema,
  }).array(),
});
```

But since fieldMeta in only known as a JsonValue in the prisma schema, we cannot simply return the result of the query to the database. Please tell me if this is OK to simply do the parsing of the database result in `implementation.ts:getDocument` into a correct field meta type ? I think it would be better to edit the database schema to indicate that the type is one of the advanced fields meta (Either TEXT meta, or DROPDOWN meta, ...), but it is certainly not the easiest way.

Do you know why it was chosen to store the field meta as a JsonValue in the database, and which of the two solutions I should prefer ?

Etrenak:
So I ended up adding it anyway. I hope I did not waste your time on previously testing the PR without fieldMeta.

I also put back the definition of `ZFieldSchema` in its original position, since I learned what `z.lazy()` does (I'm not really familiar with TS). Using `z.lazy()` seems to mess up the OpenAPI schema generation though :frowning_face:. It only displays `fields: "string"` instead of the correct schema, but that's not a big issue.

Feel free to tell me if you would like another implementation, my precedent comment still stands.

catalinpit (approved):

Etrenak:
Hey, I see this has been approved. Any idea when it might get merged ? Thanks!

Mythie (approved):

PR #1378, opened 2024-10-02
feat: undo changes in signature pad with ctrl + z

aditya-chakraborty:
## Description

Allow undoing changes in Signature Pad with Ctrl + Z or Cmd + Z

https://github.com/user-attachments/assets/ebeb6537-66db-4228-800c-9c90e3c77526

## Summary by CodeRabbit

- **New Features**
	- Enhanced translations for German, English, and French, improving the accuracy and clarity of user interface text related to document management and authentication.
	- Added keyboard shortcuts for the `SignaturePad` component, allowing users to undo actions using "Ctrl+Z" or "Cmd+Z".

- **Bug Fixes**
	- Corrected untranslated strings and improved pluralization rules across all language translations.

- **Refactor**
	- Optimized the `SignaturePad` component for better performance and responsiveness.

### DIFF
--- a/packages/ui/primitives/signature-pad/signature-pad.tsx
+++ b/packages/ui/primitives/signature-pad/signature-pad.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import type { HTMLAttributes, MouseEvent, PointerEvent, TouchEvent } from 'react';
-import { useEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import { Trans } from '@lingui/macro';
 import { Undo2 } from 'lucide-react';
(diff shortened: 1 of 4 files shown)


packages/ui/primitives/signature-pad/signature-pad.tsx
```diff
@@ -210,7 +210,33 @@ export const SignaturePad = ({
 
       onChange?.($el.current.toDataURL());
     }
-  };
+  }, [lines, defaultValue, onChange, perfectFreehandOptions]);
+
+  const handleKeyDown = useCallback(
```
Mythie, line 215: We should move the keyDown method into the useEffect callback so we don't need to use `useCallback` doing this is going to cause a lot of registrations and de-registrations as the arguments changes.

PR #1436, opened 2024-11-01
chore: enable typed signature by default

catalinpit:
Enable typed signature by default and also add the option to set a typed signature in the profile page.

## Release Notes

- **New Features**
  - Introduced the ability to input typed signatures alongside drawn signatures in the Profile Form.
  - Added a global setting for enabling or disabling typed signatures for teams.
  - Enhanced error handling for typed signatures based on team settings.
  - Improved form functionality to manage typed signatures in document preferences.
  - Added new mutation for updating typed signature settings in templates.
  - Enhanced template editing process to manage typed signature settings alongside other fields.
  - Updated the handling of signatures in various components to ensure correct processing based on input format.

- **Bug Fixes**
  - Improved rendering logic for displaying signatures, ensuring typed signatures are displayed correctly.

- **Documentation**
  - Updated localization files to include new translations related to typed signatures and document actions.

- **Chores**
  - Streamlined the Update Team Form by removing unnecessary properties, simplifying user interactions.


catalinpit:
> When i tried to sign using 'hello world' the outcome when the document is sealed has weird spacing
> 
> image

CleanShot 2024-11-12 at 13 02 12@2x

Yeah, it breaks with `hello world`, but not with others, hmmm. I tested it `my test` and my full name and it didn't break. Weird.

apps/web/src/components/forms/profile.tsx
```diff
@@ -138,6 +138,7 @@ export const ProfileForm = ({ className, user }: ProfileFormProps) => {
                     containerClassName={cn('rounded-lg border bg-background')}
                     defaultValue={user.signature ?? undefined}
                     onChange={(v) => onChange(v ?? '')}
+                    allowTypedSignature={true}
```
catalinpit, line 141: I hardcoded this since the typed signature should always be available on the profile page.

apps/web/src/app/(signing)/sign/[token]/signature-field.tsx
```diff
@@ -176,6 +192,41 @@ export const SignatureField = ({
     }
   };
 
+  useEffect(() => {
```
Comment (catalinpit) [inline apps/web/src/app/(signing)/sign/[token]/signature-field.tsx:195]: Had a bit of difficulty trying to fit the text inside the Signature field. 😬

apps/web/src/app/(signing)/sign/[token]/provider.tsx
```diff
@@ -44,6 +44,12 @@ export const SigningProvider = ({
   const [email, setEmail] = useState(initialEmail || '');
   const [signature, setSignature] = useState(initialSignature || null);
```
Comment (catalinpit) [inline apps/web/src/app/(signing)/sign/[token]/provider.tsx:45]: This wouldn't initialize `signature` to `initialSignature` all the time, so I had to use the `useEffect` below.

Comment (Mythie) [reply apps/web/src/app/(signing)/sign/[token]/provider.tsx:45]: This is because the parent component is no longer key'ing it which used to be the case. So the useState will retain the value until it's unmounted even if the prop changes

apps/web/src/app/(signing)/sign/[token]/signature-field.tsx
Comment (Mythie) [reply apps/web/src/app/(signing)/sign/[token]/signature-field.tsx:195]: You might want a layout effect here to block rendering until you get a real number, see: https://sentry.engineering/blog/perfectly-fitting-text-to-container-in-react

apps/web/src/components/forms/profile.tsx
Mythie: 👍🏻

packages/lib/server-only/pdf/insert-field-in-pdf.ts
```diff
@@ -92,45 +95,88 @@ export const insertFieldInPDF = async (pdf: PDFDocument, field: FieldWithSignatu
...
+          }
+
+          page.drawImage(image, {
+            x: imageX,
+            y: imageY,
+            width: imageWidth,
+            height: imageHeight,
+            rotate: degrees(pageRotationInDegrees),
+          });
+        } else {
```
Mythie, line 137: this old chestnut @dguyen

Mythie (changes requested):
Can I hassle you to also add this as a team setting/preference so certain teams with a requirement to only have drawn signatures don't need to remember to do this for every new document/api request.

catalinpit:
> Can I hassle you to also add this as a team setting/preference so certain teams with a requirement to only have drawn signatures don't need to remember to do this for every new document/api request.

Sure thing!

Mythie:
@coderabbitai resolve

Mythie (approved):

Issue #1452, opened 2024-11-06
Search link to document not working when in team


jsbrain:
### Issue Description

When searching for a existing document in the search bar and trying to access the result via 'enter', the link will reference the personal team when the document is actually in a team. So the link will try to access the document in `/documents/ID` instead of `/t/my-team/documents/ID`.

### Version [e.g., 2.0.1]

1.7.1

### Please check the boxes that apply to this issue report.

- [x] I have searched the existing issues to make sure this is not a duplicate.
- [x] I have provided steps to reproduce the issue.
- [ ] I have included relevant environment information.
- [ ] I have included any relevant screenshots.
- [x] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [ ] I want to work on creating a PR for this issue if approved

catalinpit:
Hey @jsbrain, thanks for reporting this!

Do you mean the search bar at the top of the page? The one marked with number `1` in this picture?

![CleanShot 2024-11-07 at 12 00 38@2x](https://github.com/user-attachments/assets/dbd6ffcd-13db-459c-b713-da9bf42385c5)

We'll fix this asap!

catalinpit:
> @catalinpit this is the one, yes!

Opened a PR for it, thanks!

PR #1453, opened 2024-11-07
fix: fix the document url in the command menu search

catalinpit:
fixes https://github.com/documenso/documenso/issues/1452

## Summary by CodeRabbit

- **New Features**
	- Enhanced document retrieval to filter based on user membership roles within teams, improving access control.
	- Added capability to retrieve document paths associated with teams.

- **Bug Fixes**
	- Improved logic for determining document visibility based on user roles and document settings.


apps/web/src/components/(dashboard)/common/command-menu.tsx
```diff
@@ -106,7 +106,7 @@ export function CommandMenu({ open, onOpenChange }: CommandMenuProps) {
 
     return searchDocumentsData.map((document) => ({
       label: document.title,
-      path: document.path,
+      path: document.teamId ? `/t/${document?.team?.url}/documents/${document.id}` : document.path,
```
dguyen, line 109: I feel like this would cause some issues with this statement which produces the above `document.path`

```
path: isOwner(document, user) ? `/documents/${document.id}` : getSigningLink(Recipient, user)
```

Can you move this logic into the `search-documents-with-keyword` instead? And maybe use the `${formatDocumentsPath(team?.url)}/${document.id}` logic instead

dguyen (approved):
Looks good, thanks

PR #1477, opened 2024-11-15
feat: download doc without signing certificate

catalinpit:
## Description

I added the option of downloading a document without the signing certificate for teams. They can disable/enable the option in the preferences tab.

The signing certificate can still be downloaded separately from the `logs` page.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new option to include a signing certificate in document preferences, enhancing user control over document settings.
	- Added a toggle for the signing certificate in the document preferences form.

- **Bug Fixes**
	- Improved logic for document sealing to conditionally include signing certificates based on team settings.

- **Translations**
	- Updated translations for German, Spanish, and French to include new entries and improve existing phrases related to document actions and settings.
	- Added new translation for "Include the Signing Certificate in the Document" across multiple languages.


packages/lib/jobs/definitions/internal/seal-document.ts
```diff
@@ -127,7 +136,11 @@ export const SEAL_DOCUMENT_JOB_DEFINITION = {
       flattenForm(pdfDoc);
       flattenAnnotations(pdfDoc);
 
-      if (certificateData) {
+      if (
```
Mythie, line 139: This probably could have been a ternary on `certificateData` e.g.

```typescript
const certificateData =
      document.team?.teamGlobalSettings?.includeSignerCertificate === false
        ? null
        : await getCertificatePdf({ documentId }).catch(() => null);
```

Mythie: Then you wouldn't need the below which seems like it would break on individual documents anyways 🤔

PR #1500, opened 2024-12-02
fix: refactor teams router

dguyen:
## Description

With the upcoming plans for SDK support, we need to ensure we have a proper OpenAPI spec, regardless if we decide on using ts-rest or trpc for generating an open api spec.

These changes are the initial prototype to bring in API response schemas.

## Changes Made

- Installed latest Zod in root package and removed zod from nested packages 
- Updated teams router with output zod schema parsers
- Added a Prisma generator to produce Zod schemas to allow us to easily create response schemas. 
- Added WIP metadata for trpc openapi as a demostration.
- Set `strictNullChecks` to true

> - Added a Prisma generator to produce Zod schemas to allow us to easily create response schemas. 

I looked into packages which converts types to zod schemas to try streaming this easier, but I'm not sure it will work well


TSP-Dev:
Issue reported at https://github.com/documenso/documenso/issues/1544

PR #1515, opened 2024-12-06
fix: prevent accidental signatures

ephraimduncan:
CleanShot 2024-12-06 at 03 30 39

CleanShot 2024-12-06 at 03 32 34


catalinpit (commented):
Wouldn't it be useful to show an error when the drawn signature is invalid? At the moment, if you have an invalid signature and click on the "Signature" field, it opens the signature modal. That may confuse signers.

I added a simple example below.

CleanShot 2024-12-06 at 10 17 27@2x

At the moment, it doesn't seem to recognise the default signatures set in the profile page. 

I set the typed signature `test sig` on my profile page, and when I go to sign a document, it opens the signature modal when I click on the "Signature" field.

CleanShot 2024-12-06 at 10 22 28@2x

packages/lib/translations/de/common.po
```diff
@@ -472,11 +472,19 @@ msgstr "Genehmigung"
 msgid "Before you get started, please confirm your email address by clicking the button below:"
 msgstr "Bitte bestätige vor dem Start deine E-Mail-Adresse, indem du auf den Button unten klickst:"
 
+<<<<<<< HEAD
```
Mythie, line 475: An attempt was made :)

PR #1516, opened 2024-12-06
fix: avoid having a drawn and typed signature at the same time

ephraimduncan:


catalinpit (commented):
Where does this happen? I couldn't reproduce it on the profile page or document signing page view.

https://github.com/user-attachments/assets/38501be3-44ba-4116-b72d-30ab5e8336c6

doug-andrade:
> Where does this happen? I couldn't reproduce it on the profile page or document signing page view.
> 
>  sigsig.mp4

If you have a Signature saved, but then you choose to type for a specific document. Next time you go sign a document it will default back to your saved signature from your profile, not the typed one.

doug-andrade:
https://www.loom.com/share/6e5f08edd90d4758864e91fbb165d84b?sid=3d0f3a6b-48ec-4459-9b5a-6d65c963ddb0

catalinpit:
> > Where does this happen? I couldn't reproduce it on the profile page or document signing page view.
> > sigsig.mp4
> 
> If you have a Signature saved, but then you choose to type for a specific document. Next time you go sign a document it will default back to your saved signature from your profile, not the typed one.

This makes sense. That's how the default saved signature should work. You can change the default signature on the profile page.

Also, the PR aims to solve something else - to avoid combining a drawn and typed signature.

ephraimduncan:
> Also, the PR aims to solve something else - to avoid combining a drawn and typed signature.

Exactly, this is a reproduction. There are instances it doesn't happen but more often than not, it happens


https://github.com/user-attachments/assets/8baed388-b2c7-4d0b-8e56-f478c9744d67

dguyen (approved):

PR #1527, opened 2024-12-11
chore: update deps

Mythie:
Slightly improves dev server performance


Mythie:
Failures are due to caching of deps

dguyen (changes requested):
no more next

dguyen (approved):

PR #1535, opened 2024-12-12
feat: add trpc openapi

dguyen:
## Description

- Add TRPC Open API demo
- Add output schemas for public endpoints Documents/Templates/Fields/Recipients


Mythie:
Interesting

PR #1537, opened 2024-12-13
fix: checkbox logic

catalinpit:
## Changes Made

- Change 1
- Change 2
- ...

## Testing Performed

- ...

## Checklist


apps/web/src/app/(signing)/sign/[token]/checkbox-field.tsx
```diff
@@ -35,6 +35,17 @@ export type CheckboxFieldProps = {
   onUnsignField?: (value: TRemovedSignedFieldWithTokenMutationSchema) => Promise<void> | void;
 };
 
+const parseCheckedValues = (customText: string): string[] => {
```
Comment (Mythie) [inline apps/web/src/app/(signing)/sign/[token]/checkbox-field.tsx:38]: You handled this a bit weirdly here and better in the `insert-field` file, instead you should extract this to a util function probably within the lib folder e.g. `packages/lib/universal/util/field-checkbox.ts` that you use between the two.


```typescript
const fromCheckboxValue = (customText: string): string[] => {
  if (!customText) {
    return []
  }

  try {
    const parsed = JSON.parse(customText);
    
    if (!Array.isArray(parsed)) {
      // Dummy throw so we can handle the `customText` as a string
      throw new Error('Parsed checkbox values are not an array');
    }

    return parsed;
  } catch {
    return customText.split(',').filter(Boolean);
  }
};

const toCheckboxValues = (values: string[]): string => {
  return JSON.stringify(values);
};
```

Mythie (approved):

PR #1547, opened 2024-12-17
feat: admin ui for disabling users

catalinpit:
My thoughts on how disabling should work.

When the user is disabled:
- disable the following:
  - all API keys - set the expiration date to `new Date()` (the current date and time)
  - all webhooks - set the `enabled` field to `false`
  -  verification tokens (if any) - set the expiration date to `new Date()` (the current date and time)
  - password reset tokens (if any) - set the expiration date to `new Date()` (the current date and time)

- delete the following:
  - passkeys (if any)

What about subscriptions, though?

What should we do with the teams created by the disabled user?

Any other thoughts?


packages/lib/server-only/user/enable-user.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const enableUser = async ({ id }: EnableUserOptions) => {
+  const user = await prisma.user.findFirst({
+    where: {
+      id,
+    },
+  });
+
+  if (!user) {
+    throw new Error('There was an error enabling the user');
```
dguyen, line 15: Use AppError with a NotFound error code

packages/trpc/server/admin-router/schema.ts
```diff
@@ -48,6 +48,20 @@ export const ZAdminDeleteUserMutationSchema = z.object({
...
+  id: z.number().min(1),
+  email: z.string().email(),
+});
+
+export type TAdminEnableUserMutationSchema = z.infer<typeof ZAdminEnableUserMutationSchema>;
+
+export const ZAdminDisableUserMutationSchema = z.object({
+  id: z.number().min(1),
+  email: z.string().email(),
+});
```
dguyen, line 61: Why are we passing email in? Can't we just use the ID?

apps/web/src/app/(dashboard)/admin/users/[id]/disable-user-dialog.tsx
```diff
@@ -0,0 +1,150 @@
...
+          title: _(msg`An unknown error occurred`),
+          variant: 'destructive',
+          description:
+            err.message ??
+            _(
+              msg`We encountered an unknown error while attempting to disable the account. Please try again later.`,
+            ),
+        });
+      }
+    }
```
Comment (dguyen) [inline apps/web/src/app/(dashboard)/admin/users/[id]/disable-user-dialog.tsx:75]: Don't follow this notation of rendering error messages directly from the backend.

It should look something like this

```typescript
} catch (err) {
  const error = AppError.parseError(err);

  const errorMessage = match(error.code)
    .with(
      AppErrorCode.NOT_FOUND,
      () => msg`Template not found or already associated with a team.`,
    )
    .with(AppErrorCode.UNAUTHORIZED, () => msg`You are not a member of this team.`)
    .otherwise(() => msg`An error occurred while moving the template.`);

  toast({
    title: _(msg`Error`),
    description: _(errorMessage),
    variant: 'destructive',
    duration: 7500,
  });
}
```

apps/web/src/app/(dashboard)/admin/users/[id]/disable-user-dialog.tsx
```diff
@@ -0,0 +1,150 @@
...
+  DialogHeader,
+  DialogTitle,
+  DialogTrigger,
+} from '@documenso/ui/primitives/dialog';
+import { Input } from '@documenso/ui/primitives/input';
+import { useToast } from '@documenso/ui/primitives/use-toast';
+
+export type DisableUserDialogProps = {
+  className?: string;
+  user: User;
```
Comment (dguyen) [inline apps/web/src/app/(dashboard)/admin/users/[id]/disable-user-dialog.tsx:29]: Since we use `user` in a lot of places to refer to the current user, I think it would be important to distinguish that this is a user being disabled, so something explicit like `userToDisable` would make more sense here

apps/web/src/app/(dashboard)/admin/users/[id]/disable-user-dialog.tsx
```diff
@@ -0,0 +1,150 @@
...
+        email,
+      });
+
+      toast({
+        title: _(msg`Account disabled`),
+        description: _(msg`The account has been disabled successfully.`),
+        duration: 5000,
+      });
+
+      router.push('/admin/users');
```
Comment (dguyen) [inline apps/web/src/app/(dashboard)/admin/users/[id]/disable-user-dialog.tsx:56]: I don't think a push is necessary since we're not deleting a resource here

packages/lib/server-only/user/disable-user.ts
```diff
@@ -0,0 +1,69 @@
...
+      ApiToken: true,
+      Webhooks: true,
+      passkeys: true,
+      VerificationToken: true,
+      PasswordResetToken: true,
+    },
+  });
+
+  if (!user) {
+    throw new Error('There was an error disabling the user');
```
dguyen, line 23: AppError

packages/lib/server-only/user/disable-user.ts
```diff
@@ -0,0 +1,69 @@
...
+      await tx.passwordResetToken.updateMany({
+        where: { userId: id },
+        data: {
+          expiry: new Date(),
+        },
+      });
+
+      await tx.passkey.deleteMany({
+        where: { userId: id },
+      });
```
dguyen, line 63: I think it's fine to leave these

dguyen:
> What about subscriptions, though?
> What should we do with the teams created by the disabled user?

Subscriptions should be fine to leave, if they are a paying customer I'd assume we would need to do a custom review

apps/web/src/app/(dashboard)/admin/users/[id]/disable-user-dialog.tsx
Comment (catalinpit) [reply apps/web/src/app/(dashboard)/admin/users/[id]/disable-user-dialog.tsx:56]: Yeah, I tried some stuff here (close the modal) and forgot to update this. 😅

packages/trpc/server/admin-router/schema.ts
catalinpit: Same functionality as deleting the user

dguyen: It doesn't make sense either way, remove the email and just use the id of the user to make it cleaner.

catalinpit: Already uses the ID to retrieve the user. 

The email is used to check if the retrieved user's email matches the provided email.

Will remove it though.

PR #1559, opened 2024-12-30
fix: default to user timezone

catalinpit:
## Description

Passes the timezone of the user uploading the document/template via the UI to the server.

If the user uploads a document/template via the API and doesn't set a timezone, it defaults to `Etc/UTC`. 

## Testing Performed

- Changed my machine's timezone and uploaded a document via the UI.
  ✅ The correct timezone was set.
- I created a template via the UI.
  ✅ The correct timezone was set.
- I generated a document from the template.
  ✅ The correct timezone was set.

- I created a document via the API without setting a timezone.
	  ✅ The default `UTC` timezone was set.
- I created a document via the API and specified a timezone.
  ✅ The correct timezone was set.

- I changed my machine's timezone again and repeated the whole process.
  ✅ The correct timezones were set.

## Checklist

### DIFF
--- a/apps/web/src/app/(dashboard)/documents/upload-document.tsx
+++ b/apps/web/src/app/(dashboard)/documents/upload-document.tsx
@@ -12,6 +12,7 @@ import { useSession } from 'next-auth/react';
 import { useLimits } from '@documenso/ee/server-only/limits/provider/client';
 import { useAnalytics } from '@documenso/lib/client-only/hooks/use-analytics';
 import { APP_DOCUMENT_UPLOAD_SIZE_LIMIT } from '@documenso/lib/constants/app';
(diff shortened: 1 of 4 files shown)


apps/web/src/app/(dashboard)/documents/upload-document.tsx
```diff
@@ -33,6 +33,7 @@ export type UploadDocumentProps = {
 export const UploadDocument = ({ className, team }: UploadDocumentProps) => {
   const router = useRouter();
   const analytics = useAnalytics();
+  const userTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
```
Mythie, line 36: Just need to check if it's in the `TIME_ZONES` export I think?

Mythie (approved):

PR #1569, opened 2025-01-03
feat: qr code on signing certificate

ephraimduncan:
## Demo

https://github.com/user-attachments/assets/66a1e744-e924-4c79-bbaf-8e6ace7f894a

## Description

Adds document access tokens and QR code functionality to enable secure document sharing via URLs. It includes a new document access page that allows viewing and downloading documents through tokenized links.

## Changes Made

- Added new `DocumentAccessToken` model and migration to support secure document access
- Created new document access page at `/q/[token]` for viewing shared documents
- Added QR code generation to signing certificates using `uqr` library

## Checklist

- [x] I have tested these changes locally and they work as expected
- [x] I have added appropriate database migrations
- [x] I have updated the documentation to reflect the new sharing functionality
- [x] I have followed the project's coding style guidelines
- [x] I have implemented proper error handling and security measures
- [x] I have tested the changes across different browsers and devices


apps/web/package.json
```diff
@@ -44,6 +44,7 @@
     "perfect-freehand": "^1.2.0",
     "posthog-js": "^1.75.3",
     "posthog-node": "^3.1.1",
+    "qrcode.react": "^4.2.0",
```
Mythie, line 47: We already have unqr for QR codes 🤔

ephraimduncan: i'm going to switch to it.

ephraimduncan: I'm going to switch from this to that.

Mythie:
I noticed and also kind of forgot that this QR code is basically only usable to document owners. This may require a new page be created for viewing/downloading the completed document using a long lived token.

Then on the page we can see if the viewer is authenticated and has access to the document so we can give them the option to view it within that context instead.

ephraimduncan:
@Mythie how does the current implementation look?

ephraimduncan:
Closing in favor of this #1755

PR #1574, opened 2025-01-08
chore: use current locale for new documents

ephraimduncan:


Mythie:
Was this requested by anyone, while timezones are inherently something you want to default I'm not 100% on language.

You may be a french speaker and so having the UI in French feels natural while the documents you send are likely to be in english if they're cross border.

PR #1578, opened 2025-01-11
feat: bulk send templates via csv

### LINKED ISSUE #1550: Bulk Send Template via CSV
On the template view page `/templates/[id]` or when on the template table `/templates` we should have an option either in the top corner or within the actions menu to bulk send the given template using a CSV ("Bulk send via CSV"). This will display a dialog that will let the user upload a given CSV which contains the values needed for the bulk send operation.

The CSV upload dialog will also include an option to download a template CSV file that is pre-formatted with the correct number of columns for the template's recipients. This template CSV will also include a single row of example values to guide users in proper formatting.

The CSV file should align with the recipient order in the template editor. For instance, a single recipient template will require one column, while a template with multiple recipients will have columns like `recipient_1`, `recipient_2`, etc. Blank values in the uploaded CSV will default to the preset values in the template, supporting scenarios where some recipient data remains fixed while others are dynamically replaced. 

To handle varying scales of uploads, the CSV processing will trigger a background task. This task will iterate over the file to create documents sequentially, with size limits of 4MB or 100 rows. Users will receive email notifications both upon completion and in case of any errors. This functionality is limited to the application UI, as API support for bulk sending is already achievable using custom logic.

### Summary

- Add a bulk send feature to templates view with a CSV upload dialog.
- Include an option to download a template CSV with the correct recipient columns and example values.
- Default blank CSV fields to preset template values.
- Process uploads using a background task with file size and row count limits.
- Notify users via email on task completion or errors.
- Restrict the feature to the application UI with no additional API support.

Mythie:
## Description

Implements a bulk send feature allowing users to upload a CSV file to create multiple documents from a template. Includes CSV template generation, background processing, and email notifications.

image
image
image

## Changes Made

- Added `TemplateBulkSendDialog` with CSV upload/download functionality
- Implemented bulk send job handler using background task system
- Created email template for completion notifications
- Added bulk send option to template view and actions dropdown
- Added CSV parsing with email/name validation

## Testing Performed

- CSV upload with valid/invalid data
- Bulk send with/without immediate sending
- Email notifications and error handling
- Team context integration
- File size and row count limits

Resolves #1550


Mythie:
@coderabbitai review

PR #1581, opened 2025-01-11
fix: refactor prisma relations

dguyen:
## Description

Refactor prisma relations to be all camel case for consistency.

## Testing

Todo:
- Double check backwards compatibility for V1 API
- Re-test everything


Mythie:
❤️

PR #1583, opened 2025-01-14
feat: add prisma json types

dguyen:
## Description

- Add types for Prisma json
- Bump typescript version (required)

## Testing Performed

- 2FA
- Normal document flow

### DIFF
--- a/packages/prisma/index.ts
+++ b/packages/prisma/index.ts
@@ -4,6 +4,7 @@ import kyselyExtension from 'prisma-extension-kysely';
 
 import type { DB } from './generated/types';
 import { getDatabaseUrl } from './helper';
+import './types/types.d.ts';
 import { remember } from './utils/remember';
 
 export const prisma = remember(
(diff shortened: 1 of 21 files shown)


packages/prisma/index.ts
```diff
@@ -4,6 +4,7 @@ import kyselyExtension from 'prisma-extension-kysely';
 
 import type { DB } from './generated/types';
 import { getDatabaseUrl } from './helper';
+import './types/types.d.ts';
```
Mythie, line 7: is that really the import order that prettier decided on?

I think this could be below everything

dguyen: Yeah prettier forces this one

Mythie (commented):
let me just check up on those UInt8Array calls, I feel like a few should be able to be dodged

Mythie:
Changes are fine, we will keep an eye on memory usage.

PR #1595, opened 2025-01-19
fix: correct redirect after document duplication

ephraimduncan:


apps/web/src/app/(dashboard)/documents/data-table-action-dropdown.tsx
```diff
@@ -233,14 +233,12 @@ export const DataTableActionDropdown = ({ row, team }: DataTableActionDropdownPr
         onOpenChange={setMoveDialogOpen}
       />
 
-      {isDuplicateDialogOpen && (
```
Mythie, line 236: I think we may have been killing it with the conditional here for a reason, can you test what happens if you open and close the dialog and then pick a different row to duplicate?

ephraimduncan: I don't think that will be a problem

CleanShot 2025-01-21 at 07 22 29

Mythie (approved):

PR #1598, opened 2025-01-20
feat: audit logs for copying links

ephraimduncan:


Mythie:
@dguyen What are your thoughts on this?

dguyen:
Note: `createAuditLog` should definitely not be a public endpoint

If we want logs, I think we should remove all instances of recipient tokens where possible, and have the copy token be a tRPC query which we can then log.

jebitese (approved):

ephraimduncan:
Closing this currently, will rework this.

PR #1607, opened 2025-01-23
chore: add label for checkbox and radio fields

catalinpit:
## Description

Added label for Radio and Checkbox fields based on the designs provided by Goutham.

## Changes Made

This is on the signing page.

https://github.com/user-attachments/assets/36f1958c-7066-4cf1-a81b-8ef909dfd049

This is in the document editor.

https://github.com/user-attachments/assets/cae8d708-1cfa-431c-b37e-3811ba7c700e


apps/web/src/app/(signing)/sign/[token]/signing-field-container.tsx
```diff
@@ -182,6 +182,23 @@ export const SigningFieldContainer = ({
           </button>
         )}
 
+        {(field.type === FieldType.RADIO || field.type === FieldType.CHECKBOX) &&
+          field.fieldMeta?.label && (
+            <div
+              className={cn(
+                'absolute -top-16 left-0 right-0 rounded-md p-2 text-center text-xs text-gray-700',
```
Comment (ephraimduncan) [inline apps/web/src/app/(signing)/sign/[token]/signing-field-container.tsx:189]: Can you consider extracting the label styling logic into a shared constant since it's duplicated in both `signing-field-container.tsx` and `field-item.tsx`?

Something like this will work.
```typescript
export const FIELD_LABEL_STYLES = {
  base: 'absolute -top-16 left-0 right-0 rounded-md p-2 text-center text-xs text-gray-700',
  unsigned: 'bg-foreground/5 border-border border',
  signed: 'bg-documenso-200 border-primary border'
};
```

Comment (catalinpit) [reply apps/web/src/app/(signing)/sign/[token]/signing-field-container.tsx:189]: Sure, but is it worth it for 2 places only?

Comment (ephraimduncan) [reply apps/web/src/app/(signing)/sign/[token]/signing-field-container.tsx:189]: currently, I dont think so but only if we'll extend it to ther places

Comment (catalinpit) [reply apps/web/src/app/(signing)/sign/[token]/signing-field-container.tsx:189]: Makes sense.

Mythie (dismissed):

Mythie (approved):

PR #1614, opened 2025-01-28
feat: admin monthly active users metric

ephraimduncan:
## Demo
CleanShot 2025-01-29 at 00 30 13@2x

## Description
Adds Monthly Active Users (MAU) charts to the admin statistics dashboard. It introduces both regular and cumulative MAU visualizations that track users who have signed in within the last year. 

## Related Issue
No specific issue reference was provided in the diff.

## Changes Made
- Added new `MonthlyActiveUsersChart` component for visualizing MAU data
- Implemented new `getMonthlyActiveUsers` database query in the admin stats service

## Checklist
- [x] I have tested these changes locally and they work as expected
- [x] I have followed the project's coding style guidelines (consistent with existing components)
- [x] I have updated the dashboard UI to include the new visualizations
- [x] The changes maintain existing accessibility standards


Mythie:
This probably shouldn't be a bar chart since lastSignedIn is a current value rather than historic. The chart will be deceiving.

If you want to do this it would be worth capturing session refreshes in the user security audit log (omitting them from other queries) and using that to formulate if a user was active in a given month

Mythie (changes requested):
per above

ephraimduncan:
@Mythie 

I've switched to using `UserSecurityAuditLog` with `SIGN_IN` events instead of `lastSignedIn`. 

Do you think we should also include session refreshes as you mentioned, or should we keep it to just sign-in events for now?

PR #1615, opened 2025-01-29
feat: allow fields prefill when generating a document from a template

catalinpit:
### V1 API Prefill Values

https://github.com/user-attachments/assets/92d5cd75-9048-4eb1-8dea-8c55e563febb

### V2 API Prefill Values

https://github.com/user-attachments/assets/07401698-a04a-4d39-9649-61bac81795ef

## Changes Made

The new change allows API users to pre-fill fields with values by passing the data in the request body. Example body for V2 API endpoint `/api/v2-beta/template/use`:

```json
{
    "templateId": 1,
    "recipients": [
        {
            "id": 1,
            "email": "signer1@mail.com",
            "name": "Signer 1"
        },
        {
            "id": 2,
            "email": "signer2@mail.com",
            "name": "Signer 2"
        }
    ],
    "prefillValues": [
        {
            "id": 14,
            "fieldMeta": {
                "type": "text",
                "label": "my label",
                "placeholder": "text placeholder test",
                "text": "auto-sign value",
                "characterLimit": 25,
                "textAlign": "right",
                "fontSize": 94,
                "required": true
            }
        },
        {
            "id": 15,
            "fieldMeta": {
                "type": "radio",
                "label": "radio label",
                "placeholder": "new radio placeholder",
                "required": false,
                "readOnly": true,
                "values": [
                    {
                        "id": 2,
                        "checked": true,
                        "value": "radio val 1"
                    },
                    {
                        "id": 3,
                        "checked": false,
                        "value": "radio val 2"
                    }
                ]
            }
        },
        {
            "id": 16,
            "fieldMeta": {
                "type": "dropdown",
                "label": "dropdown label",
                "placeholder": "DD placeholder",
                "required": false,
                "readOnly": false,
                "values": [
                    {
                        "value": "option 1"
                    },
                    {
                        "value": "option 2"
                    },
                    {
                        "value": "option 3"
                    }
                ],
                "defaultValue": "option 2"
            }
        }
    ],
    "distributeDocument": false,
    "customDocumentDataId": ""
}
```

## Testing Performed

I tested the changes by making requests via both the V1 and V2 APIs with various values.

## Checklist


packages/ui/primitives/document-flow/field-items-advanced-settings/dropdown-field.tsx
```diff
@@ -105,8 +105,12 @@ export const DropdownFieldAdvancedSettings = ({
           <Trans>Select default option</Trans>
         </Label>
         <Select
-          defaultValue={defaultValue}
+          value={defaultValue}
           onValueChange={(val) => {
+            if (!val) {
```
catalinpit, line 110: the select component wouldn't show the defaultValue when passed dynamically via the API... this fixes it.

packages/ui/primitives/document-flow/field-items-advanced-settings/text-field.tsx
```diff
@@ -126,7 +126,13 @@ export const TextFieldAdvancedSettings = ({
 
         <Select
           value={fieldState.textAlign}
-          onValueChange={(value) => handleInput('textAlign', value)}
+          onValueChange={(value) => {
+            if (!value) {
```
catalinpit, line 130: same as `dropdown-field.tsx`

packages/lib/server-only/template/create-document-from-template.ts
```diff
@@ -266,19 +272,23 @@ export const createDocumentFromTemplate = async ({
...
-          positionX: field.positionX,
-          positionY: field.positionY,
-          width: field.width,
-          height: field.height,
-          customText: '',
-          inserted: false,
-          fieldMeta: field.fieldMeta,
-        })),
+        fields.map((field) => {
+          const prefillValue = prefillValues?.find((value) => value?.id === field.id);
```
catalinpit, line 276: if the field doesn't exist, the app doesn't throw an error. it simply proceeds further without adding/modifying any field.

catalinpit: should we throw an error if the user tries to pre-fill a field that doesn't exist?

packages/lib/server-only/template/create-document-from-template.ts
```diff
@@ -50,6 +51,10 @@ export type CreateDocumentFromTemplateOptions = {
     email: string;
     signingOrder?: number | null;
   }[];
+  prefillValues?: {
+    id: number;
+    fieldMeta?: TFieldMetaSchema;
+  }[];
```
dguyen, line 57: Don't allow fieldMeta to be `undefined` for now

packages/lib/server-only/template/create-document-from-template.ts
dguyen: Should throw an error to let the user know that it didn't work as intended

packages/trpc/server/template-router/schema.ts
```diff
@@ -67,6 +68,14 @@ export const ZCreateDocumentFromTemplateRequestSchema = z.object({
       'The data ID of an alternative PDF to use when creating the document. If not provided, the PDF attached to the template will be used.',
     )
     .optional(),
+  prefillValues: z
+    .array(
+      z.object({
+        id: z.number(),
+        fieldMeta: ZFieldMetaSchema,
+      }),
+    )
```
dguyen, line 77: Rename `prefillValues` to `prefillFields` or something else

Add a `.describe` to the ID to let users know this is a `fieldId`

dguyen (changes requested):

packages/lib/server-only/template/create-document-from-template.ts
dguyen: Also, validate that the field type matches the field meta data

packages/lib/server-only/template/create-document-from-template.ts
catalinpit: TS will complain like crazy. 🤪

packages/trpc/server/template-router/schema.ts
catalinpit: Alright. To my broken english, `prefillValues` sounded more appropriate. 😅

Mythie:
Silly question, shouldn't this just a be a bit more streamlined?

We should only ever really want 
- `id`: the field id
- `label` it's reasonable to customise this
- `placeholder` it's the initial default value for the recipient

Taking in the entire field meta seems a bit strange, this would also let you simplify the validation logic since you're just taking a subset of the schema from `BaseFieldMetaSchema` ?

@dguyen thoughts?

dguyen:
I think it makes sense to just make it a subset of ZBaseFieldMeta for `prefilledField` like how Lucas said

If they want to change the fields we can rethink it then

packages/lib/server-only/template/create-document-from-template.ts
```diff
@@ -258,6 +264,23 @@ export const createDocumentFromTemplate = async ({
...
+      recipient.fields.map((field) => field.id),
+    );
+
+    if (prefillFields?.length) {
+      const invalidFieldIds = prefillFields
+        .map((prefillField) => prefillField.id)
+        .filter((id) => !allTemplateFieldIds.includes(id));
+
+      if (invalidFieldIds.length > 0) {
+        throw new Error(
```
ephraimduncan, line 278: Is there a specific reason you're using `Error` over `AppError`?

Issue #1342, opened 2024-09-10
OIDC LOGIN 


charlyschulte:
### Issue Description

i get the error: Unknown argument `refresh_expires_in`. Available options are marked with ?. 
i think because my refresh token never expires and therefore is 0 

### Expected Behavior

0 should be a valid in refresh_expires_in

### Please check the boxes that apply to this issue report.

- [X] I have searched the existing issues to make sure this is not a duplicate.
- [X] I have provided steps to reproduce the issue.
- [X] I have included relevant environment information.
- [X] I have included any relevant screenshots.
- [X] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [X] I want to work on creating a PR for this issue if approved

zvn2060:
Doesn't this cause by missing field in Prisma schema?
https://github.com/documenso/documenso/blob/b8310237e4589c128970cb0cdaeebd8862acf160/packages/prisma/schema.prisma#L242-L263

I encountered same error, but the `refresh_expires_in` is 1800

celevra:
ok i'm not a next.js or prisma expert, but i've change ext_expires_in to refresh_expires_in because to comment above was 
```
// Stops next-auth from crashing when dealing with AzureAD
```

after the change it goes a step further but only until the next variable

```
documenso-documenso-1  | Invalid `prisma.account.create()` invocation:
documenso-documenso-1  |
documenso-documenso-1  | {
documenso-documenso-1  |   data: {
documenso-documenso-1  |     provider: "oidc",
documenso-documenso-1  |     type: "oauth",
documenso-documenso-1  |     providerAccountId: "ce2f22a9-3651-4041-90c7-2c2c9bf43758",
documenso-documenso-1  |     access_token: "xxx",
documenso-documenso-1  |     expires_at: 1738782434,
documenso-documenso-1  |     refresh_expires_in: 1800,
documenso-documenso-1  |     refresh_token: "xxx",
documenso-documenso-1  |     token_type: "Bearer",
documenso-documenso-1  |     id_token: "xxx",
documenso-documenso-1  |     not-before-policy: 0,
documenso-documenso-1  |     ~~~~~~~~~~~~~~~~~
documenso-documenso-1  |     session_state: "d4518b2a-c36a-40b1-9a1b-bd3c0ebd9775",
documenso-documenso-1  |     scope: "openid profile email",
documenso-documenso-1  |     userId: 3,
documenso-documenso-1  | ?   id?: String,
[...]

PR #1637, opened 2025-02-06
fix: remove non-standard fields that would cause prisma validation er…

ephraimduncan:
## Description
This PR removes non-standard fields that would cause Prisma validation errors and more maintainable as it doesn't require schema changes for provider-specific fields.

## Related Issue
#1342

## Checklist

### DIFF
--- a/packages/lib/next-auth/auth-options.ts
+++ b/packages/lib/next-auth/auth-options.ts
@@ -193,7 +193,15 @@ export const NEXT_AUTH_OPTIONS: AuthOptions = {
       idToken: true,
       allowDangerousEmailAccountLinking: true,
 
-      profile(profile) {
+      profile(profile, tokens) {
+        if (tokens && 'refresh_expires_in' in tokens) {
+          delete tokens.refresh_expires_in;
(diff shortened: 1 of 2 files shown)


celevra:
could be that there are more fields in the future, so clone needed variables in a new object, so we don't have to know all the optional variables

packages/lib/next-auth/auth-options.ts
```diff
@@ -193,7 +193,15 @@ export const NEXT_AUTH_OPTIONS: AuthOptions = {
...
 
-      profile(profile) {
+      profile(profile, tokens) {
+        if (tokens && 'refresh_expires_in' in tokens) {
+          delete tokens.refresh_expires_in;
+        }
+
+        if (tokens && 'not-before-policy' in tokens) {
+          delete tokens['not-before-policy'];
+        }
```
Mythie, line 203: `delete` makes me sad, it'd be better to just deconstruct or validate the items we do need and work backwards from that.

I noticed that this also doesn't remove the scarier oidc code.

PR #1644, opened 2025-02-08
Update web.po

celevra:
there where some grammar failures


Mythie:

https://documen.so/cla

celevra:
@Mythie done

celevra:
i think there is still a problem, the actionVerb is used on multiple Locations but in german you need different words.

here is the actionVerb:
image


and here also:
documenso

in the e-mail the actionVerb should be "unterschreibe" and in the document table it is correct with "unterschreiben"
That means we need different variables in the E-Mailtemplate

PR #1648, opened 2025-02-14
fix: prepare auth migration

dguyen:
## Description

Add schema migration in preparation for auth migration.

Should have no issues since `Session` table should not currently be in use.


Mythie (changes requested):
missing the migration file

dguyen:
@Mythie added migration file

PR #1649, opened 2025-02-14
feat: customize doc audit logs and certificate

catalinpit:
https://github.com/user-attachments/assets/36f09960-0da4-4f04-8350-2cb7c6841375


Mythie (changes requested):
A split button would be nice 😄 

https://v0.dev/chat/split-button-component-UVUaTVksS7p?b=b_cnd4FMePvTQ&p=0

packages/trpc/server/document-router/router.ts
```diff
@@ -1,5 +1,6 @@
 import { TRPCError } from '@trpc/server';
```
Mythie, line 1: This implementation won't work. Any changes to the PDF after signing will invalidate it.

We should collect whether they want the audit log attached to the pdf as a document option, with a team preference.

Then the dropdown will let you optionally download the additional items which will be done as separate downloads.

e.g. `filename_signed.pdf`, `filename_audit.pdf`, `filename_certificate.pdf`

ElTimuro:
- @catalinpit, which option is being sent to recipients?
- Can that be customized e.g. always sending with audit log?
- https://github.com/documenso/backlog-internal/issues/376 could be included here so we are done

PR #1660, opened 2025-02-19
feat: use multiselect for webhook triggers

ephraimduncan:
## Demo

CleanShot 2025-02-18 at 18 01 05


dguyen:
It is actually necessarily to create a whole new multi select component? Can't you use the primitive?

ephraimduncan:
Ideally, you could build a multiselect using the existing select primitive, but you'll need to create custom logic for tracking multiple selections and rendering badges and that would require significant customization beyond what the basic select provides.

and the multiselect component makes it easy to have multiselect.

Mythie:
Changes implemented in: https://github.com/documenso/documenso/pull/1795

PR #1676, opened 2025-02-26
feat: hide signature ui when theres no signature field

ephraimduncan:
## demo

CleanShot 2025-02-26 at 23 25 52@2x

### DIFF
--- a/apps/web/src/app/(signing)/sign/[token]/form.tsx
+++ b/apps/web/src/app/(signing)/sign/[token]/form.tsx
@@ -311,7 +311,11 @@ export const SigningForm = ({
             <>
               <form onSubmit={handleSubmit(onFormSubmit)}>
                 <p className="text-muted-foreground mt-2 text-sm">
-                  <Trans>Please review the document before signing.</Trans>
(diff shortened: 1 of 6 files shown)


apps/web/src/app/(signing)/sign/[token]/form.tsx
```diff
@@ -335,38 +339,40 @@ export const SigningForm = ({
...
+                                if (signatureValid) {
+                                  setSignature(value);
+                                }
+                              }}
+                              allowTypedSignature={document.documentMeta?.typedSignatureEnabled}
+                            />
+                          </CardContent>
+                        </Card>
+
+                        {hasSignatureField && !signatureValid && (
```
Comment (catalinpit) [inline apps/web/src/app/(signing)/sign/[token]/form.tsx:367]: is this `hasSignatureField` required if we check above?

catalinpit (approved):

apps/web/src/app/(signing)/sign/[token]/form.tsx
Comment (Mythie) [reply apps/web/src/app/(signing)/sign/[token]/form.tsx:367]: it's not

PR #1683, opened 2025-03-04
feat: bulk add fields

ephraimduncan:
## Demo

CleanShot 2025-03-04 at 02 17 47


packages/ui/primitives/document-flow/add-fields.tsx
```diff
@@ -653,6 +680,9 @@ export const AddFieldsFormPartial = ({
                       onMove={(options) => onFieldMove(options, index)}
                       onRemove={() => remove(index)}
                       onDuplicate={() => onFieldCopy(null, { duplicate: true })}
+                      onDuplicateAllPages={() =>
+                        onFieldCopy(null, { duplicate: true, duplicateAllPages: true })
```
catalinpit, line 684: Is `duplicate: true` required here?

packages/ui/primitives/document-flow/add-fields.tsx
```diff
@@ -428,6 +430,31 @@ export const AddFieldsFormPartial = ({
           return;
         }
 
+        if (duplicateAllPages) {
+          const pages = Array.from(document.querySelectorAll(PDF_VIEWER_PAGE_SELECTOR));
+          const totalPages = pages.length;
+
+          for (let pageNumber = 1; pageNumber <= totalPages; pageNumber++) {
+            if (pageNumber !== lastActiveField.pageNumber) {
+              const newField: TAddFieldsFormSchema['fields'][0] = {
```
catalinpit, line 439: this is more like nitpicking, but maybe we could extract this into a constant and re-use it. but feel free to ignore this comment.

catalinpit (approved):

packages/ui/primitives/document-flow/add-fields.tsx
ephraimduncan: good catch but yes, `duplicate: true` is required here,

currently this is the flow: 

if `duplicate` is false, it only copies the field to clipboard and returns early
if `duplicate` is true and `duplicateAllPages` is true, it creates copies across all pages
if `duplicate` is true and `duplicateAllPages` is false, it creates a single duplicate

packages/ui/primitives/document-flow/add-fields.tsx
ephraimduncan: >  we could extract this into a constant and re-use it

could you clarify with an example?

Mythie (approved):

Mythie (approved):

PR #1696, opened 2025-03-11
fix: autosigning fields with direct links

catalinpit:
## Description

The changes in `sign-direct-template.tsx` automatically fill in field values for text, number, and dropdown fields when default values are present or if the fields are read-only. In `checkbox-field.tsx`, the changes fix the checkbox signing by checking if the validation is met and improving how it saves or removes checkbox choices.

## Testing Performed

I tested the code locally with a variety of documents/fields.

## Checklist

### DIFF
--- a/apps/remix/app/components/general/direct-template/direct-template-signing-form.tsx
+++ b/apps/remix/app/components/general/direct-template/direct-template-signing-form.tsx
@@ -17,6 +17,7 @@ import {
   ZTextFieldMeta,
 } from '@documenso/lib/types/field-meta';
 import type { TTemplate } from '@documenso/lib/types/template';
+import { toCheckboxValue } from '@documenso/lib/universal/field-checkbox';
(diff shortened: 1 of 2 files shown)


apps/web/src/app/(recipient)/d/[token]/sign-direct-template.tsx
```diff
@@ -169,6 +169,55 @@ export const SignDirectTemplateForm = ({
     // Do not reset to false since we do a redirect.
   };
 
+  useEffect(() => {
```
Comment (catalinpit) [inline apps/web/src/app/(recipient)/d/[token]/sign-direct-template.tsx:172]: this signs the non-radio/non-checkboxes advanced fields if they have a default value. without this useEffect, they don't automatically sign if the document contains checkboxes/radio fields.

maybe I'm missing something.

Mythie (dismissed):

Mythie:
Can this target RR7 instead please :)

PR #1697, opened 2025-03-11
fix: rework fields

dguyen:
## Description

Rework:
- Field styling to improve visibility
- Field insertions, better alignment, centering and overflows

## Changes

General changes:

- Set default text alignment to left if no meta found
- Reduce borders and rings around fields to allow smaller fields
- Removed lots of redundant duplicated code surrounding field rendering
- Make fields more consistent across viewing, editing and signing
- Add more transparency to fields to allow users to see under fields
- No more optional/required/etc colors when signing, required fields will be highlighted as orange when form is "validating"

Highlighted internal changes:

- Utilize native PDF fields to insert text, instead of drawing text 
- Change font auto scaling to only apply to when the height overflows AND no custom font is set

⚠️ Multiline changes:

Multi line is enabled for a field under these conditions

1. Field content exceeds field width
2. Field includes a new line
3. Field type is TEXT

## [BEFORE] Field UI Signing 

image

## [AFTER] Field UI Signing 

image

## [BEFORE] Signing a checkbox

image
image

## [AFTER] Signing a checkbox

image
image

## [BEFORE] What a 2nd recipient sees once someone else signed a document

image

## [AFTER] What a 2nd recipient sees once someone else signed a document

image

## **[BEFORE]** Inserting fields

image

## **[AFTER]** Inserting fields

image

## Overflows, multilines and field alignments testing

Debugging borders:
- Red border = The original field placement without any modifications
- Blue border = The available space to overflow

### Single line overflows and field alignments 

This is left aligned fields, overflow will always go to the end of the page and will not wrap

image

This is center aligned fields, the max width is the closest edge to the page * 2

image

This is right aligned text, the width will extend all the way to the left hand side of the page

image

### Multiline line overflows and field alignments 

These are text fields that can be overflowed

image

Another example of left aligned text overflows with more text

image


apps/remix/app/components/general/document-signing/document-signing-checkbox-field.tsx
```diff
@@ -97,6 +97,11 @@ export const DocumentSigningCheckboxField = ({
 
   const onSign = async (authOptions?: TRecipientActionAuth) => {
     try {
+      // Do nothing, since the user clicked on a non checkbox area.
+      if (checkedValues.length === 0) {
```
Mythie, line 101: Is this right? Presumably there will be times where no checked items are relevant.

I assume we just need the length condition met variable instead.

dguyen: - If no checked items are relevant, then the onSign should never be called.
- If the user wants to remove a checkbox then `onRemove` callback runs instead

This if statement should only occur if the user clicks the field, but misses the checkbox (since we use an absolute button to cover the field)

It's cooked on how we use an absolute inset button for checkboxes/radios for that's for another time

Mythie (approved):

PR #1702, opened 2025-03-13
feat: better document rejection

Mythie:
## Description

Improves the existing document rejection process by actually marking a document as completed cancelling further actions.

## Related Issue

N/A

## Changes Made

- Added a new rejection status for documents
- Updated a million areas that check for document completion
- Updated email sending, so rejection is confirmed for the rejecting recipient while other recipients are notified that the document is now cancelled.

## Testing Performed

- Ran the testing suite to ensure there are no regressions.
- Performed manual testing of current core flows.


dguyen (approved):
lgtm

PR #1749, opened 2025-04-15
feat: signing reminders

### LINKED ISSUE #2147: Allow user to enable automatic reminder emails for recipients who haven't signed
### Feature Description

From what I can tell there's no automated way to send reminder emails to recipients who haven't signed a document. It would be great to have an option to automatically send reminder emails every 7 days (for example), available as a checkbox when you're initially sending the document.

### Use Case

Right now, users who send documents for signing have to manually keep track of which documents have not yet been signed. It would be great to have the option of having Documenso automatically follow up with any recipient who hasn't signed after (say) 7 days, and every 7 days thereafter. By checking a checkbox on step 4 of the document sending process (Distribute Document), the user could opt into having Documenso automatically follow up with non-compliant recipients.

### Proposed Solution

This option could appear as a new checkbox on the Distribute Document panel (step 4 of 4 when preparing a document to be sent). The label could be, "Send reminder emails every 7 days", and the tooltip/help copy could be have the headline "Reminder emails" and the body "Emails will be sent every 7 days to any recipient who hasn't signed the document."

(I'm not an engineer, so I don't know how exactly this feature would be implemented, but I assume it would require some sort of cron job process.)

### Alternatives (optional)

The only alternative, I think, is to manually check on the document status and to resend reminders manually. (I'm glad that feature exists!)

### Additional Context

_No response_

ephraimduncan:
## Demo

CleanShot 2025-04-18 at 18 50 17


packages/email/template-components/template-document-reminder.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const { _ } = useLingui();
+
+  const { actionVerb } = RECIPIENT_ROLES_DESCRIPTION[role];
+
+  return (
+    <>
+      <TemplateDocumentImage className="mt-6" assetBaseUrl={assetBaseUrl} />
+
+      <Section>
+        {/* Reminder specific text */}
```
catalinpit: these comments can be removed I think

packages/email/template-components/template-document-reminder.tsx
```diff
@@ -0,0 +1,79 @@
...
+        </Text>
+
+        {/* Primary Action Button */}
+        <Section className="mb-6 mt-8 text-center">
+          <Button
+            className="bg-documenso-500 inline-flex items-center justify-center rounded-lg px-6 py-3 text-center text-sm font-medium text-black no-underline"
+            href={signDocumentLink}
+          >
+            {match(role)
+              .with('SIGNER', () => <Trans>Sign Document</Trans>)
```
catalinpit: should we use `RecipientRole` instead of hardcoding the values?

packages/email/templates/document-reminder.tsx
```diff
@@ -0,0 +1,89 @@
...
+  assetBaseUrl = 'http://localhost:3002',
+  customBody,
+  role,
+}: DocumentReminderEmailTemplateProps) => {
+  const { i18n } = useLingui();
+  const branding = useBranding();
+
+  const action = i18n.t(RECIPIENT_ROLES_DESCRIPTION[role].actionVerb).toLowerCase();
+
+  const previewTextString = `Reminder to ${action} ${documentName}`;
```
catalinpit: ```msg`Reminder to ${action} ${documentName}`;```

packages/lib/constants/document-audit-logs.ts
```diff
@@ -19,4 +19,7 @@ export const DOCUMENT_AUDIT_LOG_EMAIL_FORMAT = {
   [DOCUMENT_EMAIL_TYPE.DOCUMENT_COMPLETED]: {
     description: 'Document completed',
   },
+  [DOCUMENT_EMAIL_TYPE.REMINDER]: {
+    description: 'Reminder',
```
catalinpit: should this be more descriptive such as "Signing Reminder"?

packages/ui/primitives/document-flow/add-settings.tsx
```diff
@@ -383,6 +385,67 @@ export const AddSettingsFormPartial = ({
...
+
+                          <FormControl>
+                            <Select {...field} onValueChange={field.onChange}>
+                              <SelectTrigger className="bg-background">
+                                <SelectValue />
+                              </SelectTrigger>
+
+                              <SelectContent>
+                                <SelectItem value={DocumentReminderInterval.NONE}>
+                                  No reminders
```
catalinpit: do these options need to be wrapped in the `Trans` tag as well?

ephraimduncan: yes, will fix it

packages/email/templates/document-reminder.tsx
ephraimduncan: thank you

packages/email/template-components/template-document-reminder.tsx
ephraimduncan: good catch

Mythie (approved):

PR #1758, opened 2025-04-23
feat: complete document 2fa

ephraimduncan:
### Document and Template Setting

https://github.com/user-attachments/assets/75c647bb-0ae6-4222-b836-ecebadab383e

### Emailing a given 2FA Token if a user does not have a 2FA secret created

https://github.com/user-attachments/assets/45fb53d5-dcd1-481d-bd69-3b90479cd81e

### Emailing a given 2FA Token if a user has a 2FA secret created

https://github.com/user-attachments/assets/b981b279-b960-4df9-b74a-70a0b0c83832


packages/lib/server-only/document/update-document.ts
```diff
@@ -135,18 +133,18 @@ export const updateDocument = async ({
     data?.globalActionAuth === undefined ? documentGlobalActionAuth : data.globalActionAuth;
 
   // Check if user has permission to set the global action auth.
-  if (newGlobalActionAuth) {
-    const isDocumentEnterprise = await isUserEnterprise({
-      userId,
-      teamId,
-    });
+  // if (newGlobalActionAuth) {
```
catalinpit, line 136: Is all of this code meant to be left commented?

apps/remix/app/components/general/document-signing/document-signing-text-field.tsx
```diff
@@ -62,7 +62,7 @@ export const DocumentSigningTextField = ({
   const [errors, setErrors] = useState(initialErrors);
   const userInputHasErrors = Object.values(errors).some((error) => error.length > 0);
 
-  const { executeActionAuthProcedure } = useRequiredDocumentSigningAuthContext();
+  const { executeActionAuthProcedure, isEnterprise } = useRequiredDocumentSigningAuthContext();
```
catalinpit, line 65: Where is `isEnterprise` being used here?

catalinpit: Same for the other fields

catalinpit (commented):
Do we also need to update the documentation? One example here: https://docs.documenso.com/developers/public-api/reference#create-document

You can search for `Enterprise accounts` on the page to find all the relevant bits

apps/remix/app/components/general/document-signing/document-signing-text-field.tsx
ephraimduncan: It's not being used, i'll remove it

packages/ui/components/document/document-global-auth-action-select.tsx
```diff
@@ -17,38 +21,51 @@ import {
...
-  },
-);
+      <SelectContent position="popper">
+        {/* Note: -1 is remapped in the Zod schema to the required value. */}
+        <SelectItem value={'-1'}>
+          <Trans>No restrictions</Trans>
+        </SelectItem>
+
+        {isDocumentEnterprise
+          ? Object.values(DocumentActionAuth)
```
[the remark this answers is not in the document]
ephraimduncan: lovely

ephraimduncan:
Closing this in favor of this #1917

PR #1760, opened 2025-04-25
feat: add document attachments feature

catalinpit:
---
name: Pull Request
---

## Changes Made

- Change 1
- Change 2
- ...

## Testing Performed

- ...

## Checklist

### DIFF
--- a/apps/remix/app/components/general/document-signing/document-signing-form.tsx
+++ b/apps/remix/app/components/general/document-signing/document-signing-form.tsx
@@ -4,8 +4,9 @@ import { msg } from '@lingui/core/macro';
 import { useLingui } from '@lingui/react';
 import { Trans } from '@lingui/react/macro';
 import { type Field, FieldType, type Recipient, RecipientRole } from '@prisma/client';
+import { Link as LinkIcon } from 'lucide-react';
 import { Controller, useForm } from 'react-hook-form';
-import { useNavigate } from 'react-router';
+import { Link, useNavigate } from 'react-router';
 
 import { useAnalytics } from '@documenso/lib/client-only/hooks/use-analytics';
 import { useOptionalSession } from '@documenso/lib/client-only/providers/session';
@@ -17,6 +18,12 @@ import type { RecipientWithFields } from '@documenso/prisma/types/recipient-with
 import { trpc } from '@documenso/trpc/react';
 import { FieldToolTip } from '@documenso/ui/components/field/field-tooltip';
 import { cn } from '@documenso/ui/lib/utils';
+import {
+  Accordion,
+  AccordionContent,
+  AccordionItem,
+  AccordionTrigger,
+} from '@documenso/ui/primitives/accordion';
 import { Button } from '@documenso/ui/primitives/button';
 import { Input } from '@documenso/ui/primitives/input';
 import { Label } from '@documenso/ui/primitives/label';
@@ -369,6 +376,35 @@ export const DocumentSigningForm = ({
                         onChange={(e) => setFullName(e.target.value.trimStart())}
                       />
                     </div>
+
+                    {document.attachments && (
+                      <Accordion type="multiple" className="mt-2">
+                        <AccordionItem value="attachments" className="border-none">
+                          <AccordionTrigger className="text-foreground mb-2 rounded border px-3 py-2 text-left hover:bg-neutral-200/30 hover:no-underline">
+                            <Trans>Attachments</Trans>
+                          </AccordionTrigger>
+
+                          <AccordionContent className="-mx-1 px-1 pt-2 text-sm leading-relaxed">
+                            <div className="flex flex-col space-y-2">
+                              {document.attachments.map((attachment, index) => (
+                                <div key={index}>
+                                  <Button variant="outline" asChild>
+                                    <Link
+                                      to={attachment.url}
+                                      target="_blank"
+                                      rel="noopener noreferrer"
+                                    >
+                                      <LinkIcon className="mr-2 h-4 w-4" />
+                                      {attachment.label}
+                                    </Link>
+                                  </Button>
+                                </div>
+                              ))}
+                            </div>
+                          </AccordionContent>
+                        </AccordionItem>
+                      </Accordion>
+                    )}
 
                     {hasSignatureField && (
                       <div>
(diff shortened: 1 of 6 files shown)


apps/remix/app/components/general/template/template-edit-form.tsx
```diff
@@ -165,6 +166,14 @@ export const TemplateEditForm = ({
       await Promise.all([
         updateTemplateSettings({
           templateId: template.id,
+          data: {
```
catalinpit, line 169: it overrides the attachments and sets them to an empty array when adding the template placeholders, hence why this bit of code.

ephraimduncan (approved):

Mythie:
Without looking too closely at the code, the UX looks like it could use some improvement.

During the authoring experience we should have attachments sit outside of the stepper above the document like we do for the legacy field warning, but instead of a popover it displays a dialog with the attachment form:

image

During the signing experience I think we would probably want similar where we show like "(X) Attachments" which opens a dialog containing the attachments.

catalinpit:
> Without looking too closely at the code, the UX looks like it could use some improvement.
> 
> During the authoring experience we should have attachments sit outside of the stepper above the document like we do for the legacy field warning, but instead of a popover it displays a dialog with the attachment form:
> 
> image
> During the signing experience I think we would probably want similar where we show like "(X) Attachments" which opens a dialog containing the attachments.

Updated!

packages/lib/server-only/attachment/set-document-attachments.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export const setDocumentAttachments = async ({
+  documentId,
+  attachments,
+  user,
+  requestMetadata,
+}: CreateAttachmentsOptions) => {
+  const document = await prisma.document.findUnique({
+    where: {
+      id: documentId,
```
Mythie: Need the team where query here otherwise you can update the attachment for any document with a crafted request.

packages/lib/server-only/attachment/set-document-attachments.ts
```diff
@@ -0,0 +1,110 @@
...
+      },
+    });
+  }
+
+  const upsertedAttachments: Attachment[] = [];
+
+  for (const attachment of attachments) {
+    if (attachment.id) {
+      const updated = await prisma.attachment.upsert({
+        where: { id: attachment.id },
```
Mythie: Needs the documentId in the where query as well otherwise you could update any attachment even if it's not related to the given document

packages/lib/server-only/attachment/set-template-attachments.ts
```diff
@@ -0,0 +1,80 @@
...
+  attachments: Pick<Attachment, 'id' | 'label' | 'url' | 'type'>[];
+};
+
+export const setTemplateAttachments = async ({
+  templateId,
+  attachments,
+}: CreateAttachmentsOptions) => {
+  const template = await prisma.template.findUnique({
+    where: {
+      id: templateId,
```
Mythie: Need the team where query here otherwise you can update the attachment for any template with a crafted request.

packages/lib/server-only/attachment/set-template-attachments.ts
```diff
@@ -0,0 +1,80 @@
...
+      },
+    });
+  }
+
+  const upsertedAttachments: Attachment[] = [];
+
+  for (const attachment of attachments) {
+    if (attachment.id) {
+      const updated = await prisma.attachment.upsert({
+        where: { id: attachment.id },
```
Mythie: Need the same as the document one for the same reasons.

packages/lib/server-only/attachment/set-template-attachments.ts
```diff
@@ -0,0 +1,80 @@
...
+      where: {
+        id: { in: toDelete.map((a) => a.id) },
+      },
+    });
+  }
+
+  const upsertedAttachments: Attachment[] = [];
+
+  for (const attachment of attachments) {
+    if (attachment.id) {
```
Mythie: If we're upserting we don't need to handle the if/else here since it'll do the update or create for us.

packages/lib/server-only/attachment/set-document-attachments.ts
```diff
@@ -0,0 +1,110 @@
...
+      where: {
+        id: { in: toDelete.map((a) => a.id) },
+      },
+    });
+  }
+
+  const upsertedAttachments: Attachment[] = [];
+
+  for (const attachment of attachments) {
+    if (attachment.id) {
```
Mythie: If we're upserting we don't need to handle the if/else here since it'll do the update or create for us.

packages/trpc/server/attachment-router/router.ts
```diff
@@ -0,0 +1,92 @@
+import { findDocumentAttachments } from '@documenso/lib/server-only/attachment/find-document-attachments';
```
dguyen: We don't need a new attachments router. See the organisation router on how we plan to structure new routes.

It would look something like this in the document router:
```
router({
  attachments:{
    get: getDocumentAttachmentRoute,
    getMany: getDocumentAttachmentsRoute,
    set // etc
  }
})
```

apps/remix/app/components/general/template/template-attachment-form.tsx
```diff
@@ -0,0 +1,181 @@
...
+        variant: 'destructive',
+        duration: 5000,
+      });
+    }
+  };
+
+  return (
+    <Dialog>
+      <DialogTrigger asChild>
+        <Button variant="outline">Attachments</Button>
```
dguyen: Missing translations, review the whole PR again to make sure we have it everywhere

apps/remix/app/components/general/template/template-attachment-form.tsx
```diff
@@ -0,0 +1,181 @@
...
+        title: 'Attachment(s) updated',
+        description: 'The attachment(s) have been updated successfully',
+      });
+
+      await refetchAttachments();
+    } catch (error) {
+      console.error(error);
+
+      toast({
+        title: 'Something went wrong',
```
dguyen: Translations

dguyen (changes requested):
Could you also remove the PO files from this PR

packages/trpc/server/attachment-router/router.ts
catalinpit: But the document router is not using this approach yet. Or what am I missing?

dguyen: We're slowly rolling it out, we can have both systems at once

PR #1762, opened 2025-04-26
chore: update French translations

tom-rh:
✨ Improved French translations for a more natural and fluent feel!


dguyen (approved):
Thanks for that!

PR #1815, opened 2025-05-29
feat: add document processing status indicator

ephraimduncan:
## Demo

- Add visual differentiation between documents waiting for signatures vs processing
- Show loading spinner when document is processing after all signatures collected

### DIFF
--- a/apps/remix/app/components/general/document/document-processing-poll.tsx
+++ b/apps/remix/app/components/general/document/document-processing-poll.tsx
@@ -1,0 +1,54 @@
+import { useEffect } from 'react';
+
+import type { DocumentStatus, RecipientRole, SigningStatus } from '@prisma/client';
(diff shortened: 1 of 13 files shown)


apps/remix/app/components/general/document/document-processing-poll.tsx
ephraimduncan, line 1: This breaks in some situations

1. Keeps refreshing the ui so dialogs and pop ups are thrown away
2. Even when the document is completed, sometimes you have to refresh the page to get the latest change rendering this useless.

dguyen:
Rebased onto main, will put this on hold a bit until we see if there's an alternative to using revalidate which that causes ui glitches

Preferably we would invalidate the trpc query but it seems like we don't use trpc in all the places that currently show the status

PR #1853, opened 2025-06-24
chore: add support option

catalinpit:
---
name: Pull Request
---

## Changes Made

- Change 1
- Change 2
- ...

## Testing Performed

- ...

## Checklist

### DIFF
--- a/apps/remix/app/components/general/org-menu-switcher.tsx
+++ b/apps/remix/app/components/general/org-menu-switcher.tsx
@@ -321,6 +321,12 @@ export const OrgMenuSwitcher = () => {
                 <Trans>Language</Trans>
               </DropdownMenuItem>
 
+              <DropdownMenuItem className="text-muted-foreground px-4 py-2" asChild>
+                <Link to="/support">
+                  <Trans>Support</Trans>
+                </Link>
+              </DropdownMenuItem>
+
               <DropdownMenuItem
                 className="text-muted-foreground hover:!text-muted-foreground px-4 py-2"
                 onSelect={async () => authClient.signOut()}
--- a/apps/remix/app/routes/_authenticated+/support.tsx
+++ b/apps/remix/app/routes/_authenticated+/support.tsx
@@ -1,0 +1,64 @@
+import { Trans } from '@lingui/react/macro';
+import { HelpCircleIcon, Link2Icon, MailIcon } from 'lucide-react';
+import { Link } from 'react-router';
+
+export default function SupportPage() {
+  return (
+    <div className="mx-auto w-full max-w-screen-xl px-4 md:px-8">
+      <div className="mb-8">
+        <h1 className="flex flex-row items-center gap-2 text-3xl font-bold">
+          <HelpCircleIcon className="text-muted-foreground h-8 w-8" />
+          <Trans>Support</Trans>
+        </h1>
+
+        <p className="text-muted-foreground mt-1">
+          <Trans>Choose a support channel below to get help with Documenso</Trans>
+        </p>
+
+        <div className="mt-6 flex flex-col gap-4">
+          <div className="rounded-lg border p-4">
+            <h2 className="flex items-center gap-2 text-lg font-bold">
+              <MailIcon className="text-muted-foreground h-5 w-5" />
+              <Link
+                to="mailto:support@documenso.com"
+                target="_blank"
+                rel="noopener noreferrer"
+                className="hover:underline"
+              >
+                <Trans>Email</Trans>
+              </Link>
+            </h2>
+
+            <p className="text-muted-foreground mt-1">
+              <Trans>
+                Send an email to <a href="mailto:support@documenso.com">support@documenso.com</a>.
+              </Trans>
+            </p>
+          </div>
+          <div className="rounded-lg border p-4">
+            <h2 className="flex items-center gap-2 text-lg font-bold">
+              <Link2Icon className="text-muted-foreground h-5 w-5" />
+              <Link
+                to="https://documen.so/discord"
+                target="_blank"
+                rel="noopener noreferrer"
+                className="hover:underline"
+              >
+                <Trans>Discord</Trans>
+              </Link>
+            </h2>
+            <p className="text-muted-foreground mt-1">
+              <Trans>
+                Join our community on{' '}
+                <a href="https://documen.so/discord" target="_blank" rel="noopener noreferrer">
+                  Discord
+                </a>{' '}
+                for community support and discussion.
+              </Trans>
+            </p>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+}


apps/remix/server/api/files.ts
```diff
@@ -72,7 +72,7 @@ export const filesRoute = new Hono<HonoEnv>()
       return c.json({ error: 'Upload failed' }, 500);
     }
   })
-  .post('/presigned-get-url', sValidator('json', ZGetPresignedGetUrlRequestSchema), async (c) => {
+  .post('/upload-pdf', sValidator('json', ZGetPresignedGetUrlRequestSchema), async (c) => {
```
catalinpit: mistake. revert this

apps/remix/app/components/general/org-menu-switcher.tsx
```diff
@@ -321,6 +321,12 @@ export const OrgMenuSwitcher = () => {
                 <Trans>Language</Trans>
               </DropdownMenuItem>
 
+              <DropdownMenuItem className="text-muted-foreground px-4 py-2" asChild>
+                <Link to={`/o/${currentOrganisation?.url}/support`}>
```
catalinpit: this will throw an error if the user navigates to this link from a non-org link. for example, if you visit this link from `https://app.documenso.com/settings/profile`, it'll throw an error. what would be your suggestion?

dguyen: Couldn't you just do `/settings/support` instead of having it on the organisation path?

You can pass the org ID in the query if you want access to it

.env.example
```diff
@@ -127,6 +127,10 @@ E2E_TEST_AUTHENTICATE_USER_EMAIL="testuser@mail.com"
 E2E_TEST_AUTHENTICATE_USER_PASSWORD="test_Password123"
 
 # [[LOGGER]]
+NEXT_PRIVATE_LOGGER_HONEY_BADGER_API_KEY=
```
dguyen: The order of this file is wrong `NEXT_PRIVATE_LOGGER_HONEY_BADGER_API_KEY` doesn't exist anymore and `NEXT_PRIVATE_LOGGER_FILE_PATH` should be under the `# [[LOGGER]]`

apps/remix/app/routes/_authenticated+/o.$orgUrl.support.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { Button } from '@documenso/ui/primitives/button';
+
+import { SupportTicketForm } from '~/components/forms/support-ticket-form';
+import { appMetaTags } from '~/utils/meta';
+
+export function meta() {
+  return appMetaTags('Support');
+}
+
+export default function SupportPage() {
```
dguyen, line 19: This page should probably be hidden if billing is not enabled (and hidden from the menu as well)

apps/remix/app/routes/_authenticated+/o.$orgUrl.support.tsx
```diff
@@ -0,0 +1,135 @@
...
+                  </Link>
+                </h2>
+
+                <p className="text-muted-foreground mt-1">
+                  <Trans>
+                    Send an email to{' '}
+                    <a href="mailto:support@documenso.com">support@documenso.com</a>.
+                  </Trans>
+                </p>
+              </div> */}
```
dguyen: Can this be cleaned up?

apps/remix/app/routes/_authenticated+/o.$orgUrl.support.tsx
```diff
@@ -0,0 +1,135 @@
...
+              <Trans>
+                Join our community on{' '}
+                <a href="https://documen.so/discord" target="_blank" rel="noopener noreferrer">
+                  Discord
+                </a>{' '}
+                for community support and discussion.
+              </Trans>
+            </p>
+          </div>
+          {currentOrganisation && subscriptionStatus === SubscriptionStatus.ACTIVE && (
```
dguyen: I think we should allow this as long as they have any subscription instead of an `ACTIVE` one, since they may be wanting support due to a billing issue

apps/remix/server/router.ts
```diff
@@ -82,6 +83,9 @@ app.route('/api/auth', auth);
 // Files route.
 app.route('/api/files', filesRoute);
 
+// Support route
+app.route('/api/support', supportRoute);
```
dguyen: Why hono route instead of TRPC? Could put it in the profile router

apps/remix/app/routes/_authenticated+/o.$orgUrl.support.tsx
dguyen: This also needs to be disabled if no plain API key is set

dguyen (changes requested):

.env.example
Mythie: This is likely due to differing git histories, should be fine with a rebase or a manual fix :)

apps/remix/app/routes/_authenticated+/o.$orgUrl.support.tsx
catalinpit: Why? This is for free users as well. They should see the Discord and Community options from what I understood.

apps/remix/app/components/general/org-menu-switcher.tsx
catalinpit: If you think that's better, okay. I need to access the org, to check the sub status. Or am I approaching it wrong?

apps/remix/app/components/forms/support-ticket-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+    resolver: zodResolver(ZSupportTicketSchema),
+    defaultValues: {
+      email: email || '',
+      subject: '',
+      message: '',
+    },
+  });
+
+  const onSubmit = async (values: TSupportTicket) => {
+    setIsLoading(true);
```
Mythie: We don't need to handle a loading state since react-hook-form has one for us in `form.formState.isSubmitting`.

apps/remix/app/routes/_authenticated+/support.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  rel="noopener noreferrer"
+                  className="hover:underline"
+                >
+                  Discord
+                </Link>{' '}
+                for community support and discussion.
+              </Trans>
+            </p>
+          </div>
+          {currentOrganisation && subscriptionStatus && (
```
Mythie: Since people self-hosting may configure plain but won't have billing enabled due to it being EE we should make the check 

`currentOrganisation && (IS_BILLING_ENABLED() && subscriptionStatus)` or whatever would make it work so if billing isn't enabled but plain is configured everything still works.

packages/trpc/server/profile-router/router.ts
```diff
@@ -91,4 +93,16 @@ export const profileRouter = router({
         requestMetadata: ctx.metadata,
       });
     }),
+
+  submitSupportTicket: authenticatedProcedure
+    .input(ZSubmitSupportTicketMutationSchema)
+    .mutation(async ({ input, ctx }) => {
+      const { email, subject, message } = input;
```
ephraimduncan: A user can enter any email and will submit a support request on behalf of another person based on this, can you add a validation to make sure that the email belongs the current user?

```js
if (email !== ctx.user.email) {
  // throw error
 }
```

packages/trpc/server/profile-router/router.ts
```diff
@@ -91,4 +93,16 @@ export const profileRouter = router({
...
+
```
ephraimduncan, line 112: Also, what happens here when the `NEXT_PRIVATE_PLAIN_API_KEY` is not set? Shouldn't there be a validation for it to prevent submission when its not set?

I was thinking maybe we can have a plain feature flag, based on the `NEXT_PRIVATE_PLAIN_API_KEY` and if its not set, nothing support related happens, submissions fail

Something like this

```js
const isPlainEnabled = () => Boolean(env('NEXT_PRIVATE_PLAIN_API_KEY'))
```

apps/remix/app/components/forms/support-ticket-form.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const form = useForm<TSupportTicket>({
+    resolver: zodResolver(ZSupportTicketSchema),
+    defaultValues: {
+      email: email || '',
+      subject: '',
+      message: '',
+    },
+  });
+
+  const isLoading = form.formState.isLoading;
```
ephraimduncan: Can you use the isLoading from the mutation for the full loading state?

```js
const { ..., isLoading } = trpc.profile.submitSupportTicket.useMutation()

// ...

const isSubmitting = isLoading || form.formState.isSubmitting;
```

This would work

apps/remix/app/routes/_authenticated+/support.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { SupportTicketForm } from '~/components/forms/support-ticket-form';
+import { appMetaTags } from '~/utils/meta';
+
+export function meta() {
+  return appMetaTags('Support');
+}
+
+export default function SupportPage() {
+  const [showForm, setShowForm] = useState(false);
+  const { user } = useSession();
```
ephraimduncan: Can this be combined as one?

ephraimduncan (dismissed):

apps/remix/app/routes/_authenticated+/support.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const { organisations } = useSession();
```
ephraimduncan: Can this be combined as one?

packages/trpc/server/profile-router/router.ts
catalinpit: If there's no Plain API key, the Plain client throws an error, so submission fails anyway.

packages/trpc/server/profile-router/router.ts
catalinpit: I didn't add that because the organization/team member opening the ticket may not be the owner, and hence the client. If an email that is not registered as the client in Plain, won't be able to open a ticket. In an ideal world, I'd use the set team email address - but not all teams have an email set.

dguyen: @catalinpit can you verify that when an email is sent this way, it shows up as 'unverified email' on plain?

catalinpit: > @catalinpit can you verify that when an email is sent this way, it shows up as 'unverified email' on plain?

I didn't get it; why should it show `unverified email`?

dguyen: @catalinpit because we want to ensure that we know whether the person who sent the email actually sent it.

For example, what happens if I put in someone else's email and request for the account to be deleted. If the email is flagged as "unverified" then we know not to act on it without reverification.

Mythie (approved):

PR #1854, opened 2025-06-25
fix: isAssistantMode was incorrectly set to true for regular recipients

ephraimduncan:


Mythie:
@ephraimduncan Please add a description for what this fixes, if it was reported by a customer just add generic information and a reproduction on your side.

PR #1884, opened 2025-07-08
fix: handle unauthorized document move error

catalinpit:
## Description

- Adds error handling for unauthorized document moving in the `DocumentMoveToFolderDialog` component.
- Displays a toast notification when the recipient tries to move a document. 
- Updates the `DocumentsTableActionDropdown` to disable the move option if the user is not the owner of the document.
- Ensures server-side validation checks for user permissions before moving documents.

## Checklist

### DIFF
--- a/packages/lib/server-only/folder/move-document-to-folder.ts
+++ b/packages/lib/server-only/folder/move-document-to-folder.ts
@@ -54,6 +54,12 @@ export const moveDocumentToFolder = async ({
     where: documentWhereClause,
   });
 
+  if (document?.userId !== userId || document?.teamId !== teamId) {
+    throw new AppError(AppErrorCode.UNAUTHORIZED, {
+      message: 'You are not allowed to move this document',
+    });
+  }
+
   if (!document) {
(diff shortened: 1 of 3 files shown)


catalinpit:
> I think team members should be able to move documents within a team if:
> 
> - They have the visibility access for it
> 
> - They are part of the team that owns the document

They still shouldn't be able to move the document. Unless I misunderstood something here https://github.com/documenso/backlog-internal/issues/344

dguyen:
They should be able to move the document, since the document belongs to the team in general, not the user

The issue is that the recipient should not be able to move the document as Lucas said in the issue

packages/lib/server-only/folder/move-document-to-folder.ts
```diff
@@ -54,6 +54,12 @@ export const moveDocumentToFolder = async ({
     where: documentWhereClause,
   });
 
+  if (document?.userId !== userId && document?.teamId !== teamId) {
```
dguyen: Wouldn't this prevent members from not being able to move the document?

dguyen (approved):

PR #1888, opened 2025-07-10
fix: resend document api v1 filtering logic

ephraimduncan:
### Description

The resend document API was not working correctly when filtering recipients. The query was filtering recipients at the database level, which could result in an empty recipients array being returned even when the document had recipients. This prevented the API from properly identifying which recipients needed reminder emails.


dguyen:
@ephraimduncan could you create a scenario where this made a difference so i can test it out?

ephraimduncan:
@dguyen 

When the document is pending, the v1 api for resending a document fails, currently on main, regardless of the recipient status whether its opened or waiting or uncompleted from the ui.

This is the curl command I used

```bash
curl http://localhost:3000/api/v1/documents/7/resend \
  --request POST \
  --header 'Authorization: api_otcehpbfaexrxcg4' \
  --header 'Content-Type: application/json' \
  --data '{
  "recipients": [
    1
  ]
}' | jq
```

Mythie (approved):

PR #1899, opened 2025-07-14
docs: api rate limit

ephraimduncan:
### Description

Add docs for API Rate Limits

### DIFF
--- a/apps/documentation/pages/developers/public-api/rate-limits.mdx
+++ b/apps/documentation/pages/developers/public-api/rate-limits.mdx
@@ -22,21 +22,6 @@ Documenso enforces rate limits on all API endpoints to ensure service stability.
   seconds before retrying.
 </Callout>
 
-### Retry Example
-
-```javascript
-async function makeAPIRequest(url, options, retries = 3) {
-  const response = await fetch(url, options);
-
-  if (response.status === 429 && retries > 0) {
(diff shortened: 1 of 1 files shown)


apps/documentation/pages/developers/public-api/rate-limits.mdx
```diff
@@ -0,0 +1,93 @@
...
+  const response = await fetch(url, options);
+
+  if (response.status === 429 && retries > 0) {
+    await new Promise((resolve) => setTimeout(resolve, 60000)); // Wait 60s
+    return makeAPIRequest(url, options, retries - 1);
+  }
+
+  return response;
+}
+```
```
dguyen: I think this section is not required

apps/documentation/pages/developers/public-api/rate-limits.mdx
```diff
@@ -0,0 +1,93 @@
...
+    "recipients": 8,
+    "directTemplates": 1
+  },
+  "quota": {
+    "documents": 5,
+    "recipients": 10,
+    "directTemplates": 3
+  }
+}
+```
```
dguyen: Same with this section, since API is supposed to be paid only, which in that case would mean they never need to worry about limits

ephraimduncan: okay, will look at removing it

dguyen (approved):

PR #1916, opened 2025-07-21
feat: include audit trail log in the completed doc

catalinpit:
## Description

This change allows users to include the audit trail log in the completed documents; similar to the signing certificate.

https://github.com/user-attachments/assets/d9ae236a-2584-4ad6-b7bc-27b3eb8c74d3

It also solves the issue with the text cutoff.

Screenshot 2025-07-21 at 13 29 20

## Checklist


packages/prisma/migrations/20250721085125_include_audit_trail_log/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
+ALTER TABLE "OrganisationGlobalSettings" ADD COLUMN     "includeAuditLog" BOOLEAN NOT NULL DEFAULT true;
```
dguyen: Default should be false I think

packages/lib/utils/organisations.ts
```diff
@@ -115,6 +115,7 @@ export const generateDefaultOrganisationSettings = (): Omit<
     documentLanguage: 'en',
     includeSenderDetails: true,
     includeSigningCertificate: true,
+    includeAuditLog: true,
```
dguyen: Default to false I think

packages/prisma/schema.prisma
```diff
@@ -731,6 +731,7 @@ model OrganisationGlobalSettings {
   documentLanguage          String             @default("en")
   includeSenderDetails      Boolean            @default(true)
   includeSigningCertificate Boolean            @default(true)
+  includeAuditLog           Boolean            @default(true)
```
dguyen: Here as well

packages/lib/jobs/definitions/internal/seal-document.handler.ts
```diff
@@ -148,6 +149,13 @@ export const run = async ({
       }).catch(() => null)
     : null;
 
+  const auditLogData = settings.includeAuditLog
+    ? await getAuditLogPdf({
+        documentId,
+        language: document.documentMeta?.language,
+      }).catch(() => null)
```
dguyen: Could you also pull the logging we have in the `seal-document.ts` file into here as well. Same for the above function.

PR #1923, opened 2025-07-25
feat: implement recipients autosuggestions

catalinpit:
## Description

This pull request introduces recipient autosuggestions for the email and name fields in the document editor, when creating or editing documents. It allows users to select recipients from their previous document recipients and team members.

The changes include a new `AutocompleteInput` component with full keyboard navigation (arrow keys for navigation, enter to select, escape to close).

name-field
email-field

## Checklist


Mythie:
It would be cool to grab team members in the suggestions, they way we could do it is:

- Grab at most 5 recipient suggestions at all times
- Also attempt to grab 1 team member suggestion

If a team member suggestion could be found then we will show 4 recipient suggestions and 1 team member suggestion, otherwise just show all 5 team member suggestions.

packages/lib/server-only/recipient/get-recipient-suggestions.ts
```diff
@@ -0,0 +1,109 @@
...
+    distinct: ['email'],
+    orderBy: {
+      document: {
+        createdAt: 'desc',
+      },
+    },
+    take: 5,
+  });
+
+  if (teamId) {
```
catalinpit, line 63: this if statement may be redundant

packages/trpc/server/recipient-router/router.ts
```diff
@@ -559,4 +562,30 @@ export const recipientRouter = router({
...
+   */
+  getRecipientSuggestions: authenticatedProcedure
+    .input(ZGetRecipientSuggestionsRequestSchema)
+    .output(ZGetRecipientSuggestionsResponseSchema)
+    .query(async ({ input, ctx }) => {
+      const { teamId, user } = ctx;
+      const { query } = input;
+
+      ctx.logger.info({
+        input: {
```
catalinpit: the logger may need more info

packages/trpc/server/recipient-router/find-recipient-suggestions.ts
```diff
@@ -0,0 +1,34 @@
...
+    const { teamId, user } = ctx;
+    const { query } = input;
+
+    ctx.logger.info({
+      input: {
+        query,
+      },
+    });
+
+    const suggestions = await getRecipientSuggestions({
```
catalinpit, line 25: should the code from the `get-recipient-suggestions` server file be here or how it is now with the code split? trying to understand the new way of `tRPC`-ing.

packages/trpc/server/recipient-router/find-recipient-suggestions.ts
```diff
@@ -0,0 +1,34 @@
...
+export const findRecipientSuggestionsRoute = authenticatedProcedure
+  .input(ZGetRecipientSuggestionsRequestSchema)
+  .output(ZGetRecipientSuggestionsResponseSchema)
+  .query(async ({ input, ctx }) => {
+    const { teamId, user } = ctx;
+    const { query } = input;
+
+    ctx.logger.info({
+      input: {
+        query,
```
catalinpit, line 21: more logger info?

Mythie (approved):

PR #1929, opened 2025-07-28
fix: default pagination on documents list API

samuel-cglg:
## Description

When requesting a list of documents via GET /documents endpoint without `perPage` parameter, the API sends back only one document per page.

The default pagination parameters are conflicted between the schema definition (1 document per page declared) vs the implementation (10 documents per page).

Endpoint query schema (`packages/api/v1/schema.ts`):
```ts
export const ZGetDocumentsQuerySchema = z.object({
  page: z.coerce.number().min(1).optional().default(1),
  perPage: z.coerce.number().min(1).optional().default(1),
});
```

Enpoint implementation (`packages/api/v1/implementation.ts`):
```ts
  getDocuments: authenticatedMiddleware(async (args, user, team) => {
    const page = Number(args.query.page) || 1;
    const perPage = Number(args.query.perPage) || 10;
```

See the perPage default value difference between schema (1) and implementation (10).

## Related Issue

none

## Changes Made

Changed the schema definition to match its implementation (10 documents per page by default).

## Testing Performed

Tested in local

## Checklist


samuel-cglg:
Just added the same change for the get templates endpoint as it has the same issue.

ephraimduncan (approved):
Thanks for the fix

Mythie (approved):

PR #1937, opened 2025-08-04
feat: org insights

ephraimduncan:


ElTimuro:
@ephraimduncan pretty neat so far. Here are my points:

- No need for period vs alltime. Just Volume for the selcted time frame is enough
CleanShot 2025-08-19 at 16 48 12@2x
- "Volume" needs to be clarified, what is it? created docs? signed docs?
- 

I added more things here, but I think we need to get this live so I can get a feel if it helps and what would be really interesting before I keep asking for more: https://github.com/documenso/backlog-internal/issues/441

PR #1943, opened 2025-08-06
feat: reset user 2fa from admin panel

catalinpit:
## Description

Allow admins to disable the two factor authentication from the admin panel.

## Testing Performed

Tested it locally.

## Checklist


packages/trpc/server/admin-router/router.ts
```diff
@@ -51,6 +52,9 @@ export const adminRouter = router({
   stripe: {
     createCustomer: createStripeCustomerRoute,
   },
+  user: {
+    reset2FA: reset2FARoute,
```
catalinpit: don't like the `2FA` part in these names. may need to update the names to something better

apps/remix/app/routes/_authenticated+/admin+/users.$id.tsx
```diff
@@ -219,10 +220,11 @@ const AdminUserPage = ({ user }: { user: User }) => {
         />
       </div>
 
-      <div className="mt-16 flex flex-col items-center gap-4">
```
catalinpit, line 222: `items-center` caused the cards to have uneven width if one of the cards had less text

PR #1944, opened 2025-08-06
fix: send signing request email after the document status is updated

samuel-cglg:
## Description

When sending a document for signing, emails for recipients are sent before the document status is updated.
In this case, the job "send.signing.requested.email" fails because it cannot find the document with a PENDING status.
In our self hosted instance, it generates ~10 errors in the logs (from this background job) every time a document is sent (we are using a `LocalJobProvider`, not inngest for background tasks).

## Related Issue

none

## Changes Made

When sending a document, the background job "send.signing.requested.email" is created after the document status is set to PENDING.

## Testing Performed

Tested in local: the document is created as usual, the email notification is send and no error are found in the logs.

## Checklist


catalinpit:
Good catch. Thanks for the PR, we'll review it asap!

Mythie (approved):

Mythie:
Thanks for that!

PR #1946, opened 2025-08-07
feat: add organisation sso portal

dguyen:
## Description

Allow organisations to manage an SSO OIDC compliant portal. This method is intended to streamline the onboarding process and paves the way to allow organisations to manage their members in a more strict way.

**Example:**
- Organisation owner configures portal
- User signs in via the portal
- They are automatically added to the organisation

**Security**
Since we are offloading authentication to an unknown provider, we use additional email confirmation to determine whether the user wants to link their account to the organisation.

### Organisation login form

image

### Organisation SSO form

image

### Email 

Text changes based on Link or Create flows

image


Mythie (approved):

daKmoR:
hey 👋

should this be available in the cloud version as well - or is this for self hoster only?

catalinpit:
> hey 👋
> 
> should this be available in the cloud version as well - or is this for self hoster only?

Hi Thomas, SSO is an enterprise license feature for both hosted and self-hosted.

PR #1963, opened 2025-08-13
fix: prevent document unsigning on edit

dguyen:
## Description

This fixes the issue where if a partially signed document goes through the authoring process, the document becomes "unsigned" even if no recipients/fields were changed.

### DIFF
--- a/packages/lib/server-only/recipient/set-document-recipients.ts
+++ b/packages/lib/server-only/recipient/set-document-recipients.ts
@@ -134,6 +134,9 @@ export const setDocumentRecipients = async ({
         existingRecipient.id === recipient.id || existingRecipient.email === recipient.email,
     );
 
+    const canPersistedRecipientBeModified =
+      existing && canRecipientBeModified(existing, document.fields);
+
     if (
       existing &&
       hasRecipientBeenChanged(existing, recipient) &&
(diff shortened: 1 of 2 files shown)


jkbudde:
@dguyen wouldn't this create a compliance concern if the document could be modified after a signature has already been captured without first requiring it to be re-signed? _This makes sense if nothing in the document itself was actually modified, but that may be an extra complexity._

Every other signing platform that I have used requires re-signing when a document is edited. For example, [PandaDoc](https://support.pandadoc.com/en/articles/9714684-edit-sent-documents) clears signature and initials fields when editing an already sent document and forces the document to be re-sent.

dguyen:
@jkbudde currently the document owner can only modify a recipient and their fields if they have not touched the document yet, which is compliant based on what I've been told.

This commit is resolving the issue where the user goes into the editor which triggers an unsigning even though nothing happened.

jkbudde:
@dguyen that appeared to be the case based on the source 👍🏻

Just flagged this because I’ve burned hours explaining signature authenticity stuff to non-tech folks before. Didn’t mean to imply that its not compliant, just know this specific detail could make business/legal teams do a double-take and wanted to clarify. Looks good to me!

PR #1967, opened 2025-08-15
feat: add API support for folders

catalinpit:
## Description

Add API support for folders.

## Testing Performed

Tested the endpoints locally.

## Checklist

### DIFF
--- a/packages/trpc/server/folder-router/router.ts
+++ b/packages/trpc/server/folder-router/router.ts
@@ -14,15 +14,19 @@ import { FolderType } from '@documenso/lib/types/folder-type';
 
 import { authenticatedProcedure, router } from '../trpc';
 import {
+  ZCreateFolderResponseSchema,
   ZCreateFolderSchema,
   ZDeleteFolderSchema,
   ZFindFoldersRequestSchema,
   ZFindFoldersResponseSchema,
+  ZFolderSchema,
   ZGenericSuccessResponse,
   ZGetFoldersResponseSchema,
   ZGetFoldersSchema,
+  ZMoveDocumentToFolderResponseSchema,
   ZMoveDocumentToFolderSchema,
   ZMoveFolderSchema,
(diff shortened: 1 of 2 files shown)


dguyen:
Let me do a final review before this goes in lucas

Yash4898:
@ElTimuro This is Yash from Shree Mundade's Team. We require the same. It would be helpful as for each user at our platform need has multiple documents. And fetching them at once is difficult.

catalinpit:
> Let me do a final review before this goes in lucas

👀

dguyen (changes requested):
1. Remove the "Move document" and "Move template" endpoints and instead allow setting the "folderId"  in the update document and templates endpoints

2. Remove the "Move Folder" endpoint and update the "update folder" endpoint to support parentId

3. Restrict the outputs for all new endpoints, (example return void for deleteFolder, don't blindly return full objects using ZFolderSchema).

Reasoning:
- Reduce area of attack
- Remove redundant functions 

Please ensure you update the functions accordingly so it checks permissions correctly.

catalinpit:
> 1. Remove the "Move document" and "Move template" endpoints and instead allow setting the "folderId"  in the update document and templates endpoints
> 2. Remove the "Move Folder" endpoint and update the "update folder" endpoint to support parentId
> 3. Restrict the outputs for all new endpoints, (example return void for deleteFolder, don't blindly return full objects using ZFolderSchema).
> 
> Reasoning:
> 
> * Reduce area of attack
> * Remove redundant functions
> 
> Please ensure you update the functions accordingly so it checks permissions correctly.

Thanks for the feedback! Will do.

Regarding the `deleteFolder`, I didn't get it. It uses the `ZSuccessResponseSchema`, no?

```tsx
export const ZSuccessResponseSchema = z.object({
  success: z.boolean(),
  type: ZFolderTypeSchema.optional(),
});
```

This isn't ok and should return void?

PR #1986, opened 2025-08-23
feat: change direction of radio field

ephraimduncan:
## Demo

Screenshot 2025-08-23 at 4 39 08 AM


catalinpit (commented):
[radio-field.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/user-attachments/video/1d9c450d-7f64-47d3-bc93-51d7bf38430f.mov)

Nice work! Though, the "horizontal" radio field keeps changing direction as you move it around. Did that happen to you too?

ephraimduncan:
@catalinpit I double checked and nothing so far

Novapixel1010:
@catalinpit I have the same issue on version v1.12.9 but with check marks. I also took a video 😆

PR #1988, opened 2025-08-23
fix: add maxLength limits to document input fields

ephraimduncan:


Mythie (changes requested):
If we're doing this we should cover the template inputs of the same kind and also enforce it on the router input schemas so a custom request can't be crafted.

PR #1998, opened 2025-08-26
feat: signin with microsoft

ephraimduncan:
Sign in with Microsoft OAuth

### DIFF
--- a/apps/remix/app/components/forms/signup.tsx
+++ b/apps/remix/app/components/forms/signup.tsx
@@ -6,7 +6,7 @@ import { msg } from '@lingui/core/macro';
 import { useLingui } from '@lingui/react';
 import { Trans } from '@lingui/react/macro';
 import { useForm } from 'react-hook-form';
-import { FaIdCardClip } from 'react-icons/fa6';
+import { FaIdCardClip, FaMicrosoft } from 'react-icons/fa6';
 import { FcGoogle } from 'react-icons/fc';
 import { Link, useNavigate, useSearchParams } from 'react-router';
 import { z } from 'zod';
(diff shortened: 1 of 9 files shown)


packages/lib/constants/auth.ts
```diff
@@ -6,13 +6,18 @@ export const SALT_ROUNDS = 12;
 export const IDENTITY_PROVIDER_NAME: Record<string, string> = {
   DOCUMENSO: 'Documenso',
   GOOGLE: 'Google',
+  MICROSOFT: 'Microsoft',
   OIDC: 'OIDC',
 };
```
ephraimduncan, line 11: This isn't used anywhere in the application, should it be removed? or there's a plan for it?

apps/remix/app/components/forms/signup.tsx
```diff
@@ -330,6 +346,22 @@ export const SignUpForm = ({
...
+                <>
+                  <Button
+                    type="button"
+                    size="lg"
+                    variant={'outline'}
+                    className="bg-background text-muted-foreground border"
+                    disabled={isSubmitting}
+                    onClick={onSignUpWithMicrosoftClick}
+                  >
+                    <FaMicrosoft className="mr-2 h-5 w-5" />
```
rwagoner: Not sure if it's intentional, but only signin was changed to the svg color logo. Signup is still FaMicrosoft.

rwagoner:
I'm was going to submit a pull request for the Microsoft Entra SSO patch I wrote carahsoft/documenso@47e6a9702450650eb87a83ad88141ecd9d47d065 before I saw this one exists. A few notes:

- The environment variables NEXT_PRIVATE_MICROSOFT_CLIENT_ID and NEXT_PRIVATE_MICROSOFT_CLIENT_SECRET should be included in .env.example and packages/tsconfig/process-env.d.ts.
- Please add NEXT_PRIVATE_MICROSOFT_TENANT_ID="common" to substitue into the well known URL. This allows the auth to be used for a specific tenant for on-premise deployments.

Mythie (approved):

PR #2006, opened 2025-08-28
feat: unlink documents from deleted organization

catalinpit:
## Description

This change unlinks the documents from deleted organizations, and moves the documents to the organization owner if it exists. If not, it moves them to the service account. The reason for this is to retain documents for recipients.

## Checklist

### DIFF
--- a/packages/lib/server-only/envelope/orphan-envelopes.ts
+++ b/packages/lib/server-only/envelope/orphan-envelopes.ts
@@ -40,7 +40,6 @@ export const orphanEnvelopes = async ({ teamId }: OrphanEnvelopesOptions) => {
     data: {
       userId: serviceAccount.id,
       teamId: serviceAccount.ownedOrganisations[0].teams[0].id,
-      deletedAt: new Date(),
     },
   });
 


packages/trpc/server/organisation-router/delete-organisation.ts
```diff
@@ -37,6 +59,67 @@ export const deleteOrganisationRoute = authenticatedProcedure
...
+        where: {
+          id: {
+            in: documentIds,
+          },
+          status: DocumentStatus.DRAFT,
+        },
+      });
+
+      if (organisationOwner && organisationOwner.ownedOrganisations.length > 0) {
+        const ownerPersonalTeam = organisationOwner.ownedOrganisations[0].teams[0];
```
catalinpit: afaik, users will always have a personal organisation with a team, right?

catalinpit (commented):
.

packages/lib/server-only/document/handle-document-ownership-on-deletion.ts
```diff
@@ -0,0 +1,78 @@
...
+          in: documentIds,
+        },
+        status: {
+          not: DocumentStatus.DRAFT,
+        },
+      },
+      data: {
+        userId: organisationOwner.id,
+        teamId: ownerPersonalTeam.id,
+        deletedAt: new Date(),
```
catalinpit: does it need the `deletedAt` property if it's moved to the org owner?

packages/lib/server-only/envelope/orphan-envelopes.ts
```diff
@@ -0,0 +1,53 @@
...
+      status: {
+        in: [DocumentStatus.PENDING, DocumentStatus.REJECTED, DocumentStatus.COMPLETED],
+      },
+    },
+    data: {
+      userId: serviceAccount.id,
+      teamId: serviceAccount.ownedOrganisations[0].teams[0].id,
+      deletedAt: new Date(),
+    },
+  });
```
dguyen, line 44: I'm not sure if I'm missing something but wouldn't just having the second operation be enough?

packages/trpc/server/team-router/delete-team.ts
```diff
@@ -8,15 +14,36 @@ export const deleteTeamRoute = authenticatedProcedure
...
+    }
+
     ctx.logger.info({
       input: {
         teamId,
       },
     });
 
+    const transferTeam = transferTeamId
+      ? await getTeamById({ userId: user.id, teamId: transferTeamId }).catch(() => undefined)
```
dguyen: Feels like we should throw an error instead of just orphaning the items

Mythie: Yeah probably

packages/lib/server-only/envelope/orphan-envelopes.ts
Mythie: Op, the second one isn't meant to have the deletedAt set in data

PR #2068, opened 2025-10-08
fix: organisation invite member should be case insensitive

rwagoner:
## Description

The organisation invite member is case sensitive when performing the lookup. Email addresses are not case sensitive so the lookup should not be case sensitive.

## Changes Made

- Added the case insensitive mode to the prisma query

## Testing Performed

- Verified the organisation invite member is now case insensitive.

## Checklist

- [X ] I have tested these changes locally and they work as expected.
- [X ] I have followed the project's coding style guidelines.


dguyen (approved):
Thanks!

PR #2082, opened 2025-10-19
feat: add admin organisation creation with user

ephraimduncan:
## Demo

https://github.com/user-attachments/assets/538a86b3-54a9-4dce-bc99-9b103080ca47


dguyen:
@ephraimduncan I think we decided that we won't do a batch create user + create organisation at the same time

This feature should be limited to just creating a user, from then the admin can just create the organisation using the existing admin organisation create flow

dguyen (approved):

PR #2084, opened 2025-10-20
chore: add pre-commit hook to block .po file commits

ephraimduncan:
### Description

Added a new script `scripts/check-po-files.cjs` before each commit and block staged `.po` files from being committed.

### Why?

Sometimes, running build locally modifies the `.po` files and its possible to accidentally commit it and it will cause a ton of conflicts in the future.


mKoonrad:
Hi @ephraimduncan The change to block `.po` files is a good step. I just wanted to check on the status of the rest of the translation workflow that was recently broken.

1. Are the .po files being auto-generated after new commits?
2. Is the Crowdin sync (pushing and pulling files) fully automated now?

PR #2089, opened 2025-10-23
chore: add code styleguide

Mythie:


CODE_STYLE.md
```diff
@@ -0,0 +1,690 @@
...
+});
+```
+
+### Error Handling in TRPC
+
+```typescript
+// ✅ Catch and transform errors appropriately
+try {
+  const result = await createDocument({ userId, data });
+  return result;
```
catalinpit, line 628: blank line before `return result` 😈

Mythie: Claude will pay for this

CODE_STYLE.md
```diff
@@ -0,0 +1,691 @@
...
+  perPage = 10,
+}: FindDocumentsOptions) => {
+  // ...
+};
+
+// ✅ Destructure on separate line when needed
+const onFormSubmit = form.handleSubmit(onSubmit);
+
+// ✅ Deconstruct nested properties explicitly
+const { teamId } = ctx;
```
ephraimduncan: `teamId` isn't available in context, won't be a good example 

```tsx
const { user } = ctx
```

This will be a better example

CODE_STYLE.md
```diff
@@ -0,0 +1,691 @@
...
+### Error Handling in TRPC
+
+```typescript
+// ✅ Catch and transform errors appropriately
+try {
+  const result = await createDocument({ userId, data });
+
+  return result;
+} catch (err) {
+  return AppError.toRestAPIError(err);
```
ephraimduncan, line 630: Shows using `AppError.toRestAPIError(err)` in tRPC routes, but this method is for REST APIs only

```ts
// actual trpc pattern
  } catch (error) {
    if (error instanceof AppError) {
      throw error;  // Re-throw AppError
    }
    throw new AppError(AppErrorCode.UNKNOWN_ERROR, {
      message: 'Operation failed',
    });
  }
```

ephraimduncan (changes requested):

ephraimduncan (approved):

PR #2093, opened 2025-10-24
feat: admin member role updates

ephraimduncan:
## Demo

https://github.com/user-attachments/assets/dce1d690-1687-457a-8038-551a96c0741f

### DIFF
--- a/packages/app-tests/e2e/admin/organisations/update-organisation-member-role.spec.ts
+++ b/packages/app-tests/e2e/admin/organisations/update-organisation-member-role.spec.ts
@@ -68,15 +68,29 @@ test('[ADMIN]: promote member to owner', async ({ page }) => {
   // Test promoting a MEMBER to owner
   const memberRow = page.getByRole('row', { name: memberUser.email });
 
-  // Find and click the "Promote to owner" button for the member
-  const promoteButton = memberRow.getByRole('button', { name: 'Promote to owner' });
-  await expect(promoteButton).toBeVisible();
-  await expect(promoteButton).not.toBeDisabled();
+  // Find and click the "Update role" button for the member
+  const updateRoleButton = memberRow.getByRole('button', {
+    name: 'Update role',
+  });
+  await expect(updateRoleButton).toBeVisible();
+  await expect(updateRoleButton).not.toBeDisabled();
 
-  await promoteButton.click();
+  await updateRoleButton.click();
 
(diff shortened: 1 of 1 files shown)


apps/remix/app/components/dialogs/admin-organisation-member-update-dialog.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+  const onFormSubmit = async ({ role }: ZUpdateOrganisationMemberSchema) => {
+    try {
+      await updateOrganisationMemberRole({
+        organisationId,
+        userId: organisationMemberUserId,
+        role,
+      });
+
+      const roleLabel =
```
Mythie: use a `match(...)` call

apps/remix/app/components/dialogs/admin-organisation-member-update-dialog.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+type ZUpdateOrganisationMemberSchema = z.infer<typeof ZUpdateOrganisationMemberFormSchema>;
+
+export const AdminOrganisationMemberUpdateDialog = ({
+  trigger,
+  organisationId,
+  organisationMemberId,
+  organisationMemberName,
+  organisationMemberUserId,
+  organisationMemberRole,
```
Mythie: Looks like you're just passing in the whole `organisationMember` record or a subset of it, probably better to make the prop `organisationMember` and use the type from Prisma or a subset of it with `Pick<...>`

ephraimduncan: will do!

packages/trpc/server/admin-router/update-organisation-member-role.ts
```diff
@@ -0,0 +1,218 @@
...
+  });
+
+  await tx.organisationGroupMember.create({
+    data: {
+      id: generateDatabaseId('group_member'),
+      organisationMemberId: memberId,
+      groupId: toGroupId,
+    },
+  });
+};
```
Mythie: This would be the first instance of us taking a transaction as a parameter which feels kinda wrong, also the code is only used within this file so I'd probably just duplicate the code or consider how I can reorder my logic.

ephraimduncan: fixed

PR #2111, opened 2025-10-28
feat: autoplace fields from placeholders

catalinpit:
## Description

This PR introduces automatic detection and placement of fields and recipients based on PDF placeholders.

The placeholders have the following structure:
* `{{fieldType,recipientPosition,fieldMeta}}` - `{{text,r1,required=true,textAlign=right,fontSize=50}}`

When the user uploads a PDF document containing such placeholders, they get converted automatically to Documenso fields and assigned to recipients.

The recipients are created as follows:
- when uploading the document via the UI, Documenso creates placeholder recipients that can be edited in the document editor
- when uploading the document via the API, Documenso assigns the fields to the recipients provided in the request body.

## Checklist

### DIFF
--- a/packages/trpc/server/envelope-router/create-envelope-items.ts
+++ b/packages/trpc/server/envelope-router/create-envelope-items.ts
@@ -5,6 +5,7 @@ import {
   extractPdfPlaceholders,
 } from '@documenso/lib/server-only/pdf/auto-place-fields';
 import { findRecipientByPlaceholder } from '@documenso/lib/server-only/pdf/helpers';
+import { insertFormValuesInPdf } from '@documenso/lib/server-only/pdf/insert-form-values-in-pdf';
 import { normalizePdf } from '@documenso/lib/server-only/pdf/normalize-pdf';
 import { DOCUMENT_AUDIT_LOG_TYPE } from '@documenso/lib/types/document-audit-logs';
 import { prefixedId } from '@documenso/lib/universal/id';
@@ -93,8 +94,15 @@ export const createEnvelopeItemsRoute = authenticatedProcedure
     // For each file: normalize, extract & clean placeholders, then upload.
     const envelopeItems = await Promise.all(
       files.map(async (file) => {
-        const buffer = Buffer.from(await file.arrayBuffer());
-        const normalized = await normalizePdf(buffer);
+        let buffer = Buffer.from(await file.arrayBuffer());
+
+        if (envelope.formValues) {
+          buffer = await insertFormValuesInPdf({ pdf: buffer, formValues: envelope.formValues });
+        }
+
+        const normalized = await normalizePdf(buffer, {
+          flattenForm: envelope.type !== 'TEMPLATE',
+        });
 
         const { cleanedPdf, placeholders } = await extractPdfPlaceholders(normalized);
 


packages/lib/server-only/pdf/auto-place-fields.ts
```diff
@@ -0,0 +1,424 @@
...
+      });
+
+      resolve(placeholders);
+    });
+
+    parser.parseBuffer(pdf);
+  });
+};
+
+export const replacePlaceholdersInPDF = async (pdf: Buffer): Promise<Buffer> => {
```
catalinpit: this file needs a better organization. maybe this function shouldn't even be here

packages/lib/server-only/pdf/auto-place-fields.ts
```diff
@@ -0,0 +1,424 @@
...
+    recipientIndex: Number(indexMatch[1]),
+  };
+};
+
+/*
+  Parse field type string to FieldType enum.
+  Normalizes the input (uppercase, trim) and validates it's a valid field type.
+  This ensures we handle case variations and whitespace, and provides clear error messages.
+*/
+const parseFieldType = (fieldTypeString: string): FieldType => {
```
catalinpit: move helpers at the top or in a separate file

packages/lib/server-only/pdf/auto-place-fields.ts
```diff
@@ -0,0 +1,517 @@
...
+  const recipientIndex = Number(indexMatch[1]);
+
+  return {
+    email: `recipient.${recipientIndex}@documenso.com`,
+    name: `Recipient ${recipientIndex}`,
+    recipientIndex,
+  };
+};
+
+export const insertFieldsFromPlaceholdersInPDF = async (
```
catalinpit: should this function be outside this file? in the `fields` folder maybe?

packages/lib/server-only/pdf/normalize-pdf.ts
```diff
@@ -13,6 +14,7 @@ export const normalizePdf = async (pdf: Buffer) => {
   removeOptionalContentGroups(pdfDoc);
   await flattenForm(pdfDoc);
   flattenAnnotations(pdfDoc);
+  const pdfWithoutPlaceholders = await removePlaceholdersFromPDF(pdf);
```
catalinpit: Initially added it here just for testing purposes, but I have a feeling this may not be the right place for this. I believe only a small number of the uploaded PDFs would contain placeholders. So, useless processing for those that don't.

packages/lib/server-only/pdf/auto-place-fields.ts
```diff
@@ -0,0 +1,387 @@
...
+  recipientId: number;
+  pageNumber: number;
+  pageX: number;
+  pageY: number;
+  width: number;
+  height: number;
+};
+
+/*
+  Questions for later:
```
catalinpit: remove these

packages/lib/server-only/envelope/create-envelope.ts
```diff
@@ -390,4 +395,51 @@ export const createEnvelope = async ({
 
     return createdEnvelope;
   });
+
+  for (const envelopeItem of createdEnvelope.envelopeItems) {
```
catalinpit: Maybe move the logic somewhere else.

PR #2170, opened 2025-11-11
fix: remove redundant endpoint

dguyen:
Remove duplicate redundant endpoint


Mythie (approved):

mKoonrad:
@dguyen Is there any problem with E2E tests right now? I found that your PR also fails the Playwright tests. My recent PRs also fails. It doesn't seem correct that tests return different results for the same code. I had situations where code was correct, and only after running tests few times (with an empty commit) did they pass.

dguyen:
@mKoonrad don't worry about the playwright tests, there's some issues with it where it's randomly flakey.

mKoonrad:
@dguyen Thanks! Meanwhile, I discovered that tests are more likely to succeed when GitHub Action has no any tasks in progress. It may be related to performance. Maybe `waitForTimeout` function in E2E files isn't good with slower servers.

PR #2181, opened 2025-11-12
fix: add default values for envelope field meta

dguyen:
🙏


packages/trpc/server/document-router/router.ts
```diff
@@ -38,8 +37,6 @@ export const documentRouter = router({
   search: searchDocumentRoute,
   share: shareDocumentRoute,
 
-  download: downloadDocumentRoute,
```
Mythie, line 41: Looks like an accident

dguyen: thanks

PR #2201, opened 2025-11-15
feat: add env variable for oidc scopes

Nevercold:
## Description

I noticed that where I set up OIDC, you can't change the scopes.
I then saw in the code that they are simply hardcoded.

## Changes Made

I added a variable that accepts the scopes.
The default value remains the same as before, so other systems will not be affected.

## Testing Performed

I started it locally and logged in.

## Checklist


Mythie:

https://documen.so/cla

Nevercold:
> 
> https://documen.so/cla

Done :)

Mythie:
Silly question, when would we actually want to change the scopes from the default needed to allow OIDC? The application would never use any additional scopes and the ones hardcoded are required to ensure all the attributes we want are returned when fetching a given user from the OIDC provider.

Nevercold:
> Silly question, when would we actually want to change the scopes from the default needed to allow OIDC? The application would never use any additional scopes and the ones hardcoded are required to ensure all the attributes we want are returned when fetching a given user from the OIDC provider.

For example, we have developed our own OIDC server. This does not use the standard scopes, which means that the application cannot be connected to it. I think it's important to adjust the scopes. You could consider making a note so that the sys admin knows what data documenso needs to make it work properly.

Mythie:
> > Silly question, when would we actually want to change the scopes from the default needed to allow OIDC? The application would never use any additional scopes and the ones hardcoded are required to ensure all the attributes we want are returned when fetching a given user from the OIDC provider.
> 
> For example, we have developed our own OIDC server. This does not use the standard scopes, which means that the application cannot be connected to it. I think it's important to adjust the scopes. You could consider making a note so that the sys admin knows what data documenso needs to make it work properly.

I mean in the case of the sysadmin this would be the same scopes that almost every provider needs no?

It just seems a bit weird to configure our already minimal scopes ahaha

Nevercold:
> > > Silly question, when would we actually want to change the scopes from the default needed to allow OIDC? The application would never use any additional scopes and the ones hardcoded are required to ensure all the attributes we want are returned when fetching a given user from the OIDC provider.
> > 
> > 
> > For example, we have developed our own OIDC server. This does not use the standard scopes, which means that the application cannot be connected to it. I think it's important to adjust the scopes. You could consider making a note so that the sys admin knows what data documenso needs to make it work properly.
> 
> I mean in the case of the sysadmin this would be the same scopes that almost every provider needs no?
> 
> It just seems a bit weird to configure our already minimal scopes ahaha

The ones that were hardcoded before are now configurable and set as default valuess, so that the existing server does not break down.
Or am I misunderstanding something? :D

PR #2205, opened 2025-11-17
fix: add field drag drop colors

dguyen:
## Description

Update drag drop fields to be always colored, instead of only when it gets dragged/hovered.

image

image


dguyen:
Pending internal discussion on whether to add.

PR #2206, opened 2025-11-18
fix: envelope template recipient placeholders

dguyen:
## Description

- Fix missing template placeholders for envelopes
- Fix redirect user out of editor back to either documents or templates depending on their current state


Mythie:
Ideally we remove the requirement for the name/email to populated like I did for the embedding ones. This is mostly controlled on the client side.

You can use `z.union(z.string().length(0), z.string().email())` to ensure it's either an empty string or email provided.

Caveat being how we handle people using a template and not providing a final email for the document.

Mythie:
Closing to do it the right way as discussed

PR #2232, opened 2025-11-22
feat: add envelope audit logs endpoint

ephraimduncan:


dguyen (changes requested):
The `findDocumentAuditLog` stuff you're using is intended for internal usage only, you shouldn't be using `filterForRecentActivity` or `cursor` params.

Follow other examples we have of `find` endpoints, it should be standard.

This PR should probably only require these 3 files:
- packages/trpc/server/envelope-router/find-envelope-audit-logs.ts
- packages/trpc/server/envelope-router/find-envelope-audit-logs.types.ts
- packages/trpc/server/envelope-router/router.ts

PR #2248, opened 2025-11-26
feat: add default recipients for teams and orgs

catalinpit:
## Description

https://github.com/user-attachments/assets/b8966ce8-e539-4276-b66b-d31eef8a86b1

LE: This is how the UI looks now

Screenshot 2025-12-02 at 14 27 11

## Checklist


packages/lib/server-only/envelope/create-envelope.ts
```diff
@@ -306,8 +307,22 @@ export const createEnvelope = async ({
 
     const firstEnvelopeItem = envelope.envelopeItems[0];
 
+    const defaultRecipients = settings.defaultRecipients
+      ? ZDefaultRecipientsSchema.parse(settings.defaultRecipients)
+      : [];
+
+    const mappedDefaultRecipients: CreateEnvelopeRecipientOptions[] = defaultRecipients.map(
```
catalinpit, line 353: do this so typescript stops throwing errors

packages/lib/utils/recipient-formatter.ts
```diff
@@ -7,6 +7,6 @@ export const extractInitials = (text: string) =>
     .slice(0, 2)
     .join('');
 
-export const recipientAbbreviation = (recipient: Recipient) => {
+export const recipientAbbreviation = (recipient: Pick<Recipient, 'name' | 'email'>) => {
```
catalinpit, line 10: this function uses only the name and email properties, so I modified it to get rid of the TS errors when you're not passing a recipient with all the properties.

Mythie:
I was originally skeptical but I like the direction this is going in.

I think we potentially need to determine whether or not recipients added via this method are locked or not (meaning they can't be removed), we can probably lock this by default and just ensure you can't remove anyone in the default list for now and then re-evaluate later on if someone requests something different.

Mythie (approved):

PR #2251, opened 2025-11-27
fix: migrate certificate generation

dguyen:
## Description

Generate certificates and audit logs using Konva instead of browserless.

This should:
- Reduce the changes of generations failing
- Improve sealing speed


dguyen:
Todo: Remember to update the send audit log since we've made some changes.

PR #2257, opened 2025-11-27
fix: sync envelope state after direct link changes

ephraimduncan:


dguyen:
@ephraimduncan was this tested? It doesn't seem to work for me

ephraimduncan:
@dguyen should work now, can you look at it again?

PR #2264, opened 2025-12-01
feat: add external ID to use template

catalinpit:
## Description

Allow the users to override the template's external ID in the API request body.

## Checklist

### DIFF
--- a/packages/lib/types/document-meta.ts
+++ b/packages/lib/types/document-meta.ts
@@ -95,11 +95,6 @@ export const ZDocumentMetaDistributionMethodSchema = z
   .nativeEnum(DocumentDistributionMethod)
   .describe('The distribution method to use when sending the document to the recipients.');
 
-export const ZDocumentMetaExternalIdSchema = z
-  .string()
-  .max(255)
-  .describe('A custom external ID you can use to identify the document.');
-
 export const ZDocumentMetaTypedSignatureEnabledSchema = z
   .boolean()
   .describe('Whether to allow recipients to sign using a typed signature.');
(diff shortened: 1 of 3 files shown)


packages/trpc/server/template-router/schema.ts
```diff
@@ -152,6 +153,7 @@ export const ZCreateDocumentFromTemplateRequestSchema = z.object({
       distributionMethod: ZDocumentMetaDistributionMethodSchema.optional(),
       emailSettings: ZDocumentEmailSettingsSchema.optional(),
       language: ZDocumentMetaLanguageSchema.optional(),
+      externalId: ZDocumentMetaExternalIdSchema.optional(),
```
dguyen: Put this on the top level like how we did it for the envelopes/use endpoint

packages/lib/types/document-meta.ts
```diff
@@ -95,6 +95,11 @@ export const ZDocumentMetaDistributionMethodSchema = z
   .nativeEnum(DocumentDistributionMethod)
   .describe('The distribution method to use when sending the document to the recipients.');
 
+export const ZDocumentMetaExternalIdSchema = z
+  .string()
+  .max(255)
+  .describe('A custom external ID you can use to identify the document.');
```
dguyen: We already have an `ZDocumentExternalIdSchema` export in the application

dguyen (changes requested):

packages/trpc/server/template-router/schema.ts
catalinpit: Made more sense to me to be in the override part, but I moved it.

PR #2128, opened 2025-11-04
feat: generates coordinates for bounding boxes

ephraimduncan:
## Demo - Auto detect recipients and fields

https://github.com/user-attachments/assets/aa39e001-b6e9-48f0-bf7f-bb20a98a5da2

PR #2271, opened 2025-12-03
feat: add ai detection for recipients and fields

Mythie:
Use Gemini to handle detection of recipients and fields within documents.

Opt in using organisation or team settings.

Replaces #2128 since the branch was cursed and would include dependencies that weren't even in the lock file.

https://github.com/user-attachments/assets/e6cbb58f-62b9-4079-a9ae-7af5c4f2e4ec

## Concerns

- We had to use streaming responses for the requests which means the code is a little bit sloppy.
- Gemini 3 is incredibly accurate for fields but incurs a latency penalty
- We may need to request higher rate limits asap.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -70,6 +70,7 @@ COPY --from=builder /app/out/json/ .
 COPY --from=builder /app/out/package-lock.json ./package-lock.json
 
 COPY --from=builder /app/lingui.config.ts ./lingui.config.ts
+COPY --from=builder /app/patches ./patches
 
 RUN npm ci
 
@@ -108,6 +109,8 @@ WORKDIR /app
 COPY --from=builder --chown=nodejs:nodejs /app/out/json/ .
 # Copy the tailwind config files across
 COPY --from=builder --chown=nodejs:nodejs /app/out/full/packages/tailwind-config ./packages/tailwind-config
+# Copy the patches across
+COPY --from=builder --chown=nodejs:nodejs /app/patches ./patches
 
 RUN npm ci --only=production
 


Mythie:
@coderabbitai review

.env.example
```diff
@@ -147,6 +147,17 @@ NEXT_PUBLIC_USE_INTERNAL_URL_BROWSERLESS=false
...
+# [[AI]]
+# OPTIONAL: Google Cloud Project ID for Vertex AI.
+GOOGLE_VERTEX_PROJECT_ID=""
+# OPTIONAL: Google Cloud region for Vertex AI. Defaults to "global".
+GOOGLE_VERTEX_LOCATION="global"
+# OPTIONAL: API key for Google Vertex AI (Gemini). Get your key from:
+# https://console.cloud.google.com/vertex-ai/studio/settings/api-keys
+GOOGLE_VERTEX_AI_API_KEY=""
+# OPTIONAL: Enable AI field detection debug mode to save preview images with bounding boxes
+NEXT_PUBLIC_AI_FIELDS_DEBUG_PREVIEW=
```
dguyen: Nitpick

Don't think this is used? Same in turbo.json

apps/remix/app/components/forms/document-preferences-form.tsx
```diff
@@ -509,6 +513,56 @@ export const DocumentPreferencesForm = ({
             )}
           />
 
+          <FormField
```
dguyen: I think this needs to be hidden if no AI API key is detected and maybe also disable this if document is pending?

apps/remix/server/api/ai/detect-fields.ts
```diff
@@ -0,0 +1,152 @@
...
+      }).catch(() => null);
+
+      if (!team) {
+        throw new AppError(AppErrorCode.UNAUTHORIZED, {
+          message: 'You do not have access to this team',
+        });
+      }
+
+      // Check if AI features are enabled for the team
+      const aiFeaturesEnabled =
```
dguyen: Nitpick: team returns derivied settings

dguyen (approved):

apps/remix/server/api/ai/detect-fields.ts
Mythie: So it does

PR #2295, opened 2025-12-07
fix: local job retry loop for webhook calls

gps949:
## Description

  This PR prevents the local job provider from endlessly requeuing failed webhook jobs. We now flag the initial submission as a non-retry and mark subsequent submissions as
  retries so the existing retry counter and maxRetries guard start working again.

  ## Related Issue

  Fixes the data/log explosion caused by WebhookCall insertions on repeated webhook failures when running with the local job provider.

  ## Changes Made

  - Pass isRetry: false when the job is first enqueued so the handler receives an explicit flag.
  - Pass isRetry: true when a failed job is requeued, allowing the retried counter to increment and the maxRetries protection to fire.

  ## Testing Performed

  - Not run (logic-only fix; environments with the local provider will exercise the new behavior automatically).

  ## Checklist

  ## Additional Notes

  Once this lands, consider adding a background cleanup job (or data retention policy) for WebhookCall to shrink inflated tables created before the fix.


Mythie:

https://documen.so/cla

PR #2317, opened 2025-12-12
refactor: simplify form validation and enhance recipient handling

catalinpit:
## Description

This PR closes https://github.com/documenso/backlog-internal/issues/540.

## Checklist


Mythie:
The frontend changes are correct, but it looks like you're trying to be too smart on the backend by attempting to link it to the existing recipient using that email.

You don't need to do this and probably don't want to do this either. Since an email can appear as a recipient multiple times we would simply create a new recipient with the provided email and have the fields linked to that rather than attempting an upsert to an existing recipient.

PR #2323, opened 2025-12-15
fix: assignment operator for directRecipientName

catalinpit:
## Checklist


dguyen (approved):

packages/lib/server-only/template/create-document-from-direct-template.ts
```diff
@@ -238,7 +238,7 @@ export const createDocumentFromDirectTemplate = async ({
       }
 
       if (templateField.type === FieldType.NAME && directRecipientName === undefined) {
-        directRecipientName === signedFieldValue?.value;
+        directRecipientName = signedFieldValue?.value;
```
Mythie, line 241: @copilot `directRecipientName` is a constant so we need to go up in this file and change it to `let`

PR #2322, opened 2025-12-15
feat: document file conversion

ephraimduncan:
### Description

Enable uploads of DOCX/JPEG/PNG by converting to PDF via Gotenberg while preserving the original file for download, and update UI messaging/validation and download handling accordingly.

### Changes Made

- Added upload constants/allowlist and Gotenberg config; added server-side conversion for non-PDF uploads in document/envelope creation with new conversion error handling.
- Persisted original source files and MIME types on DocumentData with Prisma migrations; updated API queries/types to expose originalMimeType/originalData.
- Updated file serving/downloading to return the original file when requested and to honor Content-Disposition filenames.
- Updated upload UI/dropzones to accept PDF/DOCX/JPEG/PNG and refreshed error copy; added Gotenberg to dev compose and new test assets; adjusted an E2E upload step.

### Checklist

- [x] I have tested these changes locally and they work as expected.
- [ ] I have added/updated tests that prove the effectiveness of these changes.
- [x] I have updated the documentation to reflect these changes, if applicable.
- [x] I have followed the project's coding style guidelines.
- [ ] I have addressed the code review feedback from the previous submission, if applicable.

### Demo - Side by Side. Document in Microsoft Word and Signed in Documenso

Screenshot 2026-01-06 at 8 28 04 PM (2)

### Demo Files

[Original Docx File](https://github.com/user-attachments/files/24459295/Documenso_Partnership_Agreement.docx)

[Docx File Signed on Documenso](https://github.com/user-attachments/files/24459296/Documenso_Partnership_Agreement_signed.pdf)

PR #2345, opened 2025-12-22
refactor: consolidate get-pdf

tedliang:
## Description

Consolidate get-audit-logs-pdf and get-certificate-pdf into get-pdf.
Make it easier to support different conversion service, eg. Gotenberg (once #2322 is merged)

## Related Issue

N/A

## Changes Made

- Extract lang cookie key.
- Remove 92 lines of duplicated code for pdf conversion with Playwright.

## Testing Performed

- Tested on browser Chrome.

## Checklist

- [X] I have added/updated tests that prove the effectiveness of these changes.
- [X] I have updated the documentation to reflect these changes, if applicable.
- [X] I have addressed the code review feedback from the previous submission, if applicable.


Mythie:
Just so we're aware https://github.com/documenso/documenso/pull/2251 will remove the dependency on Playwright/Chromium and keep it all native. We've just been a little slow to merge it

tedliang:
> Just so we're aware #2251 will remove the dependency on Playwright/Chromium and keep it all native. We've just been a little slow to merge it

Sounds good to create the pdf natively. But it's a big change. Would be better to make it switchable in the beginning. Will update this PR to accommodate the changes for #2251.

packages/lib/jobs/definitions/internal/seal-document.handler.ts
```diff
@@ -493,24 +485,13 @@ export const getCertificateAndAuditLogData = async ({
     : null;
 
   const getAuditLogDataPromise = settings.includeAuditLog
-    ? getAuditLogsPdf({
-        documentId: legacyDocumentId,
-        language: documentMeta.language,
-      }).catch((e) => {
+    ? getAuditLogPdfByPlaywright(payload).catch((e) => {
```
tedliang: for #2251, we can make it switchable like this:
```
? match(env('NEXT_PRIVATE_PDF_GENERATOR'))
      .with('playwright', async () => await getAuditLogPdfByPlaywright(payload))
      .otherwise(async () => await generateAuditLogPdf(payload)).catch((e) => {
```

PR #2396, opened 2026-01-17
fix(api): add take parameter to template search query for pagination

### LINKED ISSUE #2395: [API v2] Pagination ignored in /api/v2/templates endpoint
### Issue Description

The findTemplates service implementation ignores the perPage parameter. This causes the /api/v2/template endpoint to return all matching templates regardless of the requested limit, breaking pagination functionality

### Steps to Reproduce

Ensure the database has more than 1 template (e.g. 2).
Make a request to the V2 template endpoint requesting only 1 item per page:

```
bash
curl "http://localhost:3000/api/v2/template?page=1&perPage=1" \
  -H "Authorization: Bearer YOUR_API_KEY"
```

### Expected Behavior

The API should return a single template item in the specific page data, respecting the perPage=1 limit.

### Current Behavior

The API returns all available templates in the data array (e.g. 2 items), even though the metadata in the response correctly acknowledges perPage: 1.

```
// Request: perPage=1
// Response:
{
  "data": [
    { "id": 10, ... },
    { "id": 9, ... }  // <--- Should not be here on page 1
  ],
  "count": 2,
  "currentPage": 1,
  "perPage": 1,
  "totalPages": 2
}

### Screenshots (optional)

_No response_

MacOs

Chrome

### Version [e.g., 2.0.1]

2.3.1

- [x] I want to work on creating a PR for this issue if approved

jormigo:
## Description

This PR fixes a bug in the `/api/v2/template` endpoint where the pagination parameter `perPage` was being ignored. Previously, the endpoint would return all matching templates regardless of the requested limit, which could lead to performance issues and incorrect API behavior.

## Related Issue

Fixes #2395

## Changes Made

- Updated [packages/lib/server-only/template/find-templates.ts](cci:7://file:///Users/jorge/dev/documenso/packages/lib/server-only/template/find-templates.ts:0:0-0:0) to include `take: perPage` in the `prisma.envelope.findMany` call. This ensures the database query respects the requested page size.

## Testing Performed

I have verified the changes manually using `curl` against a local instance:

- **Before Fix**: Requesting `perPage=1` returned all 2 templates in the database.
- **After Fix**: Requesting `perPage=1` returns exactly 1 template as expected.
- Verified that `page=2` correctly returns the subsequent results.
- Validated that existing pagination for `/api/v2/envelope` and `/api/v2/document` remains unaffected.

## Checklist

## Additional Notes

The response metadata (`perPage`, `totalPages`, etc.) was already correct, but the `data` array contained more items than it should have. This change aligns the `data` length with the metadata.

### DIFF
--- a/packages/lib/server-only/template/find-templates.ts
+++ b/packages/lib/server-only/template/find-templates.ts
@@ -79,6 +79,7 @@ export const findTemplates = async ({
         },
       },
       skip: Math.max(page - 1, 0) * perPage,
+      take: perPage,
       orderBy: {
         createdAt: 'desc',
       },


Mythie:
Great catch!

PR #2037, opened 2025-09-18
feat: add option to disable OIDC login prompt=login parameter and upd…

KarloDerEchte:
…ate environment configuration

## Description

This pull request adds a 
## Changes Made

- Change 1
- Change 2
- ...

## Testing Performed

- Tested feature X in scenario Y.
- Ran unit tests for component Z.
- Tested on browsers A, B, and C.
- ...

## Checklist

- [ ] I have tested these changes locally and they work as expected.
- [ ] I have added/updated tests that prove the effectiveness of these changes.
- [ ] I have updated the documentation to reflect these changes, if applicable.
- [ ] I have followed the project's coding style guidelines.
- [ ] I have addressed the code review feedback from the previous submission, if applicable.

PR #2421, opened 2026-01-27
fix: preserve prompt parameter in OAuth authorize URL builder

Mythie:
The prompt option was being discarded for OAuth authorize URLs after
adding support for the NEXT_PRIVATE_OIDC_PROMPT env var. This meant
select_account (used elsewhere) was not being passed through.

Now defaults prompt to the provided option (or 'login'), and only
overwrites it when a valid OIDC prompt env var is set. Also adds a
type guard to validate the env var value.


KarloDerEchte:
This issue returns the bug fixed in #2037 

Using the parameter none returns in a invalid state error, using 'consent' works but as it says always prompts the user to give his consent.

Issue #2397, opened 2026-01-18
Users duplicated when OIDC signup used in app


Pravpraskumar:
### Issue Description

When app is configured with OIDC system and if users try to signup with it. emails are considered as case sensitive and accounts are duplicated with same email as one at lower case and other at mixed case.

### Steps to Reproduce

1. Configure OIDC.
2. Signup account using OIDC login
3. Observe if an existing account exists with an email with lower case.
4. Try to login with the user email which is having mixed case.
5. Observe the user is duplicating in the system.

### Expected Behavior

Always users sign up with OIDC also should be case insensitive, and no duplication should happen.

### Please check the boxes that apply to this issue report.

- [ ] I have searched the existing issues to make sure this is not a duplicate.
- [x] I have provided steps to reproduce the issue.
- [ ] I have included relevant environment information.
- [ ] I have included any relevant screenshots.
- [ ] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [ ] I want to work on creating a PR for this issue if approved

ElTimuro:
- this should not happen
- moving to prioritization

PR #2438, opened 2026-01-31
fix: normalize OIDC email to lowercase to prevent duplicate users

misha366:
## Description

The `validateOauth()` function in `handle-oauth-callback-url.ts` returns the email from OIDC JWT claims without normalizing it to lowercase. This causes duplicate user creation when the same person signs up via OIDC (with mixed-case email like Test@Example.COM) and then via email/password (with test@example.com)

## Related Issue

#2397

## Changes Made

- Normalized email to lowercase in the return value of `validateOauth()` by calling `.toLowerCase()`, consistent with how all other auth flows handle emails (e.g. `create-user.ts`, `email-password.ts`)

## Testing Performed

- Most popular OIDC providers (e.g. keycloak, google) normalize emails to lowercase before returning them in tokens, which makes this bug hard to reproduce. To reproduce the issue, a custom mock OIDC server was written that returns a mixed-case email (Test@Example.COM) without normalization.
- Video demonstrating the reproduction: https://go.screenpal.com/watch/cOVwbqn3aFE
- Verified that after the fix, signing up via mock OIDC with Test@Example.COM and then via email/password with test@example.com correctly links to the same user instead of creating a duplicate.

## Checklist

- [+] I have tested these changes locally and they work as expected.
- [+] I have followed the project's coding style guidelines.

## Additional Notes

The fix is a single-line change in `validateOauth()` which is shared by both the regular OAuth callback and the organisation OAuth callback (`handle-oauth-organisation-callback-url.ts`), so both flows are covered


Mythie:
This is a good PR, raises some internal questions on how we handle cases where a duplicate already exists. We can't cleanly migrate those and so this would need some coordinated effort.

PR #2440, opened 2026-02-02
fix: bulk actions improvements

catalinpit:
### Persist selections

https://github.com/user-attachments/assets/b6dadef4-d4ef-4fcf-bd7f-3049e7d28b3e

#### Before
	
Screenshot 2026-02-02 at 11 18 05

#### After

Screenshot 2026-02-02 at 11 19 21

### Cancel dialog

The "Cancel" button for the delete dialog now works properly by closing the dialog when clicked.

542031312-d448978b-d15c-4124-a650-b082d3a38e2d

## Checklist

### DIFF
--- a/apps/documentation/pages/developers/_meta.js
+++ b/apps/documentation/pages/developers/_meta.js
@@ -13,6 +13,7 @@ export default {
     title: 'API & Integration Guides',
   },
   'public-api': 'Public API',
-  embedding: 'Embedding',
+  embedding: 'Embedded Signing',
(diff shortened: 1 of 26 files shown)


packages/lib/client-only/hooks/use-session-storage.ts
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
```
catalinpit, line 1: hook from here https://usehooks.com/usesessionstorage

PR #2471, opened 2026-02-11
fix: return all documents from V1 API regardless of folder

ephraimduncan:
## Summary

The `GET /api/v1/documents` endpoint only returned root-level documents, silently hiding any documents placed in folders from API consumers. This also adds `folderId` to both the list and single-document response schemas.

## Problem

`findDocuments()` defaults `folderId` to `null` when no value is provided, meaning only root-level documents are returned. The V1 API never passed `folderId`, so folder documents were invisible. Neither endpoint exposed `folderId` in the response, so consumers couldn't know which folder a document belonged to.

## Changes

- **`packages/api/v1/schema.ts`** — Add optional `folderId` query param to `ZGetDocumentsQuerySchema`. Add `folderId: z.string().nullish()` to `ZSuccessfulDocumentResponseSchema` (inherited by `ZSuccessfulGetDocumentResponseSchema`).
- **`packages/api/v1/implementation.ts`** — `getDocuments` passes `folderId` and `skipFolderFilter` to `findDocuments()` and includes `folderId` in the list response. `getDocument` includes `folderId` in the single-document response.
- **`packages/lib/server-only/document/find-documents.ts`** — Add `skipFolderFilter?: boolean` to `FindDocumentsOptions`. When true, the folder WHERE clause is skipped entirely, returning all documents. Existing callers are unaffected since they don't pass this flag.

## Behavior

| Request | Before | After |
|---|---|---|
| `GET /api/v1/documents` | Root docs only | All docs (root + folders) |
| `GET /api/v1/documents?folderId=abc` | Not supported | Docs in folder `abc` only |
| `GET /api/v1/documents?folderId=nonexistent` | Not supported | Empty array, 200 OK |
| `GET /api/v1/documents/:id` response | No `folderId` field | Includes `folderId` |

## Breaking Change

`GET /api/v1/documents` now returns more documents than before. This is intentional — the previous behavior was a bug that silently hid folder documents from API consumers. The new `folderId` response field is additive and won't break existing response parsing.

### DIFF
--- a/packages/api/v1/implementation.ts
+++ b/packages/api/v1/implementation.ts
@@ -68,7 +68,6 @@ export const ApiContractV1Implementation = tsr.router(ApiContractV1, {
       userId: user.id,
       teamId: team.id,
       folderId: args.query.folderId,
-      skipFolderFilter: args.query.folderId === undefined,
     });
 
     return {
--- a/packages/api/v1/schema.ts
+++ b/packages/api/v1/schema.ts
@@ -37,9 +37,7 @@ export const ZGetDocumentsQuerySchema = z.object({
   perPage: z.coerce.number().min(1).optional().default(10),
   folderId: z
     .string()
-    .describe(
-      'Filter documents by folder ID. When omitted, returns all documents regardless of folder.',
-    )
+    .describe('Filter documents by folder ID. When omitted, returns root documents.')
     .optional(),
 });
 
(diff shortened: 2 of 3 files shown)


Mythie (changes requested):
Don't skip folderId, just keep the query arg as you have already.

Mythie (approved):

PR #2476, opened 2026-02-11
fix: guard send dialogs against org claim races

ephraimduncan:
## Description
Fix intermittent 500 errors when opening send flows before organisation claim data is fully available.

## Changes Made
- Guard org email-domain checks in the modern send dialog to avoid null-access during first render.
- Apply the same guard and conditional sender-email lookup in the legacy editor

## Checklist


dguyen:
@ephraimduncan how did you replicate this issue in the first place?

ephraimduncan:
I was facing the issue on my dev environment, as well when I pulled from main.

- Upload a new document under the envelope
- Add a recipient
- Add a single signature field
- Immediately after you add the field, send the document, you get 500.
- Go back and open and send again and you can send it.

dguyen:
The issue doesn't have anything to do with the claims/emails. Going to close this and recreate a PR to resolve this issue.

Issue #2466, opened 2026-02-10
Team Creation: Team URL field doesn't strip quote from accentuated letters (é)


landure:
### Issue Description

When creating a new team with accents in the name: "Développement", the Team URL field is set to "développement".
The quote isn't stripped.

### Steps to Reproduce

1. create a team
2. use a name with quotes: "àéèê"

### Expected Behavior

Quotes are stripped: "aeee"

### Current Behavior

Quotes aren't stripped: "àéèê"

### Operating System [e.g., Windows 10]

Ubuntu GNU/Linux

### Browser [e.g., Chrome, Firefox]

Firefox

### Version [e.g., 2.0.1]

2.6.0

### Please check the boxes that apply to this issue report.

- [x] I have searched the existing issues to make sure this is not a duplicate.
- [x] I have provided steps to reproduce the issue.
- [x] I have included relevant environment information.
- [ ] I have included any relevant screenshots.
- [x] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [ ] I want to work on creating a PR for this issue if approved

Jahangirbabar2000:
Hi! I'd like to pick up this issue. I've submitted a fix in [PR #2489](https://github.com/documenso/documenso/pull/2489). Happy to make any adjustments if needed.

P.S: I've already signed the CLA.

PR #2489, opened 2026-02-13
fix: strip diacritics from team URL slug generation

Jahangirbabar2000:
## Description

When creating a team with accented characters in the name (e.g. "Développement", "Ström"), the auto-generated Team URL slug retains diacritical marks. Since the Team URL validator only allows letters, numbers, dashes, and underscores, this causes validation to fail — making it impossible to create teams with accented names.
This PR adds Unicode NFD normalization to the mapTextToUrl function in TeamCreateDialog.tsx to strip diacritical marks before lowercasing and replacing spaces.

## Related Issue

Fixes #[2466](https://github.com/documenso/documenso/issues/2466)

## Changes Made

1. Added .normalize('NFD') to decompose accented characters into their base letter + combining accent mark
2. Added .replace(/[\u0300-\u036f]/g, '') to remove the combining diacritical marks, leaving only ASCII base characters
3. Both steps are applied before the existing .toLowerCase() and .replace(/\s+/g, '-') transformations

## Testing Performed

- Tested team creation with Ström — URL field now correctly generates strom instead of ström
- Tested with Développement — generates developpement
- Tested with Équipe Générale — generates equipe-generale
- Tested with São Paulo — generates sao-paulo
- Tested with naïve — generates naive
- Tested with non-accented names like Engineering — no change in behavior
- Verified that the Team URL validation passes after the fix and teams can be created successfully

## Checklist

## Additional Notes

The fix is a single function change in mapTextToUrl. The approach uses the standard String.normalize('NFD') method which decomposes characters like é into e + a combining acute accent, then strips all combining marks via regex. This covers French, Spanish, German, Portuguese, Nordic, and other Latin-script diacritics.

### DIFF
--- a/apps/remix/app/components/dialogs/team-create-dialog.tsx
+++ b/apps/remix/app/components/dialogs/team-create-dialog.tsx
@@ -127,7 +127,11 @@ export const TeamCreateDialog = ({ trigger, onCreated, ...props }: TeamCreateDia
   };
 
   const mapTextToUrl = (text: string) => {
-    return text.toLowerCase().replace(/\s+/g, '-');
+    return text
+      .normalize('NFD')
+      .replace(/[\u0300-\u036f]/g, '')
+      .toLowerCase()
+      .replace(/\s+/g, '-');
   };
 
   const dialogState = useMemo(() => {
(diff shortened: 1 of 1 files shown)


Mythie:

https://documen.so/cla

Jahangirbabar2000:
Hi @Mythie, I've already signed the CLA a while back — might not be syncing correctly on your end. Happy to re-sign if needed, just let me know!

dguyen (approved):
Looks good to me, thanks!

PR #2508, opened 2026-02-16
feat: add faceted table toolbars and multi-value filters

ephraimduncan:
## Description

Introduces faceted table toolbars and multi-value filters for documents and templates tables, including reusable filter primitives and server/query updates to support the new filter model.

## Demo

02-documents-status-popover-selected
05-templates-type-popover-selected
06-template-detail-source-popover-selected

## Changes Made

- Added dedicated toolbar components for documents, templates, and template detail document tables.
- Added shared table toolbar constants and a new `data-table-faceted-filter` primitive.
- Removed legacy search/period/sender filter components that were replaced by faceted toolbars.
- Updated server-side query/schema/search-param handling for multi-value filters.
- Updated E2E fixtures/specs for team documents and manage templates filter behavior.


packages/trpc/server/template-router/schema.ts
```diff
@@ -289,7 +290,9 @@ export const ZUpdateTemplateRequestSchema = z.object({
 export const ZUpdateTemplateResponseSchema = ZTemplateLiteSchema;
 
 export const ZFindTemplatesRequestSchema = ZFindSearchParamsSchema.extend({
-  type: z.nativeEnum(TemplateType).describe('Filter templates by type.').optional(),
+  type: z
```
dguyen: You can't suddenly change the request input for API V2 routes

dguyen (changes requested):

packages/trpc/server/template-router/schema.ts
ephraimduncan: sure, i'm going to fix that.

ephraimduncan: Updated this to preserve API compatibility: `type` now accepts both the existing single enum value and comma-separated multi-values for the faceted toolbar use case.

dguyen: Lets just leave it as a single enum for now instead of causing type changes.

TemplateType can currently only be public or private. If nothing is passed in it should return both types.

ephraimduncan:
Superseded by #3145 — multi-value filtering landed via the filter pill design instead of table toolbars.

PR #2520, opened 2026-02-19
feat: better ratelimiting

Mythie:
Replace hono-rate-limiter with a Prisma/PostgreSQL bucketed counter
approach that works correctly across multiple instances without sticky
sessions.

- Add RateLimit model with composite PK (key, action, bucket) and atomic upsert
- Create rate limit factory with window parsing, bucket computation, and fail-open
- Define auth-tier and API-tier rate limit instances
- Add Hono middleware, rateLimitResponse helper, and tRPC assertRateLimit helper
- Wire rate limit headers through AppError constructor (was declared but never assigned)
- Apply rate limits to auth routes (email-password, passkey), tRPC routes
  (2FA email, link org account), API routes, and file upload endpoints
- Add cleanup cron job for expired rate limit rows (batched delete every 15 min)
- Remove hono-rate-limiter dependency

### DIFF
--- a/packages/auth/server/routes/email-password.ts
+++ b/packages/auth/server/routes/email-password.ts
@@ -2,6 +2,7 @@ import { sValidator } from '@hono/standard-validator';
 import { compare } from '@node-rs/bcrypt';
 import { UserSecurityAuditLogType } from '@prisma/client';
 import { Hono } from 'hono';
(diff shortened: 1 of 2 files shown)


Mythie:
@coderabbitai review

PR #2554, opened 2026-03-02
feat: add pdf image renderer

dguyen:
## Description

Replace the PDF renderer with an custom image renderer.

This allows us to remove the "react-pdf" dependency and allows us to use a virtual list to improve performance.


packages/lib/client-only/hooks/use-is-page-in-dom.ts
```diff
@@ -0,0 +1,34 @@
...
+  useEffect(() => {
+    const selector = `${PDF_VIEWER_PAGE_SELECTOR}[data-page-number="${pageNumber}"]`;
+
+    setIsPageInDom(document.querySelector(selector) !== null);
+
+    const observer = new MutationObserver(() => {
+      setIsPageInDom(document.querySelector(selector) !== null);
+    });
+
+    observer.observe(document.body, {
```
Mythie: Don't target document.body

PR #2581, opened 2026-03-07
fix: performance improvements

dguyen:
## Description

Various performance and UI fixes:

- Fix pdfjs worker issue 
- Stop revalidation on document/template layout
- UI fixes


Mythie:
Worker Sec comment is all I would have as well

PR #2602, opened 2026-03-11
feat: add envelope pdf replacement

dguyen:
## Description

Allow replacing envelope items in the envelope editor.

The embedding PR for this is here:
https://github.com/documenso/embeds/pull/70

### DIFF
--- a/packages/lib/server-only/envelope-item/replace-envelope-item-pdf.ts
+++ b/packages/lib/server-only/envelope-item/replace-envelope-item-pdf.ts
@@ -135,6 +135,8 @@ export const UNSAFE_replaceEnvelopeItemPdf = async ({
           },
         },
       });
+
+      didFieldsChange = true;
     }
 
     if (recipients.length > 0 && placeholders.length > 0) {


dguyen:
@coderabbitai review

packages/lib/server-only/envelope-item/replace-envelope-item-pdf.ts
```diff
@@ -0,0 +1,142 @@
...
+  envelopeItemId,
+  oldDocumentDataId,
+  data,
+  user,
+  apiRequestMetadata,
+}: UnsafeReplaceEnvelopeItemPdfOptions): Promise<UnsafeReplaceEnvelopeItemPdfResult> => {
+  const buffer = Buffer.from(await data.file.arrayBuffer());
+  const normalized = await normalizePdf(buffer, {
+    flattenForm: envelope.type !== 'TEMPLATE',
+  });
```
Mythie, line 86: If we're flattening here we will need formData passed from documentMeta

PR #2604, opened 2026-03-12
feat: allow editing pending envelope titles

dguyen:
## Description

Add ability to modify the envelope titles and envelope item titles for pending envelopes.

Note: This PR will conflict a little with #2602 so this should be merged afterwards and fixed.

We will probably need to resolve the audit logs and some additional restriction logic.


dguyen:
@coderabbitai review

PR #2608, opened 2026-03-12
fix: improve webhook execution

Mythie:
Webhook URLs were being fetched without validating whether they
resolved to private/loopback addresses, exposing the server to SSRF.

Current SSRF is best effort and fail open, you should never host services that
you cant risk exposure of.

This extracts webhook execution into a shared module that validates
URLs against private IP ranges (including DNS resolution), enforces
timeouts, and disables redirect following. The resend route now
queues through the job system instead of calling fetch inline.

### DIFF
--- a/packages/trpc/server/webhook-router/resend-webhook-call.ts
+++ b/packages/trpc/server/webhook-router/resend-webhook-call.ts
@@ -5,10 +5,14 @@ import { buildTeamWhereQuery } from '@documenso/lib/utils/teams';
 import { prisma } from '@documenso/prisma';
 
 import { authenticatedProcedure } from '../trpc';
-import { ZResendWebhookCallRequestSchema } from './resend-webhook-call.types';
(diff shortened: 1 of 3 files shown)


packages/trpc/server/webhook-router/resend-webhook-call.ts
```diff
@@ -1,20 +1,14 @@
...
 
 import { authenticatedProcedure } from '../trpc';
-import {
-  ZResendWebhookCallRequestSchema,
-  ZResendWebhookCallResponseSchema,
-} from './resend-webhook-call.types';
+import { ZResendWebhookCallRequestSchema } from './resend-webhook-call.types';
 
 export const resendWebhookCallRoute = authenticatedProcedure
   .input(ZResendWebhookCallRequestSchema)
```
dguyen, line 14: Always have an `.output`

PR #2619, opened 2026-03-14
perf: parallelize async operations in duplicateEnvelope

ephraimduncan:
Parallelizes independent async operations in `duplicateEnvelope` using `Promise.all`, eliminating two sequential waterfalls: counter increment with document meta creation, and recipient creation across all recipients.


packages/lib/server-only/envelope/duplicate-envelope.ts
```diff
@@ -136,35 +136,37 @@ export const duplicateEnvelope = async ({ id, userId, teamId }: DuplicateEnvelop
...
+              })),
+            },
           },
         },
-      },
-    });
-  }
+      }),
+    ),
+  );
```
dguyen: I'd use a pmap 5 for this one incase

PR #2621, opened 2026-03-16
perf: add prefetch="intent" to nav Link components

ephraimduncan:
### Description

Enables React Router's intent-based prefetching on Link components across the app, preloading route data and modules on hover/focus for faster perceived navigation.


dguyen:
I don't think we want to do a massive change like this without fully understanding the consequences, if we're were to do this it would be better to start small.

PR #2639, opened 2026-03-21
feat: auto insert email and date fields

dguyen:
## Description

Auto insert following fields:
- Date fields
- Email fields (only when email is provided)


apps/remix/app/components/general/document-signing/envelope-signing-provider.tsx
```diff
@@ -83,14 +86,62 @@ export interface EnvelopeSigningProviderProps {
...
+    if (field.type !== FieldType.DATE || field.inserted) {
+      return field;
+    }
+
+    return {
+      ...field,
+      customText: formattedDate,
+      inserted: true,
+      fieldMeta: {
+        ...(typeof field.fieldMeta === 'object' && field.fieldMeta !== null ? field.fieldMeta : {}),
```
Mythie: Spreading null or undefined results in them being an empty object. We probably don't need these sloppy checks.

PR #2641, opened 2026-03-21
feat: add option to disable document rejection per org/team

### LINKED ISSUE #2561: Add option to disable document rejection per team/organisation
### Feature Description

We're self-hosting Documenso as an internal document signing tool. In our workflow, rejecting a document is not a valid action — recipients are always expected to sign. Currently, there is no way to hide or disable the "Reject" button on the signing page.

### Use Case

For self-hosted instances used as a personal or internal DocuSign replacement, rejection often doesn't make sense. Documents are pre-agreed and only need a signature. Showing a reject button introduces unnecessary confusion and an undesired path in the workflow.
Ideally, this setting could also be overridden per document (similar to how language and visibility work today), so teams that need rejection for some documents but not others can handle both cases.

### Proposed Solution

Add a toggle to the existing Document Preferences (on organisation and/or team level) such as "Allow document rejection" (default: enabled). When disabled, the reject option would not be shown to recipients on the signing page.
This would fit naturally alongside the existing preferences like signature settings, document visibility, and sender details.

### Alternatives (optional)

_No response_

### Additional Context

_No response_

- [x] I want to work on creating a PR for this issue if approved

oniani1:
Fixes #2561

Adds an "Allow Document Rejection" toggle to org and team document preferences. When disabled, the reject button is hidden on the signing page.

The setting follows the existing inheritance pattern: org sets the default, teams can override, and it can also be set per-document via DocumentMeta for more granular control.

Changes:
- Added `allowDocumentRejection` to OrganisationGlobalSettings (default true), TeamGlobalSettings (nullable), and DocumentMeta (nullable)
- Added the toggle to the document preferences form
- V2 signing page now checks both the embed context and documentMeta setting
- V1 signing page now accepts and respects the prop
- Template-created documents inherit the setting from the template

### DIFF
--- a/apps/remix/app/components/forms/document-preferences-form.tsx
+++ b/apps/remix/app/components/forms/document-preferences-form.tsx
@@ -76,6 +76,7 @@ export type TDocumentPreferencesFormSchema = {
   includeSenderDetails: boolean | null;
   includeSigningCertificate: boolean | null;
   includeAuditLog: boolean | null;
+  allowDocumentRejection: boolean | null;
   signatureTypes: DocumentSignatureType[];
   defaultRecipients: TDefaultRecipients | null;
   delegateDocumentOwnership: boolean | null;
@@ -93,6 +94,7 @@ type SettingsSubset = Pick<
   | 'includeSenderDetails'
   | 'includeSigningCertificate'
   | 'includeAuditLog'
+  | 'allowDocumentRejection'
   | 'typedSignatureEnabled'
   | 'uploadSignatureEnabled'
   | 'drawSignatureEnabled'
@@ -134,6 +136,7 @@ export const DocumentPreferencesForm = ({
     includeSenderDetails: z.boolean().nullable(),
     includeSigningCertificate: z.boolean().nullable(),
     includeAuditLog: z.boolean().nullable(),
+    allowDocumentRejection: z.boolean().nullable(),
     signatureTypes: z.array(z.nativeEnum(DocumentSignatureType)).min(canInherit ? 0 : 1, {
       message: msg`At least one signature type must be enabled`.id,
     }),
@@ -156,6 +159,7 @@ export const DocumentPreferencesForm = ({
       includeSenderDetails: settings.includeSenderDetails,
       includeSigningCertificate: settings.includeSigningCertificate,
       includeAuditLog: settings.includeAuditLog,
+      allowDocumentRejection: settings.allowDocumentRejection,
       signatureTypes: extractTeamSignatureSettings({ ...settings }),
       defaultRecipients: settings.defaultRecipients
         ? ZDefaultRecipientsSchema.parse(settings.defaultRecipients)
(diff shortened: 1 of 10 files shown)


Mythie (approved):

oniani1:
Hey @Mythie, just checking in -- is there anything else needed on this one? Noticed it's been approved for a few days, wanted to make sure nothing's blocking the merge on my end.

oniani1:
Hey @Mythie, just circling back -- this has been approved for a couple weeks now. Anything blocking the merge on your end?

dguyen (changes requested):
There seems to be some build issues, could you please resolve those?

oniani1:
Rebased on main to pick up the signing reminders changes from #1749. Conflicts in schema.prisma and document.ts were straightforward, both features just coexist side by side.

oniani1:
Hey @dguyen, the only failing checks are the Vercel deploy previews — those need a team member to authorize the fork deployment, so not something I can resolve on my end. All the actual CI checks (GitGuardian, PR title validation, labeler) pass green. Let me know if there's anything else needed!

HiradRoshandel:
Hey folks is there any update on this PR merge?

gauravbsinghal:
Hi @oniani1, I'm sharing this from [Niro](https://github.com/apxlabs-ai/niro), a PR pentesting tool we're building to help OSS maintainers catch security regressions before merge. I ran it against this PR and it found the following issue.

Any recipient holding a signing token can reject a document even when that document has rejection disabled, overriding the sender’s policy and driving the document into the rejected/cancelled workflow.

## Reproduction Steps

Precondition: use fixture `disabled_rejection_signing_envelope` from `/niro/config/fixtures.yaml` (`documentId` 2, recipient token `-hCWKPHM8aC4LOG4GUHXO`, `document_meta.allowDocumentRejection: false`).
1. Send a direct reject request instead of using the signing UI:
   `curl -i -X POST 'http://host.docker.internal:3000/api/trpc/recipient.rejectDocumentWithToken' -H 'content-type: application/json' --data '{"json":{"documentId":2,"token":"-hCWKPHM8aC4LOG4GUHXO","reason":"niro direct rejection probe run 1"}}'`
   Expected secure behavior: reject the request with a 4xx error because document rejection is disabled, and leave the recipient in `NOT_SIGNED`.
   Observed behavior: `HTTP/1.1 200 OK` and JSON result `signingStatus:"REJECTED"`, `rejectionReason:"niro direct rejection probe run 1"`, `signedAt:"2026-06-15T23:29:07.850Z"`.
2. Repeat the same call to confirm reproducibility:
   `curl -i -X POST 'http://host.docker.internal:3000/api/trpc/recipient.rejectDocumentWithToken' -H 'content-type: application/json' --data '{"json":{"documentId":2,"token":"-hCWKPHM8aC4LOG4GUHXO","reason":"niro direct rejection probe run 2"}}'`
   Observed behavior again: `HTTP/1.1 200 OK` and JSON result `signingStatus:"REJECTED"`, `rejectionReason:"niro direct rejection probe run 2"`, `signedAt:"2026-06-15T23:29:08.152Z"`.
3. Source review confirms the server mutation path has no `allowDocumentRejection` guard: `packages/trpc/server/recipient-router/router.ts` forwards directly to `packages/lib/server-only/document/reject-document-with-token.ts`, and that function updates the recipient to `SigningStatus.REJECTED` without checking `envelope.documentMeta.allowDocumentRejection`.

## Remediation Guidance

Enforce the policy in `packages/lib/server-only/document/reject-document-with-token.ts`, not only in the signing-page components. Load `envelope.documentMeta` with the recipient query and, before the recipient update transaction, throw the existing `AppError(AppErrorCode.INVALID_REQUEST, ...)` when `envelope.documentMeta.allowDocumentRejection === false` (treat `null` as the inherited/default-allowed path). Keep the UI hiding logic as defense in depth, but make the server-side reject path authoritative.

If this does not match the intended security model for this PR, please reply with the expected behavior and I will treat that as signal to improve the test.

catalinpit:
Hi @oniani1, thanks for the work and sorry for the slow progress on our side. If you wish to fix the conflicts, we can merge this.

catalinpit:
Thanks for starting the work on this. We'll continue it in https://github.com/documenso/documenso/pull/3362, which expands on it: it saves the setting when an organisation or team is updated, adds support for the mobile menu, V1, and embeds, and lets users override it for each document in the editor. It also checks the setting on the server in `rejectDocumentWithToken`. The schema and the logic for determining the setting stay the same as in this PR.

PR #2642, opened 2026-03-21
fix: clean up orphaned DocumentData after envelope item updates

### LINKED ISSUE #2493: [BUG] DocumentData is duplicating on envelope updates
### Issue Description

Hello to all project contributors! I recently noticed a behavior in the project and would like to ask a question and, if applicable, report a possible bug.

I'm noticing strange behavior with the DocumentData table and the methods: `putNormalizedPdfFileServerSide()` and `putPdfFileServerSide()`.

Both functions call the `createDocumentData()` function, which uploads and creates a new record in the DocumentData table.

All functions work perfectly; the problem is that there are flows that call the put functions when the objective is simply to update the data of an existing document, which generates duplicate records in the table.

There are flows that "compensate" for this behavior and update the `"DocumentData"."data"` attribute with the data from the most recent document, but even so, the duplicate records remain.

Should I change the flows to call the `updateFile() `function? Or is this data duplication not a bug but an audit feature?

### Steps to Reproduce

1. Create a new envelope
2. Finish signing the envelope.
3. Check DocumenteData table

### Expected Behavior

_No response_

### Current Behavior

_No response_

### Screenshots (optional)

_No response_

_No response_

_No response_

### Version [e.g., 2.0.1]

_No response_

- [ ] I have searched the existing issues to make sure this is not a duplicate.
- [ ] I have provided steps to reproduce the issue.

oniani1:
Fixes #2493

When envelope items are updated (during sealing or field creation with whiteout regions), the code creates a new DocumentData record and updates the EnvelopeItem foreign key to point to it. The old DocumentData record was left orphaned in the database.

This follows the same cleanup pattern already used in replace-envelope-item-pdf.ts, which correctly deletes the old record after replacing. Applied the same approach to:

- seal-document.handler.ts: deletes old DocumentData after sealing completes
- create-envelope-fields.ts: deletes old DocumentData after whiteout region update

### DIFF
--- a/packages/lib/jobs/definitions/internal/seal-document.handler.ts
+++ b/packages/lib/jobs/definitions/internal/seal-document.handler.ts
@@ -311,6 +311,13 @@ export const run = async ({
       });
     });
 
+    // Delete orphaned DocumentData records that were replaced during sealing.
+    for (const { oldDocumentDataId } of newDocumentData) {
+      await prisma.documentData.delete({
+        where: { id: oldDocumentDataId },
+      });
+    }
+
     return {
       envelopeId: envelope.id,
       envelopeStatus: envelope.status,
--- a/packages/lib/server-only/field/create-envelope-fields.ts
+++ b/packages/lib/server-only/field/create-envelope-fields.ts
@@ -308,6 +308,8 @@ export const createEnvelopeFields = async ({
       continue;
     }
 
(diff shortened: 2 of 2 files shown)


oniani1:
Hey, checking in on this one too -- let me know if anything needs adjusting.

PR #2653, opened 2026-03-25
chore: Update labeler.yml

ElTimuro:
added docs label to labeler and fixed app: web label

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -2,8 +2,7 @@
   - apps/remix/**
 
 'type: documentation'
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -1,5 +1,9 @@
 'apps: web':
-  - apps/web/**
+  - apps/remix/**
+
+'type: documentation'
+  - apps/documentation/**
```
dguyen: `apps/documentation` no longer exists

it should be `apps/docs/**` or something

Issue #2650, opened 2026-03-24
Add env var to allow webhooks to private/internal URLs for self-hosted instances


jpsimonsen:
### Feature Description

PR #2608 added SSRF protection that blocks webhook URLs resolving to private/loopback IPs. This breaks self-hosted instances where webhook targets are on private networks (VPC, Docker, Tailscale, LAN). There is currently no way to opt out of this check.

### Use Case

Self-hosted Documenso instance operate within private networks (VPC, Docker, Tailscale, LAN) where webhook targets resolve to private IPs (10.x.x.x, 172.16.x.x, 192.168.x.x). After upgrading to the latest Docker image (post-v2.8.0), all webhooks to these targets fail with: "Webhook URL resolves to a private or loopback address". There is no workaround without patching the container.

### Proposed Solution

Add a NEXT_PRIVATE_WEBHOOK_ALLOW_PRIVATE_URLS env var. When set to "true", skip the private URL check in assertNotPrivateUrl().

`if (process.env.NEXT_PRIVATE_WEBHOOK_ALLOW_PRIVATE_URLS === 'true') {
  return;
}`

### Alternatives (optional)

Routing webhooks through a public IP (unnecessary complexity for private network setups)

### Additional Context

Introduced in PR #2608 (merged March 13, 2026)

### Please check the boxes that apply to this feature request.

- [x] I have searched the existing feature requests to make sure this is not a duplicate.
- [x] I have provided a detailed description of the requested feature.
- [x] I have explained the use case or scenario for this feature.
- [x] I have included any relevant technical details or design suggestions.
- [x] I understand that this is a suggestion and that there is no guarantee of implementation.
- [ ] I want to work on creating a PR for this issue if approved

Mythie:
Sounds good, if you have the changes done already on your fork feel free to open a PR to save time on getting this through.

scoobs-mystery-bytes:
has this moved forward at all? this would be fantastic

jakkepedder:
Yes, since v2.9.0

Mythie:
Yeah this has been live since v2.9.0, I'll get our docs updated to include the env var.

PR #2654, opened 2026-03-25
feat: webhook allow private hosts

jpsimonsen:
## Description

Adds NEXT_PRIVATE_WEBHOOK_SSRF_BYPASS_HOSTS — a comma-separated list of hostnames or IPs that are permitted to resolve to private/loopback addresses when used as webhook targets.

## Related Issue

PR #2608 added SSRF protection that blocks webhook URLs resolving to private IPs. This is correct for the hosted product, but breaks self-hosted instances where webhook targets live on the same private network (Docker, VPC, Tailscale, LAN).

Addresses #2650.

## How it works

The env var is parsed once at module load into an immutable Set. The check runs at the top of assertNotPrivateUrl() before any existing logic — if the webhook URL's hostname matches an entry, all SSRF checks are skipped for that URL only.

## Changes Made

- packages/lib/server-only/webhooks/assert-webhook-url.ts: Added a bypass gate at the beginning of the assertNotPrivateUrl() function. If a host is explicitly whitelisted, the private IP check is skipped.

- .env.example: Added the new environment variable with documentation on its usage.

- Implemented explicit whitelisting only; no wildcard or blanket bypasses are permitted.

## Testing Performed

- Manual Validation: Verified that the string from the environment variable is correctly parsed into a set of unique, lowercased strings.

- Logic Check: Confirmed that when the variable is empty or unset, the system defaults to strict blocking (standard SSRF protection remains active).

- Functionality: Verified that a host listed in the bypass variable successfully skips the WEBHOOK_INVALID_REQUEST error.

## Checklist

## Additional Notes

No new tests included. The bypass Set is built at module load, so testing it requires either a separate test file or module re-importing.


Mythie:

https://documen.so/cla

PR #2677, opened 2026-04-03
feat: dialog confirmation before duplicating a field on all pages

MininuxDev:
## Description

Display a `<Dialog/>` when clicking the "duplicate on all pages" button. Only perform the action after the user has pressed the confirmation button. This is to prevent accidental press by the user, resulting in them having to manually remove every single duplicated item from each page.

For now, only applies to the new envelope editor, not the legacy editor (document-flow and template-flow).

## Related Issue

Fixes #2674

## Changes Made

- In `FieldActionButtons`, added a `<Dialog/>`, and replaced the call to `handleDuplicateSelectedFieldsOnAllPages` to display the dialog, which will be the one to call the handler instead.

## Testing Performed

- Tested that the duplication still works on several fields types, when editing a document or template (Firefox)
- Tested on mobile too (Firefox mobile)
- Tested in authoring embed (Firefox & Firefox mobile)

## Checklist

## Additional Notes

For now, I only added the dialog to the new envelope editor. The legacy editor detects the field to act on by checking the currently focused field (state variable `lastActiveField` gets set to null on blurr), which looses focus when displaying the dialog, so just adding a dialog is not enough. It's not that hard but requires more changes, I can try something if you want the legacy editor to be updated too.

For the tests, I had to make `openDocumentEnvelopeEditor` be able to create a multi page document, instead of the hardcoded 1 page document.


Mythie:

https://documen.so/cla

PR #2691, opened 2026-04-09
fix: error message (update title)

catalinpit:
Previously, all errors during template use showed a generic message. Now, server-side error details (e.g. userMessage) are displayed in the toast, and AppError instances are re-thrown instead of being swallowed by a generic DOCUMENT_SEND_FAILED error.

## Checklist


dguyen:
Surfacing the error for API V2 requests is fine

We shouldn't surface the raw error message to the front end toast. It needs to be translatable and cleaned

Mythie (approved):

PR #2703, opened 2026-04-16
feat: add turnstile captcha to auth flow

Mythie:
Integrate Cloudflare Turnstile on sign-in (invisible) and sign-up (visible) forms, with server-side verification via siteverify. 

The feature is opt-in: when NEXT_PRIVATE_TURNSTILE_SECRET_KEY is unset, verification is skipped so deployments without credentials continue to work unchanged. Likewise the widget is only rendered when a site-key is present.

Also tighten the signup rate limit from 10/hour to 3/3h to further discourage mass manual account creation.


dguyen (approved):

tinymachine:
FYI when I tried signing in on [documenso.com](https://app.documenso.com/signin) with username and password using Safari on macOS 26.3.1 and again on macOS 26.4.1, I got the error: 

> Unable to sign in
> We were unable to verify that you're human. Please try again.

It looks like Safari's Prevent Cross-Site Tracking feature, which I think is enabled by default, is not allowing Cloudflare's turnstile captcha to work properly. I was able to sign in after disabling the feature (Safari Settings > Privacy). As a workaround, once signed in I created a passkey, and then reenabled the feature, and now I can sign in using the passkey without issue.

Just mentioning this because I'm guessing I'm not the only one experiencing this issue.

Note: this error did not appear when logging in using username and password with Safari on iOS 26.4.1; I was able to sign in without issue there, even though Prevent Cross-Site Tracking was enabled.

PR #2711, opened 2026-04-20
feat: bulk download documents

ephraimduncan:
## Background

Users who select multiple documents in the list today can only move or delete them — no way to download the PDFs in bulk. This adds a **Download** action to the bulk toolbar that opens a dialog, lets the user pick signed/original per completed document, and streams each file sequentially with progress, partial-failure, and abort handling.

Escape-to-clear is also wired to the existing bulk selection so the toolbar can be dismissed from the keyboard.

## Demo
https://github.com/user-attachments/assets/7d806daa-0742-4b7b-bf23-865cfca5ff71

## Verification

1. On `/t/<team>/documents`, select 2+ documents via the row checkboxes.
2. Click **Download** in the bulk action toolbar.
3. In the dialog:
   - Titles and statuses render correctly.
   - Completed rows expose a Signed/Original select; pending/draft/rejected rows do not.
   - Clicking **Download** streams each PDF and shows `Downloading n / total…`.
   - Clicking **Stop** mid-run aborts at the next boundary.
   - On full success: "Documents downloaded" toast, dialog closes, only the successful rows are deselected.
   - On partial failure: destructive "Documents partially downloaded" toast, failed rows stay selected.
4. With a selection present, press **Esc** → toolbar and selection clear.
5. `npm run test:e2e -- bulk-document-actions` passes.


Mythie:
We probably want to turn a mutliple select download in the UI into a zip file

Mythie (approved):

PR #2725, opened 2026-04-23
feat: Redaction

AcidRaZor:
## Description                                                                                                      
   
  Adds the ability for senders to mark regions of a PDF for **true redaction** before sending it for signing. Unlike a
   drawing-layer overlay (e.g. `page.drawRectangle`), redacted content is physically removed from the PDF: each     
  affected page is rasterized with the redaction regions burned in as solid black, and the resulting image replaces   
  the original page. The original text operators, annotations, and form widgets on that page are gone — not just    
  visually obscured. Pages without redactions are left as vector content so they stay searchable and small.

  Redactions are applied when the document transitions from DRAFT to PENDING (inside `sendDocument`), so recipients   
  never receive the unredacted bytes. Both `DocumentData.data` and `initialData` are overwritten with the redacted PDF
   — resealing later cannot resurrect the original, since the seal job uses `initialData` as its source.                                                                                                                                                                                                  
                                                                                                                    
  ## Changes Made

  **Data model**                                                                                                      
  - New `Redaction` Prisma model with back-relations on `Envelope` and `EnvelopeItem`, plus migration.
                                                                                                                      
  **PDF pipeline** (`packages/lib/server-only/pdf/`, `packages/lib/server-only/redaction/`)                           
  - `apply-redactions-to-pdf.ts` — core rasterize-and-replace function using `pdfjs-dist` + `skia-canvas` for         
  rendering and `@cantoo/pdf-lib` for page removal/insertion. Pages processed in reverse index order to keep          
  subsequent indices stable. JPEG-encoded raster (q=0.95) to keep output size bounded. Early-returns when           
  `redactions.length === 0`.                                                                                          
  - `strip-pdf-metadata.ts` — deletes Info-dict keys (Title/Author/Subject/Keywords/Creator/Producer) and catalog   
  entries (Names/AcroForm/OpenAction/AA) to prevent leaks through embedded attachments, JavaScript triggers, or       
  leftover form data.
  - `apply-redactions-to-document.ts` — the storage-aware orchestrator: read → flatten (`normalizePdf`) → rasterize + 
  paint redactions → strip metadata → write new DocumentData (overwriting both `data` and `initialData`).             
  - Wired into `send-document.ts` immediately before form-value injection. Redaction rows are deleted after
  application — they represent pending-to-apply work and are meaningless once baked in.                               
                                                                                                                    
  **API**                                                                                                             
  - `packages/trpc/server/redaction-router/` — `createDocumentRedactions`, `updateDocumentRedactions`,              
  `deleteDocumentRedaction`. All guard against non-DRAFT envelope status and reject zero-size regions.                
  - Server-only CRUD functions in `packages/lib/server-only/redaction/`.
  - Editor envelope loader extended to include redactions so they round-trip through the editor.                      
                                                                                                                      
  **UI** (`apps/remix/app/components/general/envelope-editor/`)                                                       
  - New "Redact" toolbar button in the fields sidebar.                                                                
  - Click-to-place drag-drop with cursor preview.                                                                     
  - Konva-rendered redaction groups on the same page layer as fields, with their own transformer so selection states  
  don't cross. Full parity with field UX: click to select, drag to move, resize handles, trash button when selected.  
  - Redaction badge is a solid black rectangle with a centered "REDACTED" label; hover shows a red outline.           
  - Disabled entirely once the envelope leaves DRAFT.                                                                 
                                                                                                                      
  **Client state**                                                                                                    
  - `useEditorRedactions` hook mirroring `useEditorFields`.                                                           
  - Provider diffs local state against `envelopeRef.current.redactions` to fan out create/update/delete mutations.    
                                                                                                                      
  ## Testing Performed                                                                                                
                                                                                                                      
  **Automated**                                                                                                       
  - Vitest unit tests for `strip-pdf-metadata` (Info-dict delete, catalog entry removal via round-trip with         
  `doc.attach()` + `getOrCreateAcroForm()` so the assertion is honest about what's actually removed, not just the     
  method call).
  - Vitest unit tests for `apply-redactions-to-pdf`: empty-case no-op, true-redaction assertion (`getTextContent`     
  returns empty after redaction — proves content stream is actually gone, not overlaid), per-page isolation (redacted 
  page = no vector text; unredacted page still has vector text), page-count preservation, rotated source page,
  out-of-range page rejection.                                                                                        
  - Playwright E2E (`packages/app-tests/e2e/envelopes/redaction.spec.ts`) exercising the full `sendDocument` →      
  `applyRedactionsToDocument` pipeline. Places a Redaction row, sends, pulls the resulting `DocumentData` bytes back, 
  and asserts both a known top-of-page string (inside the redaction) and a known bottom-of-page string (outside, but
  also erased because rasterize-replace rewrites the whole page) are absent from the distributed PDF.                 
                                                                                                                    
  **Manual**
  - Upload + place redactions + send flow end-to-end with real PDFs.
  - Verified text cannot be selected / copy-pasted from the redacted page in Acrobat, Preview, and Chrome's PDF       
  viewer.                                                                                                             
  - `pdftotext` on the distributed PDF confirmed no trace of the redacted strings.                                    
  - Metadata verified empty via `pdfinfo`.                                                                            
                                                                                                                      
  ## Checklist                                                                                                        
                                                                                                                      
  ## Additional Notes
                                                                                                                      
  **Known scope reductions for V1** (call out if reviewers want any lifted before merge):                             
  - Documents only. Templates are not supported yet — the CRUD functions explicitly scope to `EnvelopeType.DOCUMENT`.
  - Solid black only; no reason labels, colors, or per-recipient redactions.                                          
  - Once a document is sent, redactions are permanent. There is no un-redact path — the original bytes are            
  deliberately overwritten in `DocumentData.initialData` to defeat resealing.                                         
                                                                                                                      
  **Design choices worth flagging**                                                                                   
  - Rasterize-and-replace was chosen over trying to edit the PDF content stream because `pdf-lib` does not expose   
  content-stream manipulation, and rasterization is the technique commercial redaction tools use when they apply      
  (rather than just mark) redactions. Trade-off: text on affected pages is no longer selectable/searchable. Unaffected
   pages are left as vector content.                                                                                  
  - Replaced pages lose `/Rotate` / `/CropBox` / `/UserUnit` — the raster is generated at the rotated viewport so   
  visual orientation is preserved, but downstream code that reads `page.getRotation()` on a redacted page will see 0. 
  Field placement uses page-percentage coordinates so this is a no-op for our pipeline, but flagged for anyone
  extending it.                                                                                                       
  - Saved with `useObjectStreams: false` so the output is independently auditable and no cleared objects can hide   
  inside a compressed object stream.                                                                                  
  - Redaction CRUD uses `deleteMany` server-side for idempotency — the client diff can race with concurrent creates
  and ask to delete a row that was never committed; making delete tolerant is cheaper than eliminating every          
  client-side race.

### DIFF
--- a/apps/docs/content/docs/users/documents/meta.json
+++ b/apps/docs/content/docs/users/documents/meta.json
@@ -1,4 +1,4 @@
 {
   "title": "Documents",
-  "pages": ["upload", "add-recipients", "add-fields", "send", "direct-links", "advanced"]
+  "pages": ["upload", "add-recipients", "add-fields", "redact", "send", "direct-links", "advanced"]
 }
--- a/apps/docs/content/docs/users/documents/redact.mdx
+++ b/apps/docs/content/docs/users/documents/redact.mdx
@@ -1,0 +1,151 @@
+---
+title: Redact
+description: Permanently remove sensitive content from a PDF before sending it for signing.
+---
+
+import { Callout } from 'fumadocs-ui/components/callout';
+import { Step, Steps } from 'fumadocs-ui/components/steps';
+import { Tab, Tabs } from 'fumadocs-ui/components/tabs';
+
+## Overview
+
+Redaction lets you mark regions of a PDF to be permanently removed before the document is distributed to recipients. Unlike drawing a black rectangle on top of content — which leaves the underlying text in the file and recoverable by any PDF parser — Documenso performs **true redaction**: the entire affected page is rasterized and the redaction regions are painted solid black into the pixels. The original text, annotations, and form data on that page are no longer in the PDF in any form.
+
+Redactions are applied exactly once, at the moment you send the document. Before that, they're editable and reversible. After that, they're permanent.
+
+<Callout type="warn">
+  Redaction is a one-way operation. Once you send a document with redactions, the original content is
+  destroyed and cannot be recovered — not by you, not by support, not by re-downloading. Review
+  carefully before sending.
+</Callout>
+
+## Add a Redaction
+
+{/* prettier-ignore */}
+<Steps>
+<Step>
+### Open the document in the editor
+
+Upload your PDF and proceed to the **Add Fields** step as you would for any document.
+
+</Step>
+
+<Step>
+### Click the Redact button
+
+Scroll the sidebar below the field palette to find the **Redact** section. Click the **Redact** button — the cursor becomes a small black preview rectangle that follows your mouse over the document.
+
+</Step>
+
+<Step>
+### Click to place
+
+Click anywhere on the document to drop a redaction at that position. The redaction appears as a solid black rectangle labelled "REDACTED".
+
+</Step>
+
+<Step>
+### Adjust size and position
+
+Click the redaction to select it. Drag the rectangle to reposition, or drag the corner handles to resize so it fully covers the content you want removed.
+
+</Step>
+</Steps>
+
+<Callout type="info">
+  You can place as many redactions as you need, across any pages of any document in the envelope.
+  Redactions are saved automatically as you place, move, and resize them.
+</Callout>
+
+## Move, Resize, and Delete
+
+<Tabs items={['Move', 'Resize', 'Delete']}>
+<Tab value="Move">
+
+Click a placed redaction to select it, then drag it to a new position. Redactions cannot be dragged outside the page bounds.
+
+</Tab>
+<Tab value="Resize">
+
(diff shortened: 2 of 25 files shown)


Mythie:

https://documen.so/cla

AcidRaZor:
> 
> https://documen.so/cla

Actually accepted it several hours before :)

AcidRaZor:
Closing since you're not interested

PR #2729, opened 2026-04-24
feat: added lingui reference in README.MD

vishalkumargeed:
## Description

Added lingui reference so from now on new developers can know that it's a safe practice to use ```npm run translate``` after dealing with lingui code that is :
t , msg , <Trans> </Trans>, etc 

## Additional Notes

Take a look at this conversation for reference : 
https://discord.com/channels/1132216843537485854/1133420505030987857/1497328822738223286

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,8 @@ git clone https://github.com/<your-username>/documenso
 ```sh
 npm run d
 ```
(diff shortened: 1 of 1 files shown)


dguyen:
Hi, we don't want developers to run `npm run translate` as part of their PR changes, because it causes a bunch of conflicts when merging PO files.

PR #2741, opened 2026-04-28
feat: remove default personal orgs from custom sso

dguyen:
## Description

Prevents personal organisations from being created when they sign in through an organisation's SSO portal

Pending testing since I need to setup OIDC again

### DIFF
--- a/packages/auth/server/lib/utils/handle-oauth-organisation-callback-url.ts
+++ b/packages/auth/server/lib/utils/handle-oauth-organisation-callback-url.ts
@@ -76,7 +76,7 @@ export const handleOAuthOrganisationCallbackUrl = async (
       },
     });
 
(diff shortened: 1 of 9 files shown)


packages/auth/server/lib/utils/handle-oauth-organisation-callback-url.ts
```diff
@@ -76,7 +76,7 @@ export const handleOAuthOrganisationCallbackUrl = async (
       },
     });
 
-    await onCreateUserHook(userToLink).catch((err) => {
```
Mythie, line 79: Should we do this by default? Seems like we'd want a toggle on the org if they want to enforce this right?

That way we can use the same flag to determine if they can create a personal org after signing up which may also not be allowed.

Mythie (approved):

PR #2742, opened 2026-04-28
fix: prevent stored XSS in app banner by removing unsafe HTML rendering

namelessweakl1ng:
## Description

Fixes a stored XSS vulnerability in the app banner component caused by unsafe HTML rendering.

## Related Issue

N/A

## Changes Made

- Removed usage of `dangerouslySetInnerHTML`
- Replaced it with safe text rendering using `{banner.data.content}`
- Eliminated risk of arbitrary HTML/JavaScript execution from banner content

## Testing Performed

- Verified that banner content still renders correctly as text
- Confirmed that HTML tags are no longer interpreted or executed
- Ensured no runtime errors after the change

## Checklist

## Additional Notes

This change prioritizes security by removing support for raw HTML rendering in banner content, preventing potential stored XSS attacks without introducing additional dependencies.

### DIFF
--- a/apps/remix/app/components/general/app-banner.tsx
+++ b/apps/remix/app/components/general/app-banner.tsx
@@ -16,7 +16,7 @@ export const AppBanner = ({ banner }: AppBannerProps) => {
         style={{ color: banner.data.textColor }}
       >
         <div className="flex items-center">
-          <span dangerouslySetInnerHTML={{ __html: banner.data.content }} />
+          <span>{banner.data.content}</span>
         </div>
       </div>
     </div>


Mythie:

https://documen.so/cla

namelessweakl1ng:
I've completed the CLA. Please let me know if anything else is needed.

Mythie:
The banner is purely used by admins, being able to use html is beneficial since we occasionally need to link to an announcement.

If a rogue admin were to XSS the application that seems like more like an organisational failure?

namelessweakl1ng:
That makes sense ,thanks for the clarification!

You're right that HTML support is useful for announcements and links. My initial fix removed `dangerouslySetInnerHTML` entirely to eliminate the XSS risk, but I agree that this also removes intended functionality.

A better approach might be to keep HTML support while sanitizing the content before rendering (e.g. using a library like DOMPurify) to prevent script injection while still allowing safe tags like links.

Happy to update the PR to use sanitization instead if that works for you.

PR #2749, opened 2026-04-30
refactor(ui): replace hardcoded colors with semantic tokens

ephraimduncan:
### Description

Swap hardcoded Tailwind colors (bg-white, text-gray-*, border-gray-*, bg-neutral-*, text-red-500, !ring-red-500) and their dark: overrides for the project's semantic tokens (bg-background, text-muted-foreground, border-border, bg-muted, text-destructive, !ring-destructive) across field, form, command, dropzone, signature-pad and related primitives. Also reorders Tailwind utilities to match the prettier-plugin sort order and drops the typo'd text-primary-forground class from the metric card.

No behavioral changes; theme tokens already mirror the previous light/dark palette so the redundant dark: variants are removed in the same pass.


dguyen (changes requested):
Some of these are purposefully hardcoded. For example the ones that overlay a PDF should remain as is, since PDFs always has a white background.

Eg: `envelope-recipient-field-tooltip` should not be modified

Also remove the po files.

dguyen (approved):

Issue #1701, opened 2025-03-13
Make hand-drawing signature looks better under default setting


gavinying:
## Current design
The hand-drawing signature often looks small. 

![Image](https://github.com/user-attachments/assets/0fdda107-44e0-4e52-8b37-7cc993bc9c29)

## Analysis
The reason is,
- signature image will scale down 80% (create padding) when saving
- unnecessary components (upload button and color selector) inside canvas, make user tend to leave empty in top area
- no clear indication of the area for drawing.

![Image](https://github.com/user-attachments/assets/492fe6fe-c4ac-4a4d-841c-a5ef62125dfc)

## Imrprovement suggestion
Here are my thoughts, 
- It seems unnecessary to scale down the signature image, since we will resize/maximise it when displaying it later.
- Try to keep the canvas aspects ratio to 400:150 (docusign recommended) in order to have consistency. 
- Move upload button and color selector above (out of) canvas
- Show a dashed line border of the drawing area

![Image](https://github.com/user-attachments/assets/37db0708-ecb0-45f5-85fb-c60ff18f5c48)

## Test
I have modified a version to test the effect. Happy to share if it make sense. 

![Image](https://github.com/user-attachments/assets/91533296-bb2b-410f-925d-be350519c122)

### Do you want to work on this improvement?

No

### Please check the boxes that apply to this improvement suggestion.

- [ ] I have searched the existing issues and improvement suggestions to avoid duplication.
- [ ] I have provided a clear description of the improvement being suggested.
- [ ] I have explained the rationale behind this improvement.
- [ ] I have included any relevant technical details or design suggestions.
- [ ] I understand that this is a suggestion and that there is no guarantee of implementation.

ElTimuro:
- Under Review
**2 Parts:**
- Moving the Buttons out of the field: Makes Sense
- General Render behaviour: Already changed after Editor V2

Mythie:
Rendering was updated a few months ago so they look nicer now and fit fields as best as they can rather than a generic width/height match.

For the design we'll take this under consideration but the design has changed since this issue was raised making things a bit cleaner than they were at the time.

PR #2751, opened 2026-04-30
fix: signature image sizing

sandavdesigns:
## Description

Fixes uploaded signature images appearing too small in signature fields while preserving the signature image aspect ratio.

The upload flow previously scaled uploaded signature images to 80% of the signature canvas, which added padding around the signature. In addition, signature images were not scaled up when rendered into signature fields, so smaller uploaded images could remain visibly undersized in the document preview and generated PDFs. After signing, the regular field container also kept its recipient-colored ring/background, which made signed image signatures look like they still had a colored border.

This keeps aspect ratio intact: the image grows until it reaches the first field edge. If the signature field has the same aspect ratio as the uploaded image, the image fills the field.

## Related Issue

Addresses #1701

## Changes Made

- Crop transparent or near-white whitespace around uploaded signature images before drawing them onto the signature canvas.
- Store the cropped uploaded signature image instead of the whole signature-pad canvas.
- Remove the 80% upload scaling so the signature can use the full available canvas area.
- Allow signature images to scale up proportionally to fit signature fields in the renderer and PDF insertion paths.
- Hide the field ring/background for already inserted signature fields so signed signatures do not keep the recipient-colored border in the signing UI.
- Apply the same PDF scaling behavior in both current and legacy PDF insertion code.

## Testing Performed

- Ran `git diff --check HEAD^ HEAD`.
- Verified the change is limited to signature image upload/rendering/PDF insertion logic and signed signature field display.

## Checklist

## Additional Notes

I opened this PR after hitting the issue in a self-hosted deployment where uploaded signature images were rendered much smaller than expected. The change is intentionally scoped to signature images and signed signature field display.

### DIFF
--- a/apps/remix/app/components/general/document-signing/document-signing-signature-field.tsx
+++ b/apps/remix/app/components/general/document-signing/document-signing-signature-field.tsx
@@ -254,7 +254,7 @@ export const DocumentSigningSignatureField = ({
         <img
           src={signature.signatureImageAsBase64}
           alt={`Signature for ${recipient.name}`}
-          className="h-full w-full object-contain"
+          className="h-full w-full"
         />
       )}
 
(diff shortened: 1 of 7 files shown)


Mythie:

https://documen.so/cla

PR #2782, opened 2026-05-09
fix: sanitize cmdk values to prevent DOMException with double quotes

### LINKED ISSUE #2779: Recipient names containing double-quote characters crash the signer/recipient picker (cmdk 0.2.1)
### Issue Description

## Bug                                                                                                                                                       
                  
  If any recipient's `name` contains a literal `"` character (e.g. nicknames written as `John "Killer" Doe`), the signer/recipient picker throws a         
  `DOMException` and the envelope editor renders the 500 error page.                                                                                                                    
                                                                       
                                                                                                                                                                   

### Steps to Reproduce

  ## Repro                                                                                                                                                     
                  
  1. Self-host `documenso/documenso:latest`.
  2. Create a recipient with a name like `John "Killer" Doe`.                                                                                              
  3. Open the envelope editor → try to add or edit a signer that opens the recipient picker.
  4. Page 500s. Browser console shows:                                                                                                                         
     DOMException: Element.querySelector: '[cmdk-item=""][data-value="John "Killer" Doe ..."]' is not a valid selector            

### Expected Behavior

  ## Suggested fix
                                                                                                                                                               
  Either:         

  1. **Bump `cmdk`** to a version where `data-value` is escaped via `CSS.escape()`, or                                                                         
  2. **Sanitize recipient names** at insert/update time, or escape on render before passing to `<CommandItem value={...}>`.
                                                                                                                                                               
  ## Workaround for self-hosters                                                                                                                               
                                                                                                                                                               
  ```sql                                                                                                                                                       
  UPDATE "Recipient" SET name = REPLACE(name, '"', '''') WHERE name LIKE '%"%';
  (Audit log entries are immutable, leave them alone.)                                                                                                         
                                                                                                                                                               
  ---                  

### Current Behavior

## Bug                                                                                                                                                       
                  
  If any recipient's `name` contains a literal `"` character (e.g. nicknames written as `John "Killer" Doe`), the signer/recipient picker throws a         
  `DOMException` and the envelope editor renders the 500 error page. 

  ## Cause                                                                                                                                                     
                  
  Documenso ships `cmdk@0.2.1`. That version interpolates `data-value` directly into a CSS attribute selector without escaping. Embedded `"` terminates the    
  selector early. Tracked upstream:
                                                                                                                                                               
  - https://github.com/dip/cmdk/issues/212 (regression introduced in 0.2.1)                                                                                    
  - https://github.com/dip/cmdk/issues/189
  - https://github.com/dip/cmdk/issues/387 (open; reportedly fixed on `cmdk@^1.x`)         

### Screenshots (optional)

_No response_

Linux

Firefox

### Version [e.g., 2.0.1]

0.2.1

ggoswami777:
## Problem
In `cmdk@0.2.1`, using a string with double quotes (like a recipient's name) as a `value` or text content for a `CommandItem` triggers a `DOMException`. This is because the library uses that value to build a CSS selector internally. In our Remix environment, this exception causes a complete app crash and a **500 Internal Server Error**.

## Solution
Following maintainer feedback, I have implemented a cleaner, centralized solution:
1. **Upgraded `cmdk` to `v1.1.1`**: Updated the dependency in `packages/ui`.
2. **Centralized Sanitization**: Since `cmdk v1.1.1` still exhibits the same CSS selector bug with double quotes, I moved the sanitization logic directly into the `CommandItem` primitive in `packages/ui/primitives/command.tsx`.
Now, any `CommandItem` used throughout the app will automatically strip double quotes from its `value` prop before passing it to the underlying library. This fixes the crash globally while keeping the code in individual components clean and readable.

## Before (Crash) 
image

## After (Fixed) 
image

## Key Changes
- **Dependency Update**: `cmdk` upgraded to `1.1.1`.
- **Core Primitive Update**: `CommandItem` now automatically handles quote sanitization for the `value` prop.
- **Improved Coverage**: Ensured `value` props are explicitly passed in critical areas (recipient selector, template flow) to trigger the automated fix.
## Verification
1. Created a recipient named `Test "Quote" User`.
2. Verified that opening the recipient selector no longer triggers a 500 error.
3. Verified that search still finds the user correctly.

Fixes #2779

### DIFF
--- a/apps/remix/app/components/general/envelope-editor/envelope-recipient-selector.tsx
+++ b/apps/remix/app/components/general/envelope-editor/envelope-recipient-selector.tsx
@@ -4,7 +4,14 @@ import { canRecipientFieldsBeModified } from '@documenso/lib/utils/recipients';
 import { getRecipientColorStyles } from '@documenso/ui/lib/recipient-colors';
 import { cn } from '@documenso/ui/lib/utils';
 import { Button } from '@documenso/ui/primitives/button';
-import { Command, CommandEmpty, CommandGroup, CommandInput, CommandItem } from '@documenso/ui/primitives/command';
+import {
+  Command,
+  CommandEmpty,
+  CommandGroup,
+  CommandInput,
+  CommandItem,
+  CommandList,
+} from '@documenso/ui/primitives/command';
 import { Popover, PopoverContent, PopoverTrigger } from '@documenso/ui/primitives/popover';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@documenso/ui/primitives/tooltip';
 import type { I18n } from '@lingui/core';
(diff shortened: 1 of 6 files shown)


dguyen:
I think we would prefer to just upgrade the package if that resolves it, arbitrarily modifying the input value might cause issues since it will cause the output value to also be modified (it might not matter here but it might in the future)

ggoswami777:
> I think we would prefer to just upgrade the package if that resolves it, arbitrarily modifying the input value might cause issues since it will cause the output value to also be modified (it might not matter here but it might in the future)

@dguyen  I upgraded cmdk to 1.1.1. However, I found that even in v1.1.1, the library still throws a DOMException if the value prop contains double quotes (as it uses them in a CSS selector internally).To handle this cleanly without manually modifying every usage, I've moved the sanitization into the CommandItem primitive itself.

hmesfin:
Tested this branch locally (built `documenso:pr-2782` from `docker/Dockerfile`, fresh Postgres, `NEXT_PUBLIC_UPLOAD_TRANSPORT=database`) against the original repro from #2779.

## ✅ Primary crash is fixed

Created recipients named `John "Killer" Doe` and `Jane "Ferocious" Doe` on an envelope. The signer editor loads cleanly — no `DOMException: not a valid selector`, no 500 page. The `value.replace(/"/g, '')` in `CommandItem` + cmdk@1.1.1 bump does its job. ✔️

## ⚠️ But: secondary mouse-click regression introduced by the v0.2.1 → v1.1.1 bump

In `apps/remix/app/components/general/envelope-editor/envelope-recipient-selector.tsx`, after the upgrade I observed:

- Opening the recipient picker, the default-selected signer (John) shows correctly highlighted.
- **Clicking another recipient (Jane) does nothing.** No state change, no console error, no `onSelect` firing.
- This persists after `UPDATE "Recipient" SET name = REPLACE(name, '"', '')` — so it is not caused by the sanitization patch in this PR, but by the cmdk version bump itself.

### Likely root cause

The `EnvelopeRecipientSelectorCommand` renders:

```tsx
<Command value={selectedRecipient ? selectedRecipient.id.toString() : undefined}>
  ...
  <CommandItem value={\`\${recipient.name} \${recipient.email}\`} onSelect={...}>
```

`Command.value` is set to the selected recipient's numeric ID (e.g. `"9"`), but no `CommandItem.value` ever contains that ID — item values are name+email strings. In cmdk@0.2.1 this mismatch was benign (mouse click still fired `onSelect`). In cmdk@1.1.1, `onSelect` only fires for the currently-active item, and the controlled-value mismatch appears to lock activation away from where the mouse is hovering.

### Suggested fixes (pick one)

1. Drop the controlled `value=` from `<Command>` here and let cmdk track activation natively from hover/keyboard, OR
2. Make item values agree with the controlled value: `value={recipient.id.toString()}` (and keep the visible label in `children` / a `getRecipientLabel` helper for the search filter via a separate `keywords` prop), OR
3. Audit every `<Command value=…>` consumer in the repo for the same controlled-value-doesn't-match-any-item pattern, since the same shape probably exists in other selectors (`recipient-selector.tsx`, `multiselect.tsx`, etc.).

Happy to test a follow-up commit if helpful. The crash fix itself is correct and worth landing regardless — the click regression is a separate cmdk-v1-migration concern.

ggoswami777:
@hmesfin 
I've updated the PR with a comprehensive fix that addresses both the crash and the selection regression.
Summary of changes:
1. Universal Sanitization: Modified packages/ui/primitives/command.tsx to automatically strip double quotes from value props in a memoized wrapper. This provides a global safety net against the CSS selector crash.
2. Click Fix: Removed pointer-events: none from the CommandItem primitive's disabled state. This was the primary reason mouse clicks weren't firing after the upgrade.
3. Component Audit: Updated envelope-recipient-selector.tsx, recipient-selector.tsx, and add-template-fields.tsx to include the mandatory <CommandList> wrapper and utilize dual onClick + onSelect handlers for maximum reliability.
4. Uncontrolled Root: Following the library's v1 patterns, I've moved to an uncontrolled Command root while maintaining stable, prefixed ID values in the items to prevent selection locking.

Verified across all flows (Envelope Editor and Template Fields). Ready for another look!

PR #2785, opened 2026-05-11
feat: add custom branding for signing pages

Mythie:
Platform-plan organisations and teams can now customise non-embed
signing pages with six brand colour tokens, a border-radius, and
a free-text custom CSS block (up to 256 KB).

- Stored on OrganisationGlobalSettings / TeamGlobalSettings;
  teams inherit from the org via brandingEnabled === null.
- CSS is sanitised on save (PostCSS) so we can inline it at SSR
  with no per-render parsing.
- Rendered via a nonce'd <style> scoped under .documenso-branded,
  using native CSS nesting so user selectors don't need scoping.
- Gated on the existing embedSigningWhiteLabel claim (or
  self-hosted) — reuses the embed white-label decision.

### DIFF
--- a/.agents/plans/happy-teal-tree-platform-signing-page-branding.md
+++ b/.agents/plans/happy-teal-tree-platform-signing-page-branding.md
@@ -1,0 +1,138 @@
+---
+date: 2026-05-06
+title: Platform Signing Page Branding
+---
+
+## What
+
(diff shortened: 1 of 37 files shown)


dguyen (approved):

rsalcara:
Excellent work @Mythie ! Now, the ability to import logos and favicons for customization would be important.
image

PR #2790, opened 2026-05-11
refactor: remove unnecessary DateRange type assertion

anish1204:
## Summary

Removes an unnecessary `as DateRange` type assertion since the `value`
parameter is already typed as `DateRange`.

## Changes

* Updated `handleRangeChange` parameter typing
* Removed redundant type assertion


dguyen (approved):
Thanks!

PR #2808, opened 2026-05-13
fix(web): handle duplicate organization URL update errors gracefully

anish1204:
## Description

In the Organisation section when we upate the url (which is like the UUID/slug) for the organisation if we try to fill the value with any other organisation url then we get an error as below

Screenshot (513)

Issue here is in frontend there is already a exception handling present if the value entered is already present then show that error however it could not get that error as from backend the error code coming is 500.

Also in the update-organisation there is this particular an edge case missing.

## Changes Made

Mapped duplicate organisation URL errors to AppErrorCode.ALREADY_EXISTS.
Improved error handling consistency between frontend and backend.

## Testing Performed

Tested updating an organisation URL to an already existing URL.
Verified that the frontend now shows the correct validation error message instead of an unhandled error.
Tested updating an organisation with a unique URL and confirmed successful updates still work correctly.

After Fix
image

## Checklist


anish1204:
@dguyen could you look into this PR i have explored a bug and fixed it accordingly.

dguyen (approved):
Thanks!

PR #2815, opened 2026-05-13
fix: trigger document webhooks before completed emails

### LINKED ISSUE #2814: Document completion webhooks should not depend on completed email delivery
### Problem

When the `internal.seal-document` job completes a document, it sends completed emails before triggering the `DOCUMENT_COMPLETED` webhook.

If the SMTP provider is unavailable, misconfigured, or blocked by the host network, the completed email task can fail before `triggerWebhook` runs. In that case API integrators do not receive the completion webhook even though the document has already been sealed and marked completed.

### Expected behavior

A completed document should trigger its completion webhook independently from completed email delivery. Email delivery failures should not prevent webhook dispatch.

### Suggested fix

Run webhook dispatch as an independent post-seal side effect, before or isolated from completed email delivery, so an email failure cannot block `DOCUMENT_COMPLETED` / `DOCUMENT_REJECTED` webhook delivery.

zahedshareef:
## Description

Closes #2814.

This updates the seal-document job so document completion/rejection webhooks are dispatched in their own cached job task before completed email delivery runs.

That means an SMTP outage, blocked SMTP port, or completed-email rendering/sending failure can no longer prevent `DOCUMENT_COMPLETED` / `DOCUMENT_REJECTED` webhooks from being queued after the document has already been sealed.

The completed-email logic is unchanged; it was moved into the same post-seal helper so the side-effect ordering is covered by tests.

## Testing

- `npm run test -w @documenso/lib`
- `npx biome check packages/lib/jobs/definitions/internal/seal-document.handler.ts packages/lib/jobs/definitions/internal/seal-document.post-seal.ts packages/lib/jobs/definitions/internal/seal-document.post-seal.test.ts`
- `npx tsc --noEmit --project packages/lib/tsconfig.json --pretty false` *(fails locally on existing Prisma type errors outside this change: `get-completed-fields-for-document.ts`, `get-active-subscriptions-by-user-id.ts`, `find-organisation-invoices.ts`)*

### DIFF
--- a/packages/lib/jobs/definitions/internal/seal-document.handler.ts
+++ b/packages/lib/jobs/definitions/internal/seal-document.handler.ts
@@ -8,31 +8,28 @@ import { prisma } from '@documenso/prisma';
 import { signPdf } from '@documenso/signing';
 import { PDF } from '@libpdf/core';
 import type { DocumentData, Envelope, EnvelopeItem, Field } from '@prisma/client';
-import { DocumentStatus, EnvelopeType, RecipientRole, SigningStatus, WebhookTriggerEvents } from '@prisma/client';
+import { DocumentStatus, EnvelopeType, RecipientRole, SigningStatus } from '@prisma/client';
 import { nanoid } from 'nanoid';
 import { groupBy } from 'remeda';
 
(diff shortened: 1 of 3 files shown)


dguyen:
Hi, we resolved this through another PR which delegates the email sending to a new job.

PR #2824, opened 2026-05-16
feat(remix): support serving app under a sub-path via NEXT_PUBLIC_BASE_PATH

GMCfourX4:
## Summary

Lets operators deploy Documenso behind a reverse proxy at a non-root path (e.g. `/ESign/`) without prefix stripping. Driven by a single new env var, `NEXT_PUBLIC_BASE_PATH`. When unset (the default), behaviour is unchanged.

## Why

Reverse proxies that strip prefixes (e.g. `/Signature/*` → `/*` on the way to the backend) work for HTML and most API calls, but break for SSR-rendered HTML containing root-absolute asset URLs (`/assets/app-xxx.css`) and for cookies/redirects the backend emits with leading-`/` paths. The cleanest fix is to make the app sub-path aware end-to-end rather than rely on proxy-side string surgery.

## What changed

A single env var, `NEXT_PUBLIC_BASE_PATH`, threaded through every layer that emits or consumes root-absolute URLs:

- **Build**: Vite `base`, React Router `basename`
- **Server**: Hono `.basePath()`, `serveStatic` rewriter, tRPC endpoints (internal + open-api). The internal tRPC adapter needs the prefix in `endpoint` because `@hono/trpc-server` slices it off the full URL to compute the procedure name; Hono's `basePath` alone doesn't rewrite the URL the adapter sees.
- **Runtime**: new `getBasePath()` helper in `packages/lib/constants/app.ts` (prefers explicit env, falls back to the pathname of `NEXT_PUBLIC_WEBAPP_URL`)
- **HTML**: favicon, apple-touch-icon, manifest links, `ThemeProvider` `themeAction`
- **Client fetches**: PDF upload, locale switch (×2 sites), AI field/recipient detection, embed presign-token
- **Auth**: sign-out path derived from `baseUrl`; post-auth redirects use a prefix-aware default; post-login `LOGIN_REDIRECT_PATH` derived from `getBasePath()`
- **PWA**: `site.webmanifest` icon URLs changed from absolute (`/foo.png`) to relative (`./foo.png`) so they resolve under any base path per the webmanifest spec (icons are parsed relative to the manifest URL)
- **Docker**: `NEXT_PUBLIC_BASE_PATH` build arg + env so the value is both baked into the client bundle at build time and exposed at runtime via `window.__ENV__`

20 files changed, +133/-31 lines.

## Behaviour when env var is unset

Every path is a no-op: `getBasePath()` returns `''`, Vite `base` is `/`, RR `basename` is `undefined`, Hono `basePath` is `/`, all ``fetch(`${''}/api/...`)`` calls collapse to `/api/...`. The default `site.webmanifest` icon resolution `./foo.png` relative to `/site.webmanifest` resolves to `/foo.png`, matching the previous absolute behaviour.

## Test plan

- [x] Production deployment with `NEXT_PUBLIC_BASE_PATH=/ESign` behind HAProxy forwarding `/ESign/*` → backend without prefix stripping
- [x] Verified end-to-end at the sub-path: SSR + hashed asset loading, sign-in/-up flow, theme + language switching, PDF upload, document signing, tRPC calls, OpenAPI v2 calls, PWA install
- [x] `npm run typecheck` clean against this branch
- [x] `docker build -f docker/Dockerfile .` (env var unset, validating the no-op default path) clean
- [ ] Would appreciate a maintainer running the existing e2e suite to confirm no regression at the default

Happy to split into multiple commits, change the env-var name, or restructure the helpers if you'd prefer.


Mythie:

https://documen.so/cla

GMCfourX4:
Pushed a follow-up commit covering base-path gaps that surfaced while running this branch in a real deployment under `/ESign` (behind HAProxy, no prefix stripping).

### Root-relative request

- `packages/lib/universal/upload/get-file.ts` — `/api/files/presigned-get-url`. This is the download counterpart to `put-file.ts`, which this PR already fixed, so S3-backed document retrieval was still 404ing.

### `new URL('/path', base)` discarding the sub-path

A leading-slash path replaces the base's pathname entirely, so `new URL('/signin', 'https://host/ESign')` yields `https://host/signin`:

- `packages/ee/server-only/limits/client.ts` — `/api/limits`
- `packages/auth/server/lib/utils/handle-oauth-callback-url.ts` — the three `/signin` error URLs

### Server-side redirects

Hono's `.basePath()` affects route matching but not `Location` headers, so these needed the prefix applied explicitly:

- `handle-oauth-callback-url.ts` — `redirectPath` fallbacks now default to `` `${getBasePath()}/` `` instead of `'/'`
- `handle-oauth-organisation-callback-url.ts` — `/o/:orgUrl`

### Client-side sign-out in the signing flow

- `document-signing-auth-page.tsx`, `document-signing-auth-account.tsx` — the `/signin` and `/signup` targets

### `is-valid-return-to.ts` — worth a closer look

Two defects here, and **the first is arguably a bug independent of this feature**:

```ts
if (returnToUrl.origin !== NEXT_PUBLIC_WEBAPP_URL()) {
```

This compares a URL **origin** against the **raw env value**. Any deployment whose `NEXT_PUBLIC_WEBAPP_URL` carries a path — including a trailing slash — can never satisfy it, so `isValidReturnTo` returns `false` for every `returnTo` and users always land on the default page after login instead of their intended destination. Now compares against the parsed origin.

Second, `normalizeReturnTo` returned the resolved pathname, which for a root-relative `returnTo` (`/inbox`) excludes the sub-path. It now re-applies the base path when missing and leaves already-prefixed paths untouched.

---

Formatted with the repo's pinned `@biomejs/biome@2.4.8` (`check --write`) over the changed files. Happy to split the `is-valid-return-to.ts` origin fix into its own PR if you'd rather land that independently of the sub-path work.

Mythie (approved):

PR #2835, opened 2026-05-20
fix: move document complete email to a job

dguyen:
## Description

Move the document completion email sending logic into it's own job. This allows us to send webhooks in the event that our email service is down.


roshboi (approved):
lgtm

Issue #2829, opened 2026-05-18
Faultmark scanned this repo and found 24 verified bugs


R0hit-Sriram:
Hey documenso team,

I've been building Faultmark (faultmark.com) — an AI scanner that runs a 3-model debate to verify bugs before surfacing them.
Ran it on this repo as a real-world test. Found 24 confirmed bugs. Sharing before posting anything publicly.

##  High Severity (6)

**1. envelope.recipients.sort() mutates React state in-place**
`apps/remix/app/components/general/document-signing/envelope-signing-provider.tsx:297`

The `nextRecipient` useMemo calls `envelope.recipients.sort(...)` directly, 
permanently mutating the React state object. After the first render, all 
downstream logic — including `selectedAssistantRecipient` and 
`assistantRecipients` — produces incorrect results. Only manifests with 
sequential signing enabled.

**Fix:** `[...envelope.recipients].sort(...)`

---

**2. allRecipients.sort() mutates the prop array, breaking signer lookup**
`apps/remix/app/components/general/document-signing/document-signing-page-view-v1.tsx:145`

Same mutation pattern. After the first render, the parent's array is permanently 
reordered, breaking `selectedSigner = allRecipients?.find(r => r.id === selectedSignerId)` 
and any field filtering that depends on original recipient order.

**Fix:** `[...allRecipients].sort(...)`

---

**3. handleInitialsFieldClick always submits null — user's typed initials are never saved**
`apps/remix/app/utils/field-signing/initial-field.ts:43`

When a user types initials into the dialog, the input is stored in `initialsToInsert`. 
But the return statement returns `value: initials` (the original null) instead of 
`value: initialsToInsert`. Every initials field submission silently sends null 
regardless of what the user typed.

**Fix:** Change line 43 from `value: initials` to `value: initialsToInsert`

---

**4. loadingSpinnerGroup.destroy() called twice on every DROPDOWN sign**
`apps/remix/app/components/general/envelope-signing/envelope-signer-page-renderer.tsx:338,341`

`destroy()` is called inside `.then()` at line 338 and again inside `.finally()` 
at line 341. Since `.finally()` always runs after `.then()`, the Konva group is 
destroyed twice on every successful dropdown sign. The second call either throws 
or corrupts canvas state.

**Fix:** Remove the `destroy()` at line 338, keep only the `.finally()` one.

---

**5. Admin organisations table pagination is permanently hidden**
`apps/remix/app/components/tables/admin-organisations-table.tsx:241`

[...]

PR #2838, opened 2026-05-20
fix: handleInitialsFieldClick now returns initialsToInsert instead of initials

Abzaek:
## Summary

Fixes bug #3 from the Faultmark scan (issue #2829): `handleInitialsFieldClick` always submitted `null` as the initials value because the return statement used `value: initials` (the original input parameter, always null after the dialog call) instead of `value: initialsToInsert` (the user's typed input).

### Root cause
The function stored the dialog result in `initialsToInsert` (line 31) but returned `value: initials` (line 43) — the original destructured parameter which was always null after the first successful dialog call.

### Fix
Changed line 43 from `value: initials` to `value: initialsToInsert`.

### Verification
- TypeScript compiles cleanly with `npx tsc --noEmit`

Closes ... (partially addresses #2829)

### DIFF
--- a/apps/remix/app/utils/field-signing/initial-field.ts
+++ b/apps/remix/app/utils/field-signing/initial-field.ts
@@ -40,6 +40,6 @@ export const handleInitialsFieldClick = async (
 
   return {
     type: FieldType.INITIALS,
-    value: initials,
+    value: initialsToInsert,
   };
 };


Mythie:

https://documen.so/cla

dguyen (approved):
Thanks!

PR #2839, opened 2026-05-20
fix: prevent React state mutation by spreading envelope.recipients before sort

Abzaek:
## Summary

Fixes bug #1 from the Faultmark scan (issue #2829): `envelope.recipients.sort()` was called directly on the React state object inside a `useMemo`, permanently mutating it in-place.

### Root cause
`const sortedRecipients = envelope.recipients.sort((a, b) => ...)` on line 297 sorted the array in-place, corrupting the React state. After the first render with sequential signing enabled, `selectedAssistantRecipient` and `assistantRecipients` would produce incorrect results.

### Fix
Changed to `[...envelope.recipients].sort(...)` — creating a shallow copy before sorting, leaving the original state array unmodified.

### Verification
- TypeScript compiles cleanly with `npx tsc --noEmit`

Partially addresses #2829


Mythie:

https://documen.so/cla

dguyen (approved):
Thanks

PR #2840, opened 2026-05-20
fix: prevent prop array mutation by spreading allRecipients before sort

Abzaek:
## Summary

Fixes bug #2 from the Faultmark scan (issue #2829): `allRecipients.sort()` was called directly on the prop array inside a `useMemo`, permanently reordering the parent's array.

### Root cause
`const sortedRecipients = allRecipients.sort((a, b) => ...)` on line 145 sorted the array in-place. After the first render, `selectedSigner = allRecipients?.find(r => r.id === selectedSignerId)` and field filtering that depended on original recipient order would break.

### Fix
Changed to `[...allRecipients].sort(...)` — creating a shallow copy before sorting, leaving the original prop array unmodified.

Partially addresses #2829

### DIFF
--- a/apps/remix/app/components/general/document-signing/document-signing-page-view-v1.tsx
+++ b/apps/remix/app/components/general/document-signing/document-signing-page-view-v1.tsx
@@ -142,7 +142,7 @@ export const DocumentSigningPageViewV1 = ({
       return undefined;
     }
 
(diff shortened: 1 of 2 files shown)


Mythie:

https://documen.so/cla

dguyen (approved):
Thanks

PR #2841, opened 2026-05-20
fix: remove duplicate loadingSpinnerGroup.destroy() in DROPDOWN sign

Abzaek:
## Summary

Fixes bug #4 from the Faultmark scan (issue #2829): `loadingSpinnerGroup.destroy()` was called both inside `.then()` (line 338) and `.finally()` (line 341). Since `.finally()` always runs after `.then()`, the Konva group was destroyed twice on every successful DROPDOWN sign.

### Fix
Removed the destroy() call inside `.then()` — the `.finally()` block handles cleanup in all cases (success or error).

Partially addresses #2829


Mythie:

https://documen.so/cla

dguyen (approved):

PR #2843, opened 2026-05-20
fix: prevent crash when removing last dropdown option in removeValue

Abzaek:
## Summary

Fixes bug #6 from the Faultmark scan (issue #2829): `removeValue` crashes when removing the last item from the dropdown options list.

### Root cause
After `newValues.splice(index, 1)`, the code checked `newValues[index].value`. But `newValues[index]` is the element that shifted into that position after the splice. If the removed item was the last one, `newValues[index]` is `undefined` -> runtime crash.

### Fix
Capture the `removedValue` from `currentValues[index].value` before the splice, then compare `defaultValue` against the captured value.

Partially addresses #2829


Mythie:

https://documen.so/cla

dguyen (approved):
Thanks

PR #2845, opened 2026-05-21
feat: configure audit log embedding per document

ephraimduncan:
## Summary
- add `DocumentMeta.includeAuditLog` with migration and API/editor schema support
- derive the value from team/org settings at document creation time, with per-document overrides
- use document meta during sealing so audit-log embedding is stable after creation
- add E2E coverage for defaults, overrides, and separate audit-log downloads

## Verification
- `npm run prisma:generate`
- `NEXT_PRIVATE_DATABASE_URL=postgresql://user:pass@localhost:5432/db NEXT_PRIVATE_DIRECT_DATABASE_URL=postgresql://user:pass@localhost:5432/db npx prisma validate --schema packages/prisma/schema.prisma`
- `npx biome check --write <changed files>`
- `npx tsc --noEmit -p packages/trpc/tsconfig.json`
- `git diff --check`

## Not run
- Targeted E2E spec could not run because the local database at `127.0.0.1:54320` was unavailable.

Refs documenso/backlog-internal#646


ephraimduncan:
Closed in favor of #2846 with a clearer branch name.

PR #2847, opened 2026-05-21
feat: configure audit log embedding per document

ephraimduncan:
## Summary
- add `DocumentMeta.includeAuditLog` with migration and API/editor schema support
- derive the value from team/org settings at document creation time, with per-document overrides
- use document meta during sealing so audit-log embedding is stable after creation
- add E2E coverage for defaults, overrides, and separate audit-log downloads


packages/lib/server-only/envelope/update-envelope.ts
```diff
@@ -266,36 +290,6 @@ export const updateEnvelope = async ({
...
-  //   auditLogs.length === 0 &&
-  //   data.useLegacyFieldInsertion === undefined &&
-  //   isFolderSame &&
-  //   isTemplateTypeSame &&
-  //   isPublicDescriptionSame &&
-  //   isPublicTitleSame
-  // ) {
-  //   return envelope;
-  // }
-
```
dguyen, line 317: leave this

packages/prisma/schema.prisma
```diff
@@ -524,6 +524,7 @@ model DocumentMeta {
 
   language           String                     @default("en")
   distributionMethod DocumentDistributionMethod @default(EMAIL)
+  includeAuditLog    Boolean                    @default(false)
```
dguyen, line 576: Won't this break existing flows that expect audit logs to be included

packages/lib/server-only/document-meta/upsert-document-meta.ts
```diff
@@ -91,6 +99,28 @@ export const updateDocumentMeta = async ({
...
+        sendStatus: SendStatus.SENT,
+      },
+    });
+
+    if (sentRecipientCount > 0) {
+      throw new AppError(AppErrorCode.INVALID_BODY, {
+        message: 'Audit log embedding can only be changed before the document is sent',
+      });
+    }
+  }
```
dguyen: Is there a reason why we're preventing allowing this to be changed?

dguyen (changes requested):

packages/prisma/schema.prisma
ephraimduncan: It won't. The migration backfills the column from the team/org setting, so existing documents keep their current behavior. New documents get the value from settings at creation via `extractDerivedDocumentMeta`. The `false` default only exists so the column can be non-null.

packages/lib/server-only/document-meta/upsert-document-meta.ts
ephraimduncan: No real reason, I've removed the guard. Sealing reads the value from document meta at seal time anyway, so changing it after sending still works.

packages/lib/server-only/envelope/update-envelope.ts
ephraimduncan: Restored.

PR #2856, opened 2026-05-23
fix: delete dialog flash-closes when triggered from documents table dropdown

### LINKED ISSUE #2813: Bug: Delete dialog flash-closes when triggered from documents table dropdown menu (v2.10.1)
## Bug Description

When clicking the "⋯" (three-dot) action button on a document row in the main `/documents` table and selecting **Delete**, the delete confirmation dialog appears for a single frame and then immediately closes. The dialog is unusable from this entry point.

## Steps to Reproduce

1. Navigate to `/documents`
2. Click the three-dot action button on any document row
3. Click **Delete**
4. Observe: the AlertDialog momentarily flashes onto the screen, then immediately disappears
5. The dropdown menu also closes
6. No document is deleted

## Expected Behavior

The delete confirmation dialog should stay open until the user clicks **Cancel** or **Delete** (or presses Escape).

## Actual Behavior

Dialog renders for ~1 frame then unmounts. The user has no opportunity to confirm.

## Environment

- **Documenso version:** v2.10.1 (self-hosted Docker)
- **Browser:** Chrome 148 on macOS
- **OS:** Ubuntu 22.04 LTS host
- **Deployment:** official `documenso/documenso` container behind Caddy reverse proxy

## Technical Notes

I inspected the deployed bundle (`documents-table-action-dropdown-Djg2PvBM.js`) and the delete-dialog component (in `envelope-duplicate-dialog-DmKhacRr.js`). The dedupe refactor from #2686 is present:

- The dropdown item that triggers the delete dialog correctly uses `onSelect: (e) => e.preventDefault()` to prevent the DropdownMenu from auto-closing.
- The dialog is `EnvelopeDeleteDialog` (consolidated, not the old separate `DocumentDeleteDialog`).
- The dialog controls its own `open` state internally via `useState`, with `onOpenChange: (v) => !isPending && setOpen(v)`.
- The structure is: `<Dialog><DialogTrigger asChild><DropdownMenuItem asChild onSelect={preventDefault}>…</DropdownMenuItem></DialogTrigger><DialogContent/></Dialog>` — nested inside `DropdownMenuContent`.

Despite the correct Radix pattern, the dialog still unmounts immediately. Suspect root causes (in priority order):

1. **Double `asChild` slot collision** — `Dialog.Trigger asChild` wrapping `DropdownMenu.Item asChild` may interfere with ref/event forwarding in some Radix versions.
2. **Portal/focus race** — when the DropdownMenu starts its close-and-cleanup cycle (focus restoration, pointer-events reset), the dialog inside it sees an immediate `onOpenChange(false)` from Radix's outside-click detection.
3. **`pointer-events: none` stuck on `<body>`** — known Radix issue when a Dialog opens via a DropdownMenu trigger in the same render cycle.

Notable: the **duplicate dialog** in the same dropdown uses the exact same pattern; if it works there but not here, the asymmetry suggests a subtle state-management or focus-trap bug specific to `EnvelopeDeleteDialog` rather than the trigger plumbing itself.

## Workaround

Open the document detail page and delete from there — the detail-page dropdown is a different code path and unaffected.

## Related

- #2335 (closed) — share dialog had the same symptom
- #2686 (merged, v2.9.0) — fixed the share-dialog version and consolidated delete/duplicate dialogs
- v2.11.0 release notes do not mention a follow-up fix

Happy to bisect or provide a minimal repro repo if helpful.

xonaib:
Closes #2813

## What changed

The delete dialog in the documents table action dropdown was flash-closing immediately when opened.

**Root cause:** The classic Radix UI `Dialog`-inside-`DropdownMenuContent` focus-trap collision. `EnvelopeDeleteDialog` rendered as a child of `<DropdownMenuContent>` caused both components to compete for focus on mount, making the dialog close before the user could interact with it.

**Fix:** Two-part change matching the pattern already used by `EnvelopeSaveAsTemplateDialog` and `EnvelopeRenameDialog` in the same component:

1. **`EnvelopeDeleteDialog`** — Added optional controlled `open` / `onOpenChange` props. Uncontrolled/trigger-prop usage still works for other callers.

2. **`DocumentsTableActionDropdown`** — Moved `<EnvelopeDeleteDialog>` outside `<DropdownMenuContent>` as a sibling, driven by `isDeleteDialogOpen` state. The `<DropdownMenuItem>` uses a plain `onClick` instead of the trigger-prop pattern.

## How to test

1. Go to `/documents`
2. Open the three-dot action menu on any document
3. Click **Delete** (or **Hide**)
4. The delete confirmation dialog should open and stay open ✅

Previously it would flash for one frame and immediately dismiss.

### DIFF
--- a/apps/remix/app/components/dialogs/envelope-delete-dialog.tsx
+++ b/apps/remix/app/components/dialogs/envelope-delete-dialog.tsx
@@ -24,6 +24,8 @@ type EnvelopeDeleteDialogProps = {
   id: string;
   type: EnvelopeType;
   trigger?: React.ReactNode;
+  open?: boolean;
+  onOpenChange?: (open: boolean) => void;
   onDelete?: () => Promise<void> | void;
   status: DocumentStatus;
   title: string;
@@ -34,6 +36,8 @@ export const EnvelopeDeleteDialog = ({
   id,
   type,
   trigger,
+  open: controlledOpen,
+  onOpenChange: controlledOnOpenChange,
   onDelete,
   status,
   title,
@@ -45,11 +49,20 @@ export const EnvelopeDeleteDialog = ({
 
   const deleteMessage = msg`delete`;
 
-  const [open, setOpen] = useState(false);
+  const [internalOpen, setInternalOpen] = useState(false);
   const [inputValue, setInputValue] = useState('');
   const [isDeleteEnabled, setIsDeleteEnabled] = useState(status === DocumentStatus.DRAFT);
 
   const isDocument = type === EnvelopeType.DOCUMENT;
+
+  // Support both controlled (open/onOpenChange props) and uncontrolled (trigger prop) usage.
+  const open = controlledOpen ?? internalOpen;
+
+  const onOpenChange = (value: boolean) => {
+    if (isPending) return;
+    setInternalOpen(value);
+    controlledOnOpenChange?.(value);
+  };
 
   const { mutateAsync: deleteEnvelope, isPending } = trpcReact.envelope.delete.useMutation({
     onSuccess: async () => {
@@ -71,7 +84,7 @@ export const EnvelopeDeleteDialog = ({
 
       await onDelete?.();
 
-      setOpen(false);
+      onOpenChange(false);
     },
     onError: () => {
       toast({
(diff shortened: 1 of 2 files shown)


Mythie:

https://documen.so/cla

dguyen:
I can't replicate this on MacOS using Chrome/Firefox/Safari

Do you have a video of this happening? What system/browser are you running?

xonaib:
I tried reproducing this again locally on Windows against "main", but I’m no longer able to reproduce it either. The delete confirmation dialog stays open for me now.
Given that, I’m going to re-check whether this PR is still needed or whether the issue was environment/state-specific. The change is still structurally consistent with the existing rename/save-as-template dialog pattern, but I don’t want to claim a repro I can’t currently demonstrate.

dguyen:
Thanks, I think this might have been resolved recently in another PR. I can't remember though.

PR #2868, opened 2026-05-26
fix: use functional update in setSearchParams to prevent infinite re-render loop

durgeshhhhhhh:
## Description
When `searchParams` is included in the `useEffect` dependency array, each `setSearchParams` call produces a new `searchParams` object reference, re-triggering the effect and causing an infinite re-render loop.

The issue suggested an early return guard as a fix, but the effect would still run unnecessarily on every `searchParams` change. Instead, I used the functional update pattern `setSearchParams((prev) => {...})` which removes `searchParams` from the dependency array entirely, breaking the loop at the root cause.

## Related Issue

Part of #2829

## Changes Made

- Replaced direct `searchParams` usage inside `useEffect` with functional update pattern `setSearchParams((prev) => {...})`
- Removed `searchParams` from the `useEffect` dependency array
- Used `prev` instead of `searchParams` to build updated params, ensuring 
  always-fresh value without creating a dependency


dguyen:
Hi, do you have steps to replicate this issue?

durgeshhhhhhh:
Hi @dguyen, Here are the steps to replicate:

1. Navigate to `/o/{org-url}/settings/teams`
2. Open React DevTools → Profiler tab
3. Start recording
4. Type "hello" in the search input
5. Stop recording

**Without fix:** 65+ re-renders observed for 5 keystrokes
**With fix:** ~16 re-renders for same input

**Root Cause:**

searchParams is included in the useEffect dependency array. Each setSearchParams call produces a new searchParams object reference, re-triggering the effect in a loop.

**Buggy code:**
```tsx
useEffect(() => {
  const params = new URLSearchParams(searchParams?.toString());
  params.set('query', debouncedSearchQuery);
  if (debouncedSearchQuery === '') {
    params.delete('query');
  }
  setSearchParams(params);
}, [debouncedSearchQuery, pathname, searchParams]);

```

I also checked the Sibling files and found they have an early return guard to prevent this, but this file was missing it, causing an infinite loop.

I have two possible fixes :

1. Early return guard (used in sibling files):
```tsx
if (params.toString() === searchParams?.toString()) return;
```
This stops the loop, but the effect still reruns unnecessarily on every searchParams change, resulting in 2 renders per keystroke.

2. Functional update pattern (My fix):

```tsx
useEffect(() => {
  setSearchParams((prev) => {
    const params = new URLSearchParams(prev?.toString());
    if (debouncedSearchQuery === '') {
      params.delete('query');
    } else {
      params.set('query', debouncedSearchQuery);
    }
    return params;
  });
}, [debouncedSearchQuery, pathname]);
```

I chose option 2 because prev always provides a fresh value directly from React, eliminating the need for searchParams in the dependency array entirely and preventing the stale closure issue. Effect only runs when debouncedSearchQuery or pathname changes, breaking the loop at the root cause and keeping renders minimal, making the application more efficient.

durgeshhhhhhh:
HI @dguyen, if you get a chance, review the PR and let me know if anything needs to be changed. Also, if you approve this approach, I would like to discuss applying the same functional update pattern to the sibling files as well for consistency and efficiency

Mythie (approved):

Mythie:
Cheers

durgeshhhhhhh:
Hey @Mythie ,

First of all, cheers brother

And thanks! I’m really glad you understood my approach and merged the PR. I really appreciate the review and the opportunity to contribute!

PR #2873, opened 2026-05-27
feat: allow changing field types

dguyen:
## Description

Allow changing the field types in the envelope editor without deleting and readding it.

Notes:
- Not planning to migrate partial field meta across to the new fields


roshboi (approved):
lgtm

PR #2876, opened 2026-05-27
feat: enhance document signing page with field canvas style integration

catalinpit:
This PR adds a hidden DOM “style sample” on the signing page, reads embed CSS from it, and paints Konva fields with those computed styles.


apps/remix/app/components/general/document-signing/document-signing-page-view-v2.tsx
```diff
@@ -262,12 +263,31 @@ export const DocumentSigningPageViewV2 = () => {
...
+                    data-readonly="false"
+                    style={{
+                      position: 'absolute',
+                      width: 0,
+                      height: 0,
+                      overflow: 'hidden',
+                      pointerEvents: 'none',
+                      visibility: 'hidden',
+                    }}
+                  />
```
dguyen: Sorry I'm not familiar with the requirements of this feature

How come we're not reading the CSS values they put in 

Image

catalinpit: We are reading the values they put in.

The custom CSS is injected on the signing page. The hidden probe element has the same field class/data attributes that customer CSS targets. For each field, it sets the probe attrs to match that field and call `window.getComputedStyle(probe)`, then map the resolved `background`, `border`, `radius`, and `opacity` values into the Konva canvas field.

Mythie (approved):

Issue #2878, opened 2026-05-28
Security: refresh production Docker image dependencies flagged by Trivy/Grype


luminous11111:
Hi Documenso maintainers,

We are evaluating Documenso Community Edition for a self-hosted legal document signing workflow and ran a supply-chain review against the current stable image:

- Image: docker.io/documenso/documenso:v2.11.0 (https://hub.docker.com/r/documenso/documenso)
- Manifest digest: sha256:d9b9a21841e28ebf08d747706e15319a485aa19afea09ab1c926b025c2ce4a18
- Scanners:
  - Trivy 0.70.0
  - Grype 0.112.0

Scanner summary from our review:
- Trivy: 1 critical, 39 high, 38 medium, 3 low
- Grype: 2 critical, 37 high, 51 medium, 3 low

Representative findings include:
- Critical findings located at /app/node_modules/esbuild/bin/esbuild
- axios 1.15.0 with fixed versions reported around 1.15.1/1.15.2
- protobufjs 7.5.5 with fixed version 7.5.6
- picomatch 4.0.3 with fixed version 4.0.4
- fast-uri 3.1.0 with fixed versions 3.1.1/3.1.2
- kysely 0.28.16 with fixed version 0.28.17
- @rvf/set-get 7.0.1 with fixed version 7.0.2
- OpenTelemetry package findings around 0.217.0 / 0.75.0

Questions:
1. Are these packages/binaries expected to be present in the production runtime image?
2. Is the bundled esbuild binary ever executed at runtime, or is it build-time residue?
3. Would the project accept a dependency/runtime refresh PR for the fixed versions?
4. Is there a planned patched release/image after dependency updates?
5. Would you consider adding or reopening a SECURITY.md/security reporting path?

We are not reporting an exploit and are not claiming these findings are all runtime-reachable. We are treating them as a production-runtime gate because Documenso would process sensitive legal documents and signer workflows in our use case.

Happy to help test a patched image digest with Trivy/Grype once available.

Priyanshu31102003:
Hey Documenso team, 

Ensuring the production Docker image is clear of critical supply-chain vulnerabilities is highly important for self-hosted enterprise setups. Looking closely at the scanner logs, upgrading these sub-dependencies (`axios`, `protobufjs`, `kysely`, and `@rvf/set-get`) to their respective patched versions will resolve these flags without breaking core compatibility.

[...]

PR #2905, opened 2026-06-01
fix(docker): production hardening and dependency refresh (#2878)

winnerineast:
This PR resolves Issue #2878 by implementing comprehensive security hardening for the production Docker image and upgrading vulnerable transitive dependencies.

### Key Changes

1. **Dependency Upgrades (Supply Chain Security)**:
   - Configured global npm `overrides` in the root `package.json` for transitive dependencies flagged in Trivy/Grype audits, including `axios`, `picomatch`, `ws`, `fast-uri`, `kysely`, `@rvf/set-get`, `@opentelemetry/api`, `path-to-regexp`, `semver`, and `cross-spawn`.

2. **Dockerfile Hardening & Layer Security**:
   - **No Secrets in Build**: Eliminated all default hardcoded ARG/ENV secret fallbacks (`CAFEBABE`/`DEADBEEF`) to prevent credential leakage via OCI layer history (`docker history`).
   - **Dedicated `prod-deps` Stage**: Runs `npm ci --omit=dev --ignore-scripts` followed by `npm prune --omit=dev` to ensure development dependencies never reach the final runtime.
   - **Binary Pruning**: Explicitly removed `esbuild`, `typescript`, `tsx`, and related native executables from the production `node_modules` path inside the image, removing all compiler/dev residues.
   - **Prisma Integration**: Moved Prisma Client generation (`npx prisma generate`) to the `prod-deps` stage before directory locking, then removed the source Prisma files.
   - **Directory Lock-Down**: Applied `chmod -R a-w` to production `node_modules` inside the container to prevent dynamic runtime file modifications.
   - **Minimal Runner Stage**: Adapted a strict allowlist-only `COPY` in the final `runner` stage, excluding migrations, source files (`.ts`/`.tsx`), and any tooling.

### Verification

- Successfully built the container image locally.
- Verified that all executable `esbuild` and development binaries have been 100% eradicated from `/app/node_modules/`.
- Verified that all target transitives are locked to their non-vulnerable versions in the lockfile.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -89,12 +89,14 @@ COPY --from=builder /app/patches ./patches
 # Copy schema.prisma to generate client before making node_modules read-only
 COPY --from=installer /app/packages/prisma/schema.prisma ./packages/prisma/schema.prisma
 
-# Install production dependencies, generate prisma client, and make node_modules read-only
+# Install production dependencies, generate prisma client, clean up dev binaries, and make node_modules read-only
 RUN npm ci --omit=dev --ignore-scripts \
     && npm prune --omit=dev \
     && npx prisma generate --schema ./packages/prisma/schema.prisma \
     && rm -rf ./packages/prisma/schema.prisma \
+    && rm -rf /app/node_modules/esbuild /app/node_modules/@esbuild /app/node_modules/typescript /app/node_modules/tsx \
     && chmod -R a-w /app/node_modules
+
 
 ###########################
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

Mythie:
- We update deps regularly and those are rolled into new releases
- By omitting migrations you're breaking the container contract for most users since migrations on startup is expected
- Tightening the final image to ensure there are no straggling dev deps is good though

PR #2907, opened 2026-06-01
fix(signing): enforce drawn/uploaded signature toggles on the server

abhibalu:
## Summary

A document author can disable each of the three signature-input modes independently via `documentMeta.typedSignatureEnabled`, `drawSignatureEnabled`, and `uploadSignatureEnabled`. The UI hides the inputs accordingly, but only the **typed** mode was enforced server-side. A direct API caller (or anyone bypassing the UI) could still submit a drawn or uploaded signature for a document where the author had disabled those modes, silently bypassing the author's configured policy.

## The gap

Two server-side signing paths existed, each checking only `typedSignatureEnabled`:

- `packages/lib/server-only/field/sign-field-with-token.ts:207` (V1 signing path)
- `packages/lib/utils/envelope-signing.ts:241` (V2 signing path)

A grep across `packages/lib/server-only/field` and `packages/lib/server-only/document` for `drawSignatureEnabled` / `uploadSignatureEnabled` returned no enforcement points.

## Fix

Introduce a small pure helper, `assertSignatureModeAllowed`, in `packages/lib/utils/signature-mode.ts`, and call it from both signing paths.

**Design note on drawn vs. uploaded:** the server cannot distinguish the two — both arrive as a `data:image/png;base64,…` data URL (see `packages/lib/constants/signatures.ts:4`). The helper therefore treats the `image` mode as one category and rejects it only when **both** `drawSignatureEnabled` and `uploadSignatureEnabled` are explicitly `false`. If either is enabled, image signatures continue to be accepted. This matches the existing tri-state convention where `null` / `undefined` is treated as "enabled".

The typed-mode path is unchanged in behaviour; it now flows through the same helper for consistency. Rejections raise `AppError(INVALID_BODY)` so API clients receive a structured 400 rather than an opaque 500.

## Files changed

| File | Change |
|---|---|
| `packages/lib/utils/signature-mode.ts` | new — pure helper |
| `packages/lib/utils/signature-mode.test.ts` | new — 12 unit tests |
| `packages/lib/server-only/field/sign-field-with-token.ts` | route both modes through the helper |
| `packages/lib/utils/envelope-signing.ts` | same; extend the `documentMeta` `Pick` to expose the two new flags |

## Test plan

- [x] `npx vitest run` in `packages/lib` → **121 passing, 0 failing** (12 new + 109 pre-existing)
- [x] `biome check` on touched files → clean
- [x] `tsc --noEmit` → no new errors on touched files
- [ ] Manual repro in the dev app: create a document with `drawSignatureEnabled: false` **and** `uploadSignatureEnabled: false`, attempt to sign with a drawn signature via the API/tRPC, expect a 400 with the new message. *(Not run by me; maintainer or reviewer to verify in a running env.)*

## Out of scope / follow-ups

- No e2e (Playwright) coverage added. The existing e2e suite under `packages/app-tests/e2e` does not appear to exercise the draw/upload-disabled paths; happy to add one in a follow-up if desired.
- The same tri-state rule (only `=== false` disables) is preserved. If the project would prefer "any falsy disables", that is a separate behaviour change and should be discussed before being made.

## Notes for the maintainer

This is my first contribution here — please tell me if any of the following should change:
- commit / PR title format
- where the helper lives (currently colocated in `packages/lib/utils/` alongside `sanitize-branding-css.ts` etc., matching the existing test-file colocation pattern)
- whether the typed-mode rejection should also migrate from `throw new Error(...)` to `AppError` for consistency (kept the original `throw` to minimise surface area of this PR)

### DIFF
--- a/packages/lib/server-only/field/sign-field-with-token.ts
+++ b/packages/lib/server-only/field/sign-field-with-token.ts
@@ -25,6 +25,7 @@ import {
 import type { RequestMetadata } from '../../universal/extract-request-metadata';
 import { createDocumentAuditLogData } from '../../utils/document-audit-logs';
 import { assertRecipientNotExpired } from '../../utils/recipients';
+import { assertSignatureModeAllowed } from '../../utils/signature-mode';
 import { validateFieldAuth } from '../document/validate-field-auth';
 
 export type SignFieldWithTokenOptions = {
@@ -204,8 +205,12 @@ export const signFieldWithToken = async ({
     throw new Error('Signature field must have a signature');
   }
 
-  if (isSignatureField && documentMeta?.typedSignatureEnabled === false && typedSignature) {
-    throw new Error('Typed signatures are not allowed. Please draw your signature');
+  if (isSignatureField && documentMeta) {
+    if (typedSignature) {
+      assertSignatureModeAllowed('typed', documentMeta);
+    } else if (signatureImageAsBase64) {
+      assertSignatureModeAllowed('image', documentMeta);
+    }
   }
 
   if (field.fieldMeta?.readOnly && !AUTO_SIGNABLE_FIELD_TYPES.includes(field.type)) {
--- a/packages/lib/utils/envelope-signing.ts
+++ b/packages/lib/utils/envelope-signing.ts
@@ -15,6 +15,7 @@ import {
   ZTextFieldMeta,
 } from '@documenso/lib/types/field-meta';
 import { toCheckboxCustomText, toRadioCustomText } from '@documenso/lib/utils/fields';
+import { assertSignatureModeAllowed } from '@documenso/lib/utils/signature-mode';
 import { zEmail } from '@documenso/lib/utils/zod';
 import type { TSignEnvelopeFieldValue } from '@documenso/trpc/server/envelope-router/sign-envelope-field.types';
 import { checkboxValidationSigns } from '@documenso/ui/primitives/document-flow/field-items-advanced-settings/constants';
@@ -27,7 +28,10 @@ import { z } from 'zod';
 export type ExtractFieldInsertionValuesOptions = {
   fieldValue: TSignEnvelopeFieldValue;
   field: Field;
-  documentMeta: Pick<TDocumentMeta, 'timezone' | 'dateFormat' | 'typedSignatureEnabled'>;
+  documentMeta: Pick<
+    TDocumentMeta,
+    'timezone' | 'dateFormat' | 'typedSignatureEnabled' | 'drawSignatureEnabled' | 'uploadSignatureEnabled'
+  >;
 };
 
 export const extractFieldInsertionValues = ({
@@ -238,11 +242,7 @@ export const extractFieldInsertionValues = ({
 
       const isBase64 = isBase64Image(value);
 
-      if (documentMeta.typedSignatureEnabled === false && !isBase64) {
-        throw new AppError(AppErrorCode.INVALID_BODY, {
-          message: 'Typed signatures are not allowed. Please draw your signature',
-        });
-      }
+      assertSignatureModeAllowed(isBase64 ? 'image' : 'typed', documentMeta);
 
       return {
         customText: '',
--- a/packages/lib/utils/signature-mode.test.ts
+++ b/packages/lib/utils/signature-mode.test.ts
@@ -1,0 +1,89 @@
+import { describe, expect, it } from 'vitest';
+
+import { assertSignatureModeAllowed, isSignatureModeAllowed } from './signature-mode';
+
+describe('isSignatureModeAllowed', () => {
+  describe('typed signatures', () => {
+    it('is allowed when typedSignatureEnabled is true', () => {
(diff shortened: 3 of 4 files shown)


gauravbsinghal:
Hi @abhibalu, I'm sharing this from [Niro](https://github.com/apxlabs-ai/niro), a PR pentesting tool we're building to help OSS maintainers catch security regressions before merge. I ran it against this PR and it found the following issue.

A recipient with a valid signing token can complete an envelope with an image signature despite the author disabling draw or upload, defeating that signature-mode control when the other image mode remains enabled.

## Reproduction Steps

Precondition: use `Authorization: Bearer $CR_005_TOKEN` for the authoring API and `/workspace/artifacts/source/assets/example.pdf` as the uploaded file.

1. Create an envelope with drawing disabled but uploads enabled:
   - `POST http://host.docker.internal:3000/api/v2-beta/envelope/create`
   - Header: `Authorization: Bearer $CR_005_TOKEN`
   - Multipart field `payload`:
     `{"title":"tc-sigmode-draw-disabled","type":"DOCUMENT","recipients":[{"email":"tc-sigmode-draw-disabled@example.com","name":"Recipient","role":"SIGNER","accessAuth":[],"actionAuth":[],"signingOrder":1,"fields":[{"type":"SIGNATURE","identifier":0,"page":1,"positionX":5,"positionY":5,"width":15,"height":5,"fieldMeta":{"type":"signature"}}]}],"meta":{"distributionMethod":"NONE","typedSignatureEnabled":true,"drawSignatureEnabled":false,"uploadSignatureEnabled":true}}`
   - File part: `files=@/workspace/artifacts/source/assets/example.pdf;type=application/pdf`
   - Response was `200 {"id":"envelope_dnnvhrfycayehkkb"}`.

2. Fetch the created field ID and recipient token:
   - `GET http://host.docker.internal:3000/api/v2-beta/envelope/envelope_dnnvhrfycayehkkb` with the same Authorization header returned field id `5`.
   - `POST http://host.docker.internal:3000/api/v2-beta/envelope/distribute` with JSON `{"envelopeId":"envelope_dnnvhrfycayehkkb"}` returned recipient token `l0p7CJPBrpLyFy103buUM`.

3. Submit an image signature through the recipient signing API:
   - `POST http://host.docker.internal:3000/api/trpc/envelope.field.sign`
   - Header: `Content-Type: application/json`
   - Body: `{"json":{"token":"l0p7CJPBrpLyFy103buUM","fieldId":5,"fieldValue":{"type":"SIGNATURE","value":"data:image/png;base64,iVBORw0KGgo="}}}`
   - Expected secure behavior: reject with HTTP 400 because the author set `drawSignatureEnabled:false`.
   - Observed behavior: HTTP 200 and `signedField.inserted:true`; the response stored `signature.signatureImageAsBase64:"data:image/png;base64,iVBORw0KGgo="` for envelope `envelope_dnnvhrfycayehkkb`.

4. Repeat the same attack with uploads disabled but drawing enabled:
   - Created `envelope_tcewssvmbixveezv` with `{"typedSignatureEnabled":true,"drawSignatureEnabled":true,"uploadSignatureEnabled":false,"distributionMethod":"NONE"}`.
   - Distributed recipient token: `TcSaixv-ejfPf8OB1vgR1`; field id: `6`.
   - `POST /api/trpc/envelope.field.sign` body `{"json":{"token":"TcSaixv-ejfPf8OB1vgR1","fieldId":6,"fieldValue":{"type":"SIGNATURE","value":"data:image/png;base64,iVBORw0KGgo="}}}` returned HTTP 200 with `signedField.inserted:true` and `signature.signatureImageAsBase64:"data:image/png;base64,iVBORw0KGgo="`.

5. Control check:
   - Created `envelope_wkzdzyeueromldyn` with both `drawSignatureEnabled:false` and `uploadSignatureEnabled:false`.
   - The same signing request against field id `7` and token `kn_qt6LbRDQB8ILr2KUGE` returned HTTP 400 `INVALID_BODY` with message `Drawn or uploaded signatures are not allowed. Please type your signature`.

This reproduces twice and shows the server only blocks image signatures when both image-related toggles are false, not when either disabled mode should be enforced independently.

## Remediation Guidance

The current server contract cannot enforce the author's independent draw/upload toggles because `packages/lib/utils/signature-mode.ts` collapses both into one `'image'` branch (`drawSignatureEnabled !== false || uploadSignatureEnabled !== false`), and both signing paths call `assertSignatureModeAllowed('image', ...)`. Fix this at the request/schema boundary: carry the actual signature source as `'draw' | 'upload' | 'typed'` in the signature submission payload, then update `assertSignatureModeAllowed` to check `drawSignatureEnabled` and `uploadSignatureEnabled` separately. In `packages/lib/server-only/field/sign-field-with-token.ts` and `packages/lib/utils/envelope-signing.ts`, pass the real submitted mode instead of `'image'`. If the product cannot send that distinction, the UI and API should not expose independent draw/upload toggles because the server cannot truthfully enforce them.

If this does not match the intended security model for this PR, please reply with the expected behavior and I will treat that as signal to improve the test.

PR #2917, opened 2026-06-03
fix: prevent sending emails for free organisation claims

roshboi:
fix: prevent sending emails for free organisation claims


dguyen (approved):
Wow amazing

PR #2932, opened 2026-06-04
fix: address 16 verified bugs from Faultmark report (#2829)

### LINKED ISSUE #2829: Faultmark scanned this repo and found 24 verified bugs

Abzaek:
Closes #2829.

Went through the Faultmark report and fixed 16 issues. Most were straightforward — stale deps, missing guards, things that break under specific conditions but feel obvious once you see them. A couple were worth explaining.

**Bug #8 — spinner destruction timing**

This one took a second look. The report flagged `destroy()` being called twice on the signature field's Konva spinner group. That was misleading — the real problem is more subtle. When the user needs to reauthenticate (2FA, passkey), the original code called `destroy()` in `.finally()` which fires immediately after the auth dialog opens. So the spinner vanishes while the dialog is still on screen. Then `onReauthFormSubmit` fires later and calls `destroy()` again on an already gone group.

Claude's first fix replaced `.finally()` with `.catch()`, but that introduced a leak — if the user cancels the auth dialog, the spinner never gets cleaned up and sits there on the canvas forever since the field is already signed.

The fix: a `destroySpinnerOnce` guard. Both `onReauthFormSubmit` and `.finally()` call it, but only the first call actually runs. This keeps the spinner visible during auth (correct UX) and cleans up on cancel (no leak).

**The other 15 fixes sorted by category:**

| What | Where | Why |
|------|-------|-----|
| XSS via `dangerouslySetInnerHTML` | `app-banner.tsx` | Banner content came from API with no sanitization |
| **React hooks** |
| Missing deps in `useMemo`/`useCallback` | `security.sessions.tsx`, `configure-document-recipients.tsx`, `document-signing-number-field.tsx` | Seven empty or stale dependency arrays causing stale closures |
| Stale `selectedTemplateId` in callback | `public-profile-template-manage-dialog.tsx` | Closure captured value from the wrong render |
| Stale `isOpen` from nested state | `app-command-menu.tsx` | `v.open` from the loop was being read stale |
| **Edge cases** |
| `teamUrl` could be null | `app-command-menu.tsx` | Two navigation functions would crash when user isn't in a team |
| `transferTeamId` could be -1 | `team-delete-dialog.tsx` | Sent -1 to the API as a valid user ID |
| `Math.max()` with single argument | `configure-fields-view.tsx` | No-op, returned the argument unchanged |
| `for...of` on `Storage.key()` with mutation | `document-edit-form.tsx` | Skipped every other key because keys shift during iteration |
| **Data & form** |
| Field duplication kept the wrong recipient | `configure-fields-view.tsx` | Used sidebar selection instead of the original field's recipient |
| Radio form calling both `field.onChange` and `form.setValue` | `editor-field-radio-form.tsx` | Double-write caused React to reconcile the wrong value |
| Shared `new UAParser()` across cell renders | `security.sessions.tsx` | Parser has internal state, could produce wrong output under load |
| Infinite re-render on team settings page | `o.$orgUrl.settings.teams.tsx` | `setSearchParams` in `useEffect` triggered itself |

Made sure nothing was already addressed upstream before touching it.

### DIFF
--- a/apps/remix/app/components/dialogs/public-profile-template-manage-dialog.tsx
+++ b/apps/remix/app/components/dialogs/public-profile-template-manage-dialog.tsx
@@ -184,6 +184,7 @@ export const ManagePublicTemplateDialog = ({
 
   useEffect(() => {
     const initialTemplate = directTemplates.find((template) => template.id === initialTemplateId);
+    const newSelectedId = initialTemplate?.id ?? null;
 
     if (initialTemplate) {
       setSelectedTemplateId(initialTemplate.id);
@@ -196,7 +197,7 @@ export const ManagePublicTemplateDialog = ({
       setSelectedTemplateId(null);
     }
 
-    const step = initialStep || (selectedTemplateId ? 'MANAGE' : 'SELECT_TEMPLATE');
+    const step = initialStep || (newSelectedId ? 'MANAGE' : 'SELECT_TEMPLATE');
 
     setCurrentStep(step);
 
--- a/apps/remix/app/components/dialogs/team-delete-dialog.tsx
+++ b/apps/remix/app/components/dialogs/team-delete-dialog.tsx
@@ -68,7 +68,10 @@ export const TeamDeleteDialog = ({ trigger, teamId, teamName, redirectTo }: Team
     try {
       const transferTeamId = data.transferTeamId ? parseInt(data.transferTeamId, 10) : undefined;
 
-      await deleteTeam({ teamId, transferTeamId: transferTeamId || undefined });
+      await deleteTeam({
+        teamId,
+        transferTeamId: transferTeamId && transferTeamId > 0 ? transferTeamId : undefined,
+      });
 
       await refreshSession();
 
--- a/apps/remix/app/components/embed/authoring/configure-document-recipients.tsx
+++ b/apps/remix/app/components/embed/authoring/configure-document-recipients.tsx
@@ -98,7 +98,7 @@ export const ConfigureDocumentRecipients = ({ control, isSubmitting }: Configure
       // Update the form
       replace(updatedSigners);
     },
-    [signers, replace, getValues],
+    [signers, replace, getValues, signingOrder],
   );
 
   const onDragEnd = useCallback(
@@ -120,7 +120,7 @@ export const ConfigureDocumentRecipients = ({ control, isSubmitting }: Configure
       // Update the form with new ordering
       replace(updatedSigners);
     },
-    [move, replace, getValues],
+    [move, replace, getValues, signingOrder],
   );
 
   const onSigningOrderChange = (signingOrder: DocumentSigningOrder) => {
--- a/apps/remix/app/components/embed/authoring/configure-fields-view.tsx
+++ b/apps/remix/app/components/embed/authoring/configure-fields-view.tsx
@@ -182,8 +182,8 @@ export const ConfigureFieldsView = ({
             ...structuredClone(lastActiveField),
             nativeId: undefined,
             formId: nanoid(12),
-            signerEmail: selectedRecipient?.email ?? lastActiveField.signerEmail,
-            recipientId: selectedRecipient?.id ?? lastActiveField.recipientId,
+            signerEmail: lastActiveField.signerEmail,
+            recipientId: lastActiveField.recipientId,
             pageX: lastActiveField.pageX + 3,
             pageY: lastActiveField.pageY + 3,
           };
@@ -209,8 +209,8 @@ export const ConfigureFieldsView = ({
               ...structuredClone(lastActiveField),
               nativeId: undefined,
               formId: nanoid(12),
-              signerEmail: selectedRecipient?.email ?? lastActiveField.signerEmail,
-              recipientId: selectedRecipient?.id ?? lastActiveField.recipientId,
+              signerEmail: lastActiveField.signerEmail,
+              recipientId: lastActiveField.recipientId,
               pageNumber,
             };
 
@@ -228,7 +228,7 @@ export const ConfigureFieldsView = ({
         });
       }
     },
-    [append, lastActiveField, selectedRecipient?.email, selectedRecipient?.id, toast],
+    [_, append, lastActiveField, toast],
   );
 
   const onFieldPaste = useCallback(
@@ -409,15 +409,15 @@ export const ConfigureFieldsView = ({
 
   useEffect(() => {
     const observer = new MutationObserver((_mutations) => {
-      const $page = document.querySelector(PDF_VIEWER_PAGE_SELECTOR);
+      const $page = document.querySelector<HTMLElement>(PDF_VIEWER_PAGE_SELECTOR);
 
       if (!$page) {
         return;
       }
 
       fieldBounds.current = {
-        height: Math.max(DEFAULT_HEIGHT_PX),
-        width: Math.max(DEFAULT_WIDTH_PX),
+        height: Math.max($page.offsetHeight, DEFAULT_HEIGHT_PX),
+        width: Math.max($page.offsetWidth, DEFAULT_WIDTH_PX),
       };
     });
 
--- a/apps/remix/app/components/forms/editor/editor-field-radio-form.tsx
+++ b/apps/remix/app/components/forms/editor/editor-field-radio-form.tsx
@@ -178,19 +178,17 @@ export const EditorFieldRadioForm = ({
                             className="h-5 w-5 border-foreground/30 data-[state=checked]:bg-primary"
                             checked={field.value}
                             onCheckedChange={(value) => {
-                              // Uncheck all other values.
+                              const isChecked = Boolean(value);
                               const currentValues = form.getValues('values') || [];
 
-                              if (value) {
-                                const newValues = currentValues.map((val) => ({
-                                  ...val,
-                                  checked: false,
-                                }));
+                              // Compute the full final state in a single write so we never leave
+                              // a window where two options appear checked simultaneously.
+                              const newValues = currentValues.map((val, valIndex) => ({
+                                ...val,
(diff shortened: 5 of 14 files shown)


Mythie:

https://documen.so/cla

Abzaek:
saw that the issue is already closed

PR #2949, opened 2026-06-05
fix(webhooks): scope event-trigger webhook lookup by team, not document-owner membership

henn0124:
## Problem

`document.completed` (and every other event) webhooks are never delivered for documents whose owner isn't a current org-group member of the team — most notably **any document created via a team-scoped API token**, because such tokens resolve their user to the organisation owner (`get-api-token-by-token.ts`). The **"Test webhook"** button also reports success while delivering nothing.

Reproduced on a self-hosted **v2.11.0** instance: create a document via API token → sign + complete → status `COMPLETED`, PDF sealed correctly, but **Team → Webhooks → Logs stays empty** and no delivery is attempted.

## Cause

`getAllWebhooksByEventTrigger()` — the system lookup that decides which webhooks fire on an event (e.g. from the `seal-document` job) — scopes the query with the authorization helper `buildTeamWhereQuery({ teamId, userId })`:

```ts
team: buildTeamWhereQuery({ teamId, userId }),
```

That helper additionally requires `userId` to be a current member of the team via `teamGroup → organisationGroup → organisationGroupMembers → organisationMember.userId`. It's the right scoping for *user-facing reads*, but on a webhook **event** there is no acting/authenticated user — the `userId` passed in is the **document owner**, which can be:

- the **organisation owner** (team-scoped API tokens resolve to the org owner),
- a member who has since **left the team**, or
- a self-host whose org-group rows were never backfilled.

In all of those cases the join matches nothing, `getAllWebhooksByEventTrigger` returns `[]`, and `triggerWebhook` hits its silent `return` — no delivery, no `WebhookCall` row, no error. (Same path explains the "Test webhook" false success: `triggerTestWebhook → triggerWebhook` swallows the empty result and returns `{ success: true }`.)

## Fix

Scope the event lookup by `teamId` alone. `Webhook.teamId` and `Envelope.teamId` are both required columns, so team identity fully determines delivery — the document owner's org-group membership has no bearing on which of a team's webhooks should fire. The user-membership join is retained where it belongs: the settings-list read (`getWebhooksByTeamId`) and the create/edit/delete permission gates.

## Verification

Faithful SQL reproduction of the generated relation-filter against Postgres (current vs fixed):

| Scenario | current (`buildTeamWhereQuery`) | fixed (`teamId`-only) |
|---|---|---|
| webhook on team T, document owned by **org owner** (API-token case) | **0 rows** ❌ | **1 row** ✅ |
| same team T, document owned by a user who *is* in the org-group | 1 row | 1 row |
| team whose org-group rows were never backfilled, any user | **0 rows** ❌ | **1 row** ✅ |

The middle row shows the failure is conditional on *who owns the document*, not on the team.

## Notes

- Touches three files: the query function plus its two callers (`trigger/trigger-webhook.ts` and the legacy `trigger/handler.ts`). `triggerWebhook`'s `userId` option is retained for caller compatibility but no longer scopes the lookup.
- Separately, the silent `return` on an empty match in `trigger-webhook.ts` is what hid this for so long. Happy to follow up with an observability-only PR (a debug log there, plus surfacing failed internal-job dispatch in the local jobs provider) if that'd be welcome.

### DIFF
--- a/packages/lib/server-only/webhooks/get-all-webhooks-by-event-trigger.ts
+++ b/packages/lib/server-only/webhooks/get-all-webhooks-by-event-trigger.ts
@@ -1,25 +1,42 @@
 import { prisma } from '@documenso/prisma';
 import type { WebhookTriggerEvents } from '@prisma/client';
 
-import { buildTeamWhereQuery } from '../../utils/teams';
-
 export type GetAllWebhooksByEventTriggerOptions = {
   event: WebhookTriggerEvents;
-  userId: number;
   teamId: number;
 };
 
-export const getAllWebhooksByEventTrigger = async ({ event, userId, teamId }: GetAllWebhooksByEventTriggerOptions) => {
-  return prisma.webhook.findMany({
+/**
+ * Find every enabled webhook a team has registered for a given event.
+ *
+ * This is a SYSTEM lookup, run when an event fires (e.g. a document is sealed
+ * in a background job) to decide which webhooks to deliver. It is intentionally
+ * scoped by `teamId` ALONE: webhooks belong to a team (`Webhook.teamId` is a
+ * required column) and a domain event belongs to a team, so team identity is
+ * the only thing that decides delivery.
+ *
+ * It must NOT be narrowed by the document owner's org-group membership (the
+ * `buildTeamWhereQuery` user-membership join that the settings-list query
+ * `getWebhooksByTeamId` uses). That join answers an authorization question —
+ * "may this logged-in user see this team's webhooks" — which is meaningless
+ * here: the "user" attached to a completion event is the document's owner,
+ * which may be the organisation owner (for team-scoped API tokens, see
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

henn0124:
**Verified live on a self-hosted instance.**

Built this patch into the official `v2.11.0` image (multi-arch) and deployed it to a self-host. After the change, completing a **real** document delivers the `document.completed` webhook (HTTP 200) to the registered endpoint — where previously the delivery log stayed empty for real (non-test) completions, and the "Test" button reported success while delivering nothing.

The completing document was owned by a user who was not in the team's org-group (the team-scoped-API-token / org-owner case described above), so `buildTeamWhereQuery` matched no rows and the event was dropped silently. Scoping the event lookup by `teamId` alone resolves it. Sealing/PAdES output unaffected.

henn0124:
@Mythie CLA completed

PR #2952, opened 2026-06-06
fix(lib): harden branding CSS sanitizer against escaped url() bypass

joeltco:
## Summary

`sanitizeBrandingCss` blocks dangerous value tokens (`url(`, `expression(`, `@import`, `javascript:`) with a lowercased substring match. Because CSS lets any character in a value be written as an escape sequence, these tokens can be smuggled past the check:

- `\75rl("https://attacker.example/x.png")` is `url("…")` to the browser, but contains no literal `url(` substring.
- `\55RL(…)` — fully escaped and mixed-case — slips through the same way.

`background`, `background-image`, `cursor`, etc. are not in `BLOCKED_PROPERTIES`, so this `url(` value check is the only thing preventing branding CSS (configured per-organisation, rendered to recipients) from loading external resources. The bypass re-enables exactly what the sanitizer is meant to stop: external resource loads that can leak a recipient IP / User-Agent, enable view tracking, or pull in attacker-controlled images and fonts.

## Fix

Decode CSS escape sequences before the blocklist check, so `\75rl(` collapses back to `url(` and is caught. The decoder follows [CSS Syntax 4.3.7](https://www.w3.org/TR/css-syntax-3/#consume-escaped-code-point) and maps null / out-of-range / surrogate escapes to `U+FFFD`, so it can never throw on hostile input.

## Tests

Added to `sanitize-branding-css.test.ts`:
- escaped `\75rl(...)` is dropped
- fully-escaped, mixed-case `\55RL(...)` is dropped
- a harmless value containing escapes (`\72 ed` -> `red`) is retained (guards against over-blocking)

All sanitizer tests pass (79); `biome check` is clean on the changed file.

### DIFF
--- a/packages/lib/utils/sanitize-branding-css.test.ts
+++ b/packages/lib/utils/sanitize-branding-css.test.ts
@@ -252,6 +252,31 @@ describe('sanitizeBrandingCss', () => {
       expect(result.warnings).toHaveLength(1);
       expect(result.warnings[0].kind).toBe('value');
     });
+
+    it('drops a url() smuggled past the blocklist via a CSS escape (\\75rl)', () => {
+      // `\75rl(...)` is `url(...)` to a browser but contains no literal
+      // "url(" substring, so a raw match would let it through.
+      const result = sanitizeBrandingCss('.x { background: \\75rl("http://evil"); }');
+
+      expect(result.css).not.toContain('evil');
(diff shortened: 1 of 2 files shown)


Mythie:

https://documen.so/cla

joeltco:
### Signed the CLA — ready for review, thanks!

PR #2957, opened 2026-06-07
fix: render signing page in the document's language for anonymous recipients

leandrosemog:
## Summary

Anonymous recipients always see the signing page (`/sign/<token>`, `/d/<token>`) in **English**, regardless of the document's language. The locale is resolved in `entry.server.tsx` / `root.tsx` from the cookie / `Accept-Language` **before** route loaders run, so the document's `DocumentMeta.language` is never consulted for the SSR render.

This adds a small server helper, `getRecipientDocumentLanguage(request)`, that — **only** for `/sign/<token>` and `/d/<token>` URLs — looks up the envelope's `documentMeta.language` by token and, when it is a supported language, uses it for:
- the SSR lingui activation (`entry.server.tsx`), and
- the `<html lang>` value (`root.tsx`),

falling back to the existing cookie / `Accept-Language` behaviour otherwise. The browser-language cookie still persists the **cookie value** (not the document language), so opening a `pt-BR` signing link does not permanently switch the visitor's app locale. All lookups are wrapped in `try/catch` so an i18n failure never breaks the signing page. Non-recipient routes are unaffected.

## Test plan
- Set a document's `meta.language` to e.g. `pt-BR`, open its `/sign/<token>` link in a browser sending `Accept-Language: en-US` → the signing UI renders in Portuguese and `<html lang="pt-BR">`.
- Open any non-recipient route → locale unchanged (cookie / Accept-Language).

### DIFF
--- a/apps/remix/app/entry.server.tsx
+++ b/apps/remix/app/entry.server.tsx
@@ -11,6 +11,7 @@ import type { AppLoadContext, EntryContext } from 'react-router';
 import { ServerRouter } from 'react-router';
 
 import { langCookie } from './storage/lang-cookie.server';
+import { getRecipientDocumentLanguage } from './utils/recipient-document-language.server';
 
 export const streamTimeout = 5_000;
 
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

dguyen:
This is intentional, the document language is only for emails

For the recipient, it will always try to render based on the browser language, not english.

PR #2960, opened 2026-06-08
fix: evaluate session cookie expiry at set-time rather than module-load time

### LINKED ISSUE #2666: Constant expiration date in auth cookies
### Issue Description

I'm noticing a bug in the authentication flow.

The issue is represented by login not working: after login, if credentials are correct, the user is redirected again to the login screen.

The Documenso instance is self-hosted and behind a Cloudflare tunnel. The api endpoint is proxied through a Cloudflare worker in order to allow programmatic access to the api endpoint without requiring Cloudflare authentication (i.e., to allow SDK usage).

The following `__Secure-sessionId` cookies are returned on subsequent requests. Requests are performed at 5-10 minutes intervals:

```
__Secure-sessionId=mh2f***;
Domain=my.custom.domain;
Path=/;
Expires=Thu, 26 Mar 2026 15:12:16 GMT;
HttpOnly; Secure;
SameSite=None
```

```
__Secure-sessionId=pdn3***;
Domain=my.custom.domain;
Path=/;
Expires=Thu, 26 Mar 2026 15:12:16 GMT;
HttpOnly; Secure;
SameSite=None
```

```
__Secure-sessionId=urny***;
Domain=my.custom.domain;
Path=/;
Expires=Thu, 26 Mar 2026 15:12:16 GMT;
HttpOnly; Secure;
SameSite=None
```

It's clear that the authentication flow is not working because the cookies are sent with a expiration date in the past; therefore, the browser immediately discards them and  

Things to notice:

- The behavior is consistent across browsers
- The cookie value itself changes, suggesting this may not be a caching issue
- The expiration date is constant, suggesting the date may be constant in the compiled version
- If I reload the instance, the date is updated, but it still constant across requests (updated to `Fri, 01 may 2026 09:39:18`

I think there may be a constant date which is calculated when the application is started and never updated.

### Expected Behavior

I would expect the cookie expiration date to reflect a timedelta from time of login, instead of being constant.

### Current Behavior

The cookie has a constant expiration date.

Abzaek:
## Description

This PR fixes a bug where the session cookie `expires` timestamp is computed once at module-load time and never updated, causing all session cookies to share the same fixed expiration timestamp (calculated from server start time). After `AUTH_SESSION_LIFETIME` (default 7 days) has elapsed since the server started, all newly issued session cookies have an expiration in the past, immediately logging users out.

## Related Issue

Resolves #2666

## Changes Made

- Removed the static `expires` property from the global `sessionCookieOptions` object so the expiration is not frozen at module import time.
- Added the `expires` property dynamically in `setSessionCookie()` using `expires: new Date(Date.now() + AUTH_SESSION_LIFETIME)`, ensuring the timestamp is computed at the moment the cookie is actually set.

## Note

The `deleteSessionCookie` and `setCsrfCookie` functions also use `sessionCookieOptions` — `deleteCookie` works via `maxAge` semantics and does not depend on `expires`, and `setCsrfCookie` already overrides `expires: undefined` explicitly.

## CLA

This PR is ready for review once the contributor has signed the CLA at https://documen.so/cla

### DIFF
--- a/packages/auth/server/lib/session/session-cookies.ts
+++ b/packages/auth/server/lib/session/session-cookies.ts
@@ -30,7 +30,6 @@ export const sessionCookieOptions = {
   sameSite: useSecureCookies ? 'none' : 'lax',
   secure: useSecureCookies,
   domain: getCookieDomain(),
-  expires: new Date(Date.now() + AUTH_SESSION_LIFETIME),
 } as const;
 
 export const extractSessionCookieFromHeaders = (headers: Headers): string | null => {
@@ -56,7 +55,10 @@ export const getSessionCookie = async (c: Context): Promise<string | null> => {
  * @param sessionToken - The session token to set.
  */
 export const setSessionCookie = async (c: Context, sessionToken: string) => {
-  await setSignedCookie(c, sessionCookieName, sessionToken, getAuthSecret(), sessionCookieOptions).catch((err) => {
+  await setSignedCookie(c, sessionCookieName, sessionToken, getAuthSecret(), {
+    ...sessionCookieOptions,
+    expires: new Date(Date.now() + AUTH_SESSION_LIFETIME),
+  }).catch((err) => {
     appLog('SetSessionCookie', `Error setting signed cookie: ${err}`);
 
     throw err;


Mythie:

https://documen.so/cla

dguyen:
Already fixed

PR #2961, opened 2026-06-08
fix: thread CSP nonce into DragDropContext and set __webpack_nonce__ global

### LINKED ISSUE #2872: CSP: per-request style-src-elem nonce isn't threaded into react-beautiful-dnd or react-remove-scroll-bar — vendor inline styles blocked
## Symptom

On a self-hosted v2.11.0 deployment with the default strict CSP from `apps/remix/server/security-headers.ts`, the browser console fires ~10 violations per page load on the document/envelope editor:

```
Applying inline style violates the following Content Security Policy directive 'style-src-elem 'self' 'nonce-XwUIm0MtXl36FOcgpDe6mw=='.
```

The two chunks emitting blocked styles are:
- `dnd.esm-*.js` — react-beautiful-dnd / @hello-pangea/dnd. Its `Lr(nonce)` helper sets `<style nonce="...">` *if* a nonce is passed. Documenso never passes one to `<DragDropContext nonce>` (or to `useStyleMarshal`).
- `Combination-*.js` — `react-remove-scroll-bar` / `use-sidecar` lineage. Reads from the build-time `__webpack_nonce__` global. Documenso's Vite/Rollup build doesn't emit a `__webpack_nonce__` runtime assignment from the CSP middleware's nonce.

Both of these are textbook fixes:
- `<DragDropContext nonce={nonce}>` from the root loader's `nonce` value (already plumbed into `<Scripts nonce>` and `<ServerRouter nonce>`).
- A small client-side bootstrap that sets `__webpack_nonce__ = window.__NONCE__` (or equivalent) before the chunks hydrate.

## Why "embed routes only" fix wasn't enough

`security-headers.ts` already documents that embed routes need `style-src-elem 'self' 'unsafe-inline'` because customer CSS is injected at runtime. That justification (runtime `<style>` injection that can't be nonce-stamped) is identical to what the bundled vendors do on every other page — they should be threaded the nonce instead of falling back to no-nonce.

## Repro

1. Self-host v2.11.0 with the default CSP middleware (no env overrides exist).
2. Open the envelope editor / template editor / any page using react-beautiful-dnd.
3. Browser DevTools console → CSP violations from `dnd.esm-*.js` and `Combination-*.js`.

## Suggested fix

- Wrap the lib usage in a small adapter that passes the loader-context nonce: `<DragDropContext nonce={useLoaderData().nonce}>` (or pass via context).
- Add a `<script nonce={nonce}>{`window.__webpack_nonce__ = ${JSON.stringify(nonce)}`}</script>` early in `<head>` to satisfy the `__webpack_nonce__`-reading libs.

Happy to send a PR if the maintainers agree on the approach.

Abzaek:
## Description

Fixes a CSP (Content Security Policy) violation issue where `@hello-pangea/dnd` (react-beautiful-dnd) and `react-remove-scroll-bar` / `use-sidecar` were generating inline `<style>` elements without the per-request CSP nonce.

On self-hosted deployments with strict CSP (the default from `security-headers.ts`), the browser console fires ~10 violations per page load on the envelope editor:

> Applying inline style violates Content Security Policy directive 'style-src-elem 'self' 'nonce-...'

## Changes

1. **`apps/remix/app/root.tsx`** — Added a nonce-gated `<script>` in `<head>` that sets `window.__webpack_nonce__` to the CSP nonce, satisfying `react-remove-scroll-bar` / `use-sidecar`'s `__webpack_nonce__` check (the mechanism these libs use to discover the nonce).

2. **`envelope-editor-recipient-form.tsx`** — Passed `nonce={nonceVal}` to `<DragDropContext>` using the existing `useCspNonce()` hook.

3. **`envelope-editor-upload-page.tsx`** — Same pattern on `<DragDropContext>`.

4. **`configure-document-recipients.tsx`** — Same pattern on `<DragDropContext>`.

## Root cause

The per-request CSP nonce was already plumbed through the root loader (`context.nonce` → `loaderData.nonce`) and threaded into `<Scripts nonce>`, `<Links nonce>`, and inline `<script>` / `<style>` blocks. However, two vendor libraries create their own inline styles independently:
- `@hello-pangea/dnd` accepts a `nonce` prop on `<DragDropContext>` which Documenso wasn't passing
- `react-remove-scroll-bar` reads `window.__webpack_nonce__`, which was never set at runtime

## Related issue

Closes #2872

## Verification

- `npx tsc --noEmit` passes (zero errors in any edited file)
- No `npm run build` was run (per project guidelines)


Mythie:

https://documen.so/cla

catalinpit:
Thanks for the PR! Will have to close it because it also contains unrelated changes and it's very outdated.

Issue #2247, opened 2025-11-26
Failed to Render Email/No Password or Activation Email Send


lotus-infosec:
### Issue Description

When trying to send a password reset email I get a "`Error: Failed to render email`".

I also don't get activation emails and can see that the job is hanging in the Postgres DB. Confirmed that my SMTP settings are correct and am pulling from the latest docker instance.

## LOGS:

```app_1       | 🚀 Starting Documenso...
app_1       | 
app_1       | 🔐 Checking certificate configuration...
app_1       | ✅ Certificate file found and readable - document signing is ready!
app_1       | 
app_1       | 📚 Useful Links:
app_1       | 📖 Documentation: https://docs.documenso.com
app_1       | 🐳 Self-hosting guide: https://docs.documenso.com/developers/self-hosting
app_1       | 🔐 Certificate setup: https://docs.documenso.com/developers/self-hosting/signing-certificate
app_1       | 🏥 Health check: http://localhost:3000/api/health
app_1       | 📊 Certificate status: http://localhost:3000/api/certificate-status
app_1       | 👥 Community: https://github.com/documenso/documenso
app_1       | 
app_1       | 🗄️  Running database migrations...
app_1       | Prisma schema loaded from ../../packages/prisma/schema.prisma
app_1       | Datasource "db": PostgreSQL database "documenso", schema "public" at "database:5432"
app_1       | 
app_1       | 142 migrations found in prisma/migrations
app_1       | 
app_1       | 
app_1       | No pending migrations to apply.
app_1       | 🌟 Starting Documenso server...
app_1       | [Telemetry] Telemetry is disabled. To enable, remove the DOCUMENSO_DISABLE_TELEMETRY environment variable.
database_1  | 2025-11-26 05:01:35.426 UTC [27] LOG:  checkpoint starting: time
database_1  | 2025-11-26 05:01:35.747 UTC [27] LOG:  checkpoint complete: wrote 3 buffers (0.0%); 0 WAL file(s) added, 0 removed, 0 recycled; write=0.099 s, sync=0.050 s, total=0.321 s; sync files=2, longest=0.025 s, average=0.025 s; distance=0 kB, estimate=0 kB
app_1       | Error: A component suspended while responding to synchronous input. This will cause the UI to be replaced with a loading indicator. To fix, updates that suspend should be wrapped with startTransition.
app_1       |     at renderToStringImpl (/app/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:7002:11)
[...]

PR #2968, opened 2026-06-09
fix: signing request email renders blank when organisation/team branding is enabled

### LINKED ISSUE #2953: Signing request email renders blank/empty when organisation/team branding is enabled (v2.11.0, self-hosted)
## Summary

On a self-hosted instance, when **organisation/team branding is enabled**, the **signing-request email** (`send.signing.requested.email`) is delivered with an **empty body** (no signing link/content). SMTP works — the message is delivered, it just has no rendered content. **Disabling branding** makes the email render correctly.

Auth/activation emails render fine (this looks like the same class as #2247, fixed for auth emails by #2297), but the **signing-request email path still suspends** during render.

## Environment

- Self-hosted via Docker, image `documenso/documenso:latest` = **v2.11.0** (image created 2026-05-13)
- Postgres 15, object storage = S3 (Cloudflare R2), SMTP = working (emails are delivered)
- Reproduced via both the public API and the UI

## Suspected root cause

The signing-request email template performs a `team.email.get` query inside a `<Suspense>` boundary **during synchronous email render** (`@react-email/render`), which suspends and yields an empty body. Relevant runtime log around the send:

```
[JOBS]: Triggering job send.signing.requested.email with payload { ... }
Warning: Each child in a list should have a unique "key" prop.
Check the top-level render call using <Text>.
    at Tailwind (file:///app/node_modules/@react-email/tailwind/dist/index.mjs:...)
    at Suspense
    path: "team.email.get"
```

This mirrors #2247 (resolved for auth emails by #2297 via pre-fetching), but the signing-request email still fetches `team.email.get` during render instead of pre-fetching/passing it as props.

## Steps to reproduce

1. Enable organisation/team branding.
2. Create and send a document for signing (API `POST /api/v1/documents` + `/send`, or via the UI).
3. The recipient receives a signing-request email with an **empty body**.
4. Disable branding → the same flow produces a correctly rendered email.

## Expected

The signing-request email should render its content (signing button/link, message) regardless of whether branding is enabled.

## Related

- #2247 (auth/activation emails failing to render — same "component suspended" class)
- #2297 (fix for the auth email path)
- #2596 (Organization Invitation Email - empty — likely the same underlying team-scoped email render issue)

NandiniDhanrale:
## Description

Fixes signing request emails rendering with an empty body when organisation/team branding is enabled.

## Root Cause

The TemplateFooter component renders company branding details via .map() when randing.brandingEnabled is true. However, the fragments returned by .map() lacked a key prop, causing React warnings.

@react-email/render v2.0.0 wraps all email content in <Suspense> without a fallback. When React's streaming SSR encounters components that trigger warnings (like missing keys in lists), the Suspense boundary can fail to render any content, resulting in an empty email body.

This mirrors the same class of issue as #2247 (fixed by #2297), where Suspense-related rendering issues caused blank email content.

## Fix

Changed the anonymous <>...</> fragment in packages/email/template-components/template-footer.tsx to a keyed <Fragment key={idx}>...</Fragment>.

## Testing

- Only triggered when randing.brandingEnabled && branding.brandingCompanyDetails is true
- Other email templates (bulk-send-complete, template-custom-message-body) already used proper keyed fragments
- Verified no other .map() with missing keys exist in email templates

Closes #2953


Mythie:

https://documen.so/cla

Mythie (approved):

Mythie:
Thanks for that!

PR #2980, opened 2026-06-13
fix: webhook payload contains stale deletedAt on document cancellation

Arunkoo:
## Description

The webhook payload sent on `DOCUMENT_CANCELLED` events contains a stale `deletedAt` value. When an owner/team member deletes a `COMPLETED` document, the document is correctly soft-deleted in the database (`deletedAt` is set to the current timestamp), but the webhook payload is built from the original `envelope` snapshot fetched *before* the deletion occurred, so it reports `deletedAt: null` even though the document was just deleted.

## Changes Made

- In `deleteDocument`, capture the return value of `handleDocumentOwnerDelete` (`updatedEnvelope`), which contains the post-deletion envelope state (including the updated `deletedAt`).
- Merge `updatedEnvelope` over the original `envelope` (`{ ...envelope, ...updatedEnvelope }`) before passing it to `mapEnvelopeToWebhookDocumentPayload`, so the webhook payload reflects the current `deletedAt` value.
- `handleDocumentOwnerDelete` can return `undefined` in its early-exit path (document already deleted); the merge falls back to the original `envelope` in that case, preserving existing behavior.

## Testing Performed

- Reproduced the bug locally: deleted a `COMPLETED` document with a webhook subscribed to `DOCUMENT_CANCELLED`, confirmed the payload showed `deletedAt: null` while the database row had `deletedAt` correctly set.
- Applied the fix and repeated the test: payload now shows `deletedAt` matching the actual deletion timestamp (matches `updatedAt` to the millisecond).
- Ran `npx tsc --noEmit` on `packages/lib` - no new type errors introduced (pre-existing unrelated errors in `get-completed-fields-for-document.ts`, `get-active-subscriptions-by-user-id.ts`, and `find-organisation-invoices.ts` confirmed present on `main` as well, unaffected by this change).

## Screenshots

**Before fix** - webhook payload shows `deletedAt: null` despite the document being deleted:
before webhook data1

**After fix** - webhook payload correctly shows `deletedAt` timestamp:
after webhook data

## Checklist

## Additional Notes

This is a small, focused fix scoped to a single file (`packages/lib/server-only/document/delete-document.ts`). The hard-delete path (draft/pending documents) is unaffected - `deletedAt: null` remains correct there since the row is fully removed rather than soft-deleted.


Mythie:

https://documen.so/cla

Arunkoo:
Hi @dguyen , hope you’re doing well! I came across a bug in the codebase and opened this PR with a fix. Wanted to gently follow up in case it got buried. No rush at all — whenever you get a chance to take a look, I’d really appreciate it. Happy to discuss the approach or make any changes if needed. Thanks so much for your time! 🙏

dguyen (approved):
Looks good to me, thanks!

Issue #3000, opened 2026-06-18
Reliability concern: resendDocument can partially send reminders before rejecting


bharathkumar39293:
## Reliability concern

While looking at `resendDocument`, I noticed that manual reminder emails are sent with `Promise.all(...)`.

This creates a partial side-effect scenario if one recipient email succeeds and another recipient email fails.

## Current behavior

Example with two recipients:

```txt
recipient A reminder email succeeds
recipient B reminder email fails
resendDocument rejects
caller/user sees the resend action as failed
if the user retries the resend action, recipient A can receive the reminder again
```

I reproduced this with a focused test by mocking the email transport:

- first `sendMail` succeeds
- second `sendMail` rejects
- `resendDocument` rejects
- one `EMAIL_SENT` audit log is created for the successful recipient
- retrying calls `sendMail` again for the first recipient

## Question

Is this all-or-nothing rejection behavior intentional for manual reminder sends?

I did not find comments, tests, or history indicating that partial-send-then-reject semantics are intentional. I also noticed other areas of the codebase use `Promise.allSettled(...)` where partial failure should not block the whole batch.

## Possible direction

If reminders are intended to be best-effort, `resendDocument` could use `Promise.allSettled(...)` so a partial email failure does not make the whole action look failed after some reminders were already sent.

If the current behavior is intentional, no change needed. I wanted to confirm the intended behavior before proposing a patch.

ElTimuro:
- Looking into this
- We recently had change related to this

Mythie:
This endpoint is synchronous by design since resending is a deliberate action. The odds of it failing midway through is quite low and would require your email provider to spontaneously go down which would cause all resend attempts afterwards to fail early.

Could be improved with an allSettled and letting the user know which ones failed, but honestly if someone got resent an email a couple of times due to some weird network issue with email sending then that's acceptable for now.

PR #3002, opened 2026-06-19
fix(email): avoid partial reminder resend rejection

xuezhaolan:
## Summary
- switch manual reminder resend fan-out to `Promise.allSettled(...)`
- preserve successful sends and audit logs even if one recipient email fails
- rethrow the first failed resend so callers still see an error instead of a false success

## Why
The current `Promise.all(...)` flow can partially send reminder emails and then reject the whole action when a later recipient send fails. That makes the UI/API look fully failed even though some reminders were already delivered, which encourages duplicate reminders on retry.

## Test Plan
- `git diff --check`
- `node_modules/.bin/biome check packages/lib/server-only/document/resend-document.ts`
- `node_modules/.bin/tsc --noEmit`

Refs #3000


Mythie:

https://documen.so/cla

bharathkumar39293:
@xuezhaolan Thanks for picking this up.

One thing I want to clarify: since this still rethrows after `Promise.allSettled(...)`, the caller/user will still see the resend action as failed after partial success.

So the retry scenario from #3000 may still be possible:

```txt
recipient A reminder succeeds
recipient B reminder fails
resendDocument throws
user retries
recipient A can receive another reminder
```

Is the intended behavior here to make sure all selected sends/audit logs finish before surfacing the error, or should partial reminder failures be treated as best-effort so the whole action does not look failed after some reminders were already sent?

I’m asking because #3000 was mainly about confirming the desired semantics before proposing a patch.

PR #3003, opened 2026-06-21
fix: add NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS to production compose and fix startup cert check

roshboi:
## Summary

- `NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS` was missing from `docker/production/compose.yml`,
  making it impossible to configure base64-encoded certificate signing without manually editing
  the compose file
- `docker/start.sh` only checked for a cert file on disk — if `NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS`
  was set (no volume mount), startup incorrectly warned that signing was unavailable
- `turbo.json` had the local signing variables duplicated; consolidated them and added the missing
  `NEXT_PRIVATE_SIGNING_LOCAL_FILE_ENCODING` to the canonical group

## Changes

- `docker/production/compose.yml` — add `NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS` env passthrough
- `docker/start.sh` — check `NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS` before falling back to file check
- `turbo.json` — remove duplicate signing env entries, add missing `NEXT_PRIVATE_SIGNING_LOCAL_FILE_ENCODING`

## POC

- [x] Deploy production compose locally via kind with `NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS` set (base64 cert) — confirm signing works and startup prints the correct ready message


Mythie:
This would make the config throw if it's not provided though even if you had setup the file path correctly no?

PR #3007, opened 2026-06-22
feat: add API endpoint to reject documents on behalf of recipients

Mythie:
Programmatically record an external rejection on behalf of a recipient
who declined outside the platform. Flags the rejection as external in
the audit log, optionally attributes it to a specific team member via
actAsEmail, and enforces team membership and document visibility.


dguyen (approved):

packages/lib/server-only/document/reject-document-on-behalf-of.ts
```diff
@@ -0,0 +1,207 @@
...
+  await jobs.triggerJob({
+    name: 'send.document.cancelled.emails',
+    payload: {
+      documentId: legacyDocumentId,
+      cancellationReason: reason,
+      requestMetadata: requestMetadata.requestMetadata,
+    },
+  });
+
+  return updatedRecipient;
```
dguyen, line 179: Feels like a lot of dupe code from reject-document-with-token

PR #2888, opened 2026-05-29
Respect default date format in document views

SuryaNarayanaa:
## Description

This PR makes document list/detail views respect the saved org/team Default Date Format preference.

The preference was already being stored correctly, but some document views were formatting dates with hardcoded Luxon/i18n formats, causing inconsistent displays across the app.

## Related Issue

Fixes #2880 

## Changes Made

- Added a small `formatDocumentDate` helper for rendering dates with the configured document date format
- Added `documentDateFormat` to organisation/team session preference
- Updated the Documents table to use the derived team/org date format
- Updated the Inbox table to use the derived team/org date format
- Updated the document details information panel to use the derived team/org date format
- Updated embedded authoring provider mock data to match the extended session preference shape
- Left `DocumentMeta` unchanged, so this works as a display-layer workaround without backfilling existing documents

## Testing Performed

- Ran Remix typecheck successfully in the main workspace:

```sh
npm run typecheck -w @documenso/remix
```

## Checklist

- [x] I have tested these changes locally and they work as expected
- [ ] I have added/updated tests that prove the effectiveness of these changes
- [ ] I have updated the documentation to reflect these changes, if applicable
- [x] I have followed the project's coding style guidelines
- [ ] I have addressed the code review feedback from the previous submission, if applicable

## Additional Notes

This is intentionally a minimal workaround. It does not modify existing `DocumentMeta.dateFormat` values and does not backfill old documents. Instead, affected document views now render dates using the current derived org/team Default Date Format preference

## AI Usage 
AI (codex 5.5) used to find the lines of code relating to this issue and wrote formatDocumentDate helper and it has been understood manually

Issue #2880, opened 2026-05-28
Default time format is not being respected


mKoonrad:
### Issue Description

Default time format (from settings) is not being respected in:

1. Signing Certificate (pdf)
2. Inbox
3. Main table in web app
4. Envelope details

### Steps to Reproduce

1. Set custom time format in settings

<img width="338" height="128" alt="Image" src="https://github.com/user-attachments/assets/42106731-4aeb-4c8d-8121-915ecce9dbaa" />

2. Check time format in:

- Main page `app.documenso.com/t/personal_*/documents`

<img width="496" height="479" alt="Image" src="https://github.com/user-attachments/assets/3963db2b-e246-4fad-ae5b-e4cb054ab7ef" />

- Inbox `app.documenso.com/inbox`

<img width="363" height="579" alt="Image" src="https://github.com/user-attachments/assets/53c3fca6-cd2a-42f1-aca1-cdf26ea1afe6" />

- Signing Certificate (pdf)

<img width="480" height="195" alt="Image" src="https://github.com/user-attachments/assets/d5918d79-40d9-4a05-b974-b88e1b9f7e8c" />

- Envelope details `app.documenso.com/t/*/documents/envelope_*`

<img width="828" height="238" alt="Image" src="https://github.com/user-attachments/assets/663a9eb4-f069-4cda-bb94-fade48baf432" />

### Expected Behavior

Time format should be the same

### Current Behavior

Different time format across app

### Operating System [e.g., Windows 10]

Windows 11

### Browser [e.g., Chrome, Firefox]

Chrome 148

### Version [e.g., 2.0.1]

v2.11.0

### Please check the boxes that apply to this issue report.

- [x] I have searched the existing issues to make sure this is not a duplicate.
- [x] I have provided steps to reproduce the issue.
- [x] I have included relevant environment information.
- [x] I have included any relevant screenshots.
- [x] I understand that this is a voluntary contribution and that there is no guarantee of resolution.
- [ ] I want to work on creating a PR for this issue if approved

SuryaNarayanaa:
## Reproduction of the issue

1. Open Documenso and go to **Organisation/Team Settings**.
2. Under **Document Preferences**, set **Default Date Format** as in the first image .
3. Save the settings.
4. Open the following pages and compare the displayed dates:
   - `/t/personal_*/documents`
   - `/inbox`
   - `/t/*/documents/envelope_*`
   - signing certificate PDF
[...]

PR #3041, opened 2026-06-26
fix(ui): use browser locale date format

mKoonrad:
## Description

This PR changes date format in UI to browser locale format. Dates automatically adapt to the user's browser settings. Format is friendly to all users. Native i18n function don't need luxon import.

Example before:

image

image

After:
image

image

## Related Issue

#2880

@dguyen https://github.com/documenso/documenso/pull/2888#pullrequestreview-4446800076
> For general dates across the website, it should refer to the browser locale instead.

## Changes Made

Unify date format across pages:
- `/t/personal_*/documents`
- `/t/personal_*/templates`
- `t/personal_*/templates/envelope_*`
- `/t/personal_*/documents/envelope_*`
- `t/personal_*/documents/envelope_*/logs`
- `/inbox`
- `/settings/security/activity`

## Testing Performed

- Ran locally
- Ran build

## Checklist


Mythie:

https://documen.so/cla

mKoonrad:
Could someone take a look at this? @dguyen @Mythie @ElTimuro

Mythie (changes requested):
I'm fine with using i18n.date instead of DateTime.setLocale.toFormat since that's more correct, but we shouldn't be changing the time formatting for the existing i18n.date usages.

mKoonrad:
Hi @Mythie. Sorry for late replay, but I was looking for some extra time to double check PR.

`document-page-view-information.tsx‎` and `template-page-view-information.tsx‎` (file where I changed current i18n.date) use now the same format (unified).

I changed time formats in some more places because current UI was a mess, there was no logic.

Why are dates in documents logs are completely different from logs few lines below? Is it good? This is just an example of current mess.

Mythie:
- Dates are condensed in log table lines due to less space to work with.
- Even if I wanted to change them I wouldn't do it as part of this PR but as part of something more focused around improving logs.
- Changing them can confuse existing users who have documents with audit logs in one format vs the other.

mKoonrad:
@Mythie AM/PM suffix from logs is the biggest problem for now. AM/PM isn't used in Poland and (after I did a little research) in Europe also. So why Polish, EU users should see AM/PM?

I don't need to force only my formats, but maybe you have any better idea to unify it? Eg. Use only `i18n.date` with only few date options across whole app, like:
- short,
- medium,
- long version?

Mythie:
I'd still omit that change to the logs as part of this PR and we can address it as part of another one.

mKoonrad:
@Mythie I restored default time format in logs, but with native i18n function, without luxon.

Before:
`6/26/26, 7:45 PM`

After:
`6/26/26, 7:45 PM` (when user have English UI)
`26.06.2026, 07:45 PM` (when user have Polish UI)

What about this solution? It's just a tiny change, better for non-English users.

PR #3057, opened 2026-07-02
fix: improve editor autosave

dguyen:
### Problem

On production, typing a recipient's name/email and quickly navigating to Add Fields could silently lose the typed data — the editor would fall back to the "Recipient 1" placeholder and nothing was persisted. Never reproduced locally because saves complete before navigation.

### Root cause

useEnvelopeAutosave's flush() had a lost-update race: when an older save was still in-flight, it cleared the debounce timer of newer queued data, awaited the in-flight save, and returned — never committing the newer payload. The step change then reset the form from the (stale) server response, wiping the typed data everywhere. The timer callback could also start overlapping saves that could land out of order.

### Fix

Rewrote the hook as a self-serialising queue with a latest-wins payload:
- At most one saveFn in-flight at a time; concurrent flush()/timer callers share the same commit pump.
- The pump drains until nothing is queued, so flush() only resolves once the most recent payload is persisted.
- Uses a latest-ref for saveFn (no stale closures, stable triggerSave/flush identities).

### Testing
- New E2E test (envelope-recipient-autosave-race.spec.ts) reproduces the race by holding the first envelope.recipient.set request open — red on the old hook (1 request sent, typed data never reaches the server, UI shows "Recipient 1"), green with the fix (follow-up save sent, typed recipients persisted and shown).
- Full envelope-editor-v2 suite passes (remaining failures are pre-existing and reproduce identically on the parent commit).


roshboi (approved):
lgtm

PR #3061, opened 2026-07-06
Respect hidden sender details in document invite emails

### LINKED ISSUE #3055: DocumentInviteEmailTemplate message section ignores "includeSenderDetails"
### Issue Description

I have been testing workflow on the staging environment (https://stg-app.documenso.com/) and noticed this issue. The `includeSenderDetails` option (the **Send on Behalf of Team** setting) is not being applied as expected.

https://github.com/documenso/documenso/blob/5a8335e0eb01823b560cb5644bd1db2b105195d7/packages/lib/jobs/definitions/emails/send-signing-email.handler.ts#L153-L156

https://github.com/documenso/documenso/blob/5a8335e0eb01823b560cb5644bd1db2b105195d7/packages/email/templates/document-invite.tsx#L96-L104

### Steps to Reproduce

1. Disable the **Send on Behalf of Team** option in both **Organisation Settings** and **Team Settings**

Image

2. Create an envelope with a recipient
3. Send the document

### Expected Behavior

Image

### Current Behavior

Image

Linux 7.0.14

Chrome 150

### Version [e.g., 2.13.0]

https://stg-app.documenso.com/

xonaib:
## Summary

This fixes the document invite email copy when sender details are hidden for an organisation or team.

The email template already handled the heading and preview text, but the message body could still fall back to wording that showed the sender name. Reminder emails had the same issue because the resend path did not use the resolved `includeSenderDetails` setting.

## Changes

- Use team-only wording in the document invite message body when sender details are hidden.
- Use the resolved email settings when building reminder email copy.
- Pass `includeSenderDetails` into the resend email template.

Closes #3055.

## Testing

- `npx biome check packages/email/templates/document-invite.tsx packages/lib/server-only/document/resend-document.ts`
- `npx tsc -p packages/email/tsconfig.json --noEmit`

`npx tsc -p packages/lib/tsconfig.json --noEmit` still reports existing unrelated Prisma/schema type errors outside this change.

### DIFF
--- a/packages/email/templates/document-invite.tsx
+++ b/packages/email/templates/document-invite.tsx
@@ -96,6 +96,14 @@ export const DocumentInviteEmailTemplate = ({
               <Text className="mt-2 text-base text-muted-foreground">
                 {customBody ? (
                   <TemplateCustomMessageBody text={customBody} />
+                ) : organisationType === OrganisationType.ORGANISATION && !includeSenderDetails ? (
+                  <Trans>
+                    {teamName} has invited you to {action} the document "{documentName}".
+                  </Trans>
+                ) : organisationType === OrganisationType.ORGANISATION ? (
+                  <Trans>
+                    {inviterName} on behalf of "{teamName}" has invited you to {action} the document "{documentName}".
+                  </Trans>
                 ) : (
                   <Trans>
                     {inviterName} has invited you to {action} the document "{documentName}".
--- a/packages/lib/server-only/document/resend-document.ts
+++ b/packages/lib/server-only/document/resend-document.ts
@@ -169,6 +169,7 @@ export const resendDocument = async ({ id, userId, recipients, teamId, requestMe
   const {
     branding,
     emailLanguage,
+    settings,
     organisationType,
     senderEmail,
     replyToEmail,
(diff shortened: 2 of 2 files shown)


xonaib:
@coderabbitai review

packages/email/templates/document-invite.tsx
```diff
@@ -96,6 +96,14 @@ export const DocumentInviteEmailTemplate = ({
               <Text className="mt-2 text-base text-muted-foreground">
                 {customBody ? (
                   <TemplateCustomMessageBody text={customBody} />
+                ) : organisationType === OrganisationType.ORGANISATION && !includeSenderDetails ? (
```
catalinpit: Was this a bug or something? Did you test it beforehand and didn't work properly?

catalinpit (changes requested):

packages/email/templates/document-invite.tsx
xonaib: This came from the behavior reported in #3055: the fallback body did not apply includeSenderDetails, even though the preview/header already did. I wasn't able to independently reproduce the regular-send path, and I saw your note that reminders are reproducible. I've updated the reminder path as the concrete fix and also preserved the existing self-signer reminder message instead of overwriting it for organisation documents. The template condition is now shared so its preview and fallback body stay consistent.

catalinpit: My question is about why was this change necessary? Did you come across a bug regarding this particular code? 

The changes in the resend-document file are fine.

xonaib: You're right, I did not independently reproduce a bug in that template path. The reported/reproducible issue is in reminders, so I've removed the document-invite.tsx change and narrowed the PR to resend-document.ts in 1aec479.

Mythie (approved):

catalinpit (approved):

PR #3062, opened 2026-07-06
fix: keep documents filter bar on a single line on desktop

ephraimduncan:
# Description

Keeps the documents page filter bar (status tabs, sender filter, period selector, search) on a single line at desktop widths instead of wrapping the search input onto a new line or scrolling the tabs horizontally.

- Force a single row at the `xl` breakpoint, where the full filter set fits within the `max-w-screen-xl` container
- Narrow the sender filter (200px → 160px) and period selector (192px → 160px)
- Stretch the search input to fill the remaining row width, flush with the content edge
- Below `xl`, the previous wrapping behaviour is unchanged

## Screenshot

Documents filter bar on a single line


ephraimduncan:
Reopening with the branch pushed directly to this repo.

PR #3067, opened 2026-07-07
Avoid failing resend after partial reminder delivery

### LINKED ISSUE #3000: Reliability concern: resendDocument can partially send reminders before rejecting

xonaib:
## Summary

This makes manual reminder sends handle partial email failures more carefully.

Previously, `resendDocument` used `Promise.all(...)` for recipient reminder emails. If one recipient email failed after another had already been sent, the whole resend action rejected. That made the action look failed even though a recipient had already received the reminder, and retrying could send that reminder again.

This keeps the preflight checks as-is, then lets all reminder sends settle. If at least one reminder was sent, individual failures are logged but the resend does not reject. If every attempted reminder fails, the original failure still bubbles up.

## Changes

- Use `Promise.allSettled(...)` for manual reminder email delivery.
- Log failed recipient reminder sends with the envelope and recipient details.
- Keep total send failure as a failed resend action.

Closes #3000.

## Testing

- `npx biome check packages/lib/server-only/document/resend-document.ts`

`npx tsc -p packages/lib/tsconfig.json --noEmit` still reports existing unrelated Prisma/schema type errors outside this change.


xonaib:
@coderabbitai review

PR #3069, opened 2026-07-07
fix: open document actions dropdown to the right

ephraimduncan:
Makes the document page actions dropdown alignment responsive: `end` on small screens (< 768px) so the menu stays flush with the viewport edge, and `start` on larger screens so it opens to the right of the trigger.

Follows the existing `checkIfMobile` resize-listener pattern from `configure-fields-view.tsx`.

Addresses the review feedback in https://github.com/documenso/documenso/pull/3069#pullrequestreview-4692773285


dguyen (changes requested):
This was done on purpose because this button always sits to the right.

On smaller screens it becomes uneven if it's on the "start"

image

If it's possible make it use "end" on smaller screens and "start" on larger screens. 
Otherwise I think it's better to leave as is and just close the PR.

ephraimduncan:
Closing — the right-alignment is intentional per review feedback.

PR #3070, opened 2026-07-07
feat(ui): redesign recipient field hover card

ephraimduncan:
Redesigns the popover shown when hovering a recipient field avatar in the envelope view.

- Field-first hierarchy: header shows field-type icon + "{Type} field" with inline status (Signed/Pending/Read Only) as a colored dot + label
- Recipient (name/email) moved to a recessed footer well as secondary context
- Hide-field action moved from floating over the text to a ghost icon button in the footer well
- Added a `FieldType` → icon map mirroring `field-selector.tsx`

## Screenshots

| Before | After |
| --- | --- |
| Previous hover tooltip: centered badge, title and recipient text | New hover card: field-first header with status, recipient footer well |


catalinpit (approved):
I like the new change, nice!

PR #3076, opened 2026-07-07
feat: redesign api tokens settings page

ephraimduncan:
# Redesign API tokens settings page

Aligns the API tokens settings page with the webhooks settings pattern.

## Changes

- Token creation moves from an inline form into a **dialog** triggered from the settings header. On success the dialog swaps to a reveal state showing the token in a read-only mono input with an inline copy button.
- The "Never expire" switch is folded into the expiration select as a **Never** option (default: 3 months — previously an untouched select silently created a never-expiring token).
- The token list is now a **data table** (Name / Created / Expires / Actions) with skeleton, empty, and error states. Expired tokens get a destructive badge; non-expiring tokens show "Never".
- The delete confirmation dialog footer and copy now match the other confirm dialogs (right-aligned natural-width buttons, "Delete" instead of "I'm sure! Delete it").
- Removed the now-unused `forms/token.tsx`.

## Before / After

| Before | After |
| --- | --- |
| before | after |

### Create token dialog

create dialog

### Token reveal

reveal dialog

### Delete confirmation

delete dialog

## Testing

- Exercised the full flow against the local dev server: create (including empty-name validation), token reveal + copy, table refresh, and delete confirmation.
- `npx tsc --noEmit -p apps/remix` and Biome clean.


apps/remix/app/routes/_authenticated+/t.$teamUrl+/settings.tokens.tsx
```diff
@@ -18,33 +22,88 @@ export function meta() {
...
+
+  const {
+    data: tokens,
+    isLoading,
+    isError,
+  } = trpc.apiToken.getMany.useQuery(undefined, {
+    enabled: !isUnauthorized,
+  });
+
+  const columns = useMemo(() => {
```
catalinpit, line 39: I'd say we should add a `last used` property to make it easier to see unused tokens. What do you think?

dguyen (approved):

PR #3081, opened 2026-07-08
fix: increase global API rate limits to 1000/min

Mythie:


dguyen (approved):

dguyen:
trpc should still stay on 100 imo

Mythie:
Yeah agree

Issue #3038, opened 2026-06-26
Document completion emails fail all recipients on one sendMail error


xuezhaolan:
### Problem

The `send.document.completed.emails` job sends completion emails with `Promise.all(...)` over every recipient. If one `emailTransport.sendMail(...)` call rejects, the entire job rejects immediately and later recipients are never processed.

That means a single bad recipient address, transient SMTP failure, or transport error can stop completion emails for unrelated recipients even though the document is already sealed and completed.

### Why this matters

Documenso already treats some adjacent completion side effects as independent:
- the completion webhook is triggered before the email job is queued
- CC recipients who exceed quota are skipped so other recipients can still be processed

But the main recipient completion-email fan-out still fails closed for the whole batch.

### Current code path

`packages/lib/jobs/definitions/emails/send-document-completed-emails.handler.ts`

The job currently does:

```ts
await Promise.all(
  recipientsToNotify.map(async (recipient) => {
    await emailTransport.sendMail(...)
    await prisma.documentAuditLog.create(...)
  }),
);
```

### Suggested fix

Switch the recipient fan-out to `Promise.allSettled(...)` (or equivalent per-recipient isolation), preserve successful sends/audit logs, and only surface a failure if **every** recipient send failed.

That would align completion-email behavior with the partial-success handling already added for reminder resends.

### Acceptance criteria

- one recipient send failure does not prevent other recipients from receiving completion emails
- successful sends still create their audit logs
- if all recipient sends fail, the job still reports a failure
- regression test covers partial success vs total failure

### Notes

I already checked that there is no open PR for this specific failure mode.

shivv-byri:
Hi! I'd like to work on this issue.

From my understanding, the problem is that if sending the completion email fails for one recipient, the entire email sending process stops. My plan is to investigate the email dispatch logic, isolate failures (e.g., using proper error handling or `Promise.allSettled()` where appropriate), and ensure successful recipients still receive their emails while logging failed deliveries.

If no one else is currently working on this, I'd be happy to take it. Thanks!

Rehansanjay:
Cross-linking this with #3190, because the two together constrain the fix more than either does on its own.

#3190 documents that this handler has no per-recipient idempotency: a rejection fails the whole job, the job replays **from the first recipient** (`maxRetries = 3`), and the recipients who already received the signed contract get it again — up to four times.

[...]

PR #3082, opened 2026-07-08
fix: use Promise.allSettled for completion emails to prevent total fa…

Rehansanjay:
…ilure on single error (#3038)

## Changes Made

- Change 1
- Change 2
- ...

## Testing Performed

- ...

## Checklist

### DIFF
--- a/packages/lib/jobs/definitions/emails/send-document-completed-emails.handler.ts
+++ b/packages/lib/jobs/definitions/emails/send-document-completed-emails.handler.ts
@@ -177,7 +177,7 @@ export const run = async ({ payload, io }: { payload: TSendDocumentCompletedEmai
 
   const recipientsToNotify = envelope.recipients.filter((recipient) => isRecipientEmailValidForSending(recipient));
 
-  await Promise.all(
+  const results = await Promise.allSettled(
     recipientsToNotify.map(async (recipient) => {
       // A CC recipient never asked to be part of this document, so their completion
       // email is effectively unsolicited. Meter it against the organisation email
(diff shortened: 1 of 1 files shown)


Mythie:

https://documen.so/cla

Rehansanjay:
> 
> https://documen.so/cla

i fill out the form mate is there any forums or something related to contributing to documenso out there if yes tell me i will learn that before contributing :)

Rehansanjay:
Following up — this one has been quiet since July.

**On the CLA:** @Mythie I filled in the form at documen.so/cla back on 8 July. If it didn't register against this account, let me know what you need and I'll redo it.

**What the change does**, since it's small enough to read in one sitting: `send-document-completed-emails.handler.ts` sent the completion emails under `Promise.all`, so the first rejection abandoned every remaining recipient. One bad address or a transient SMTP error, and the rest of that document's recipients silently got nothing.

It now uses `Promise.allSettled` and splits the two failure cases deliberately:

- **all** recipients failed — throw, so the job still surfaces and retries exactly as it did before
- **some** failed — log each one with the `envelopeId` and let the successful sends stand, rather than discarding good sends on a retry

+17/-1 in a single file, and all checks are green.

Happy to adjust either half of that: add a test, or make the partial-failure path do something other than log — count it, re-queue just the failed recipients, whatever fits how you'd rather see it handled.

Rehansanjay:
Closing this myself, for two reasons.

The first is policy: `CONTRIBUTING.md` now states that Documenso no longer accepts external pull requests and asks for detailed issues instead. Leaving this open just adds to a queue that is closed by design.

The second is that I no longer think the patch is right, and that matters more. In [#3038](https://github.com/documenso/documenso/issues/3038) I cross-linked [#3190](https://github.com/documenso/documenso/issues/3190): this handler has no per-recipient idempotency, and the job replays from the *first* recipient on failure with `maxRetries = 3`. So swapping `Promise.all` for `Promise.allSettled` does stop one bad address blocking the rest, but it forces the partial-failure branch into a choice where both answers are wrong without a per-recipient marker — rethrow and the recipients who already received the signed document get it again, up to four times; swallow and a genuinely undelivered completion email disappears with only a log line. This PR takes the second option. That is a real behaviour change to sold-document delivery, and it should not land on the strength of a three-line diff from someone outside the team.

The analysis lives on #3038, which is where Documenso has asked for it. @shivv-byri claimed that issue in June — the idempotency constraint is the part worth knowing before implementing it.

No action needed here; the issue is the contribution.

PR #3083, opened 2026-07-08
fix: auto-fit inserted field text so v2 sealing never clips values

reeseherber:
## Summary

FS-162157 (Zandra Craig): sealed contracts showed only the recipient's first name in NAME fields. The DB stored the full value (`customText = "Amanda Haydon"`) but the sealed PDF rendered `Amanda`; the DATE field on the same contract dropped its trailing ` PM`.

**Root cause:** internalVersion=2 envelopes seal fields through the Konva canvas overlay (`insertFieldInPDFV2`). `upsertFieldText` painted `customText` at the configured font size (typically 12) into a fixed-size `Konva.Text` with `wrap: 'word'` — Konva silently drops wrapped lines that exceed the fixed height. The v1 pdf-lib path scales text to fit, and typed signatures already auto-fit, so generic text fields were the only lossy path.

**Fix:** when a real value is rendered (inserted, or readOnly-prefill), shrink the font via `fitFontSize` (floor `MIN_STANDARD_FONT_SIZE = 8`) using a probe that mirrors the painted node's `wrap`/`lineHeight`/`letterSpacing`/font family, per the explicit-measurer gotcha. Placeholder/label rendering is untouched.

**Blast radius on prod (estimate):** ~647 completed envelopes with clipped DATE fields and ~89 with clipped NAME fields since 2026-04-01 go-live. Per Reese: no re-sealing — fix forward only. DB data was never lost.

## Verification

- New repro test `render-generic-text-field.test.ts` with the exact prod field dimensions — failed with exactly `"Amanda"` before the fix, green after (24/24 in the field-renderer suite).
- Exercised the real `insertFieldInPDFV2` seal path with the prod field rows: overlay PDF now contains `Amanda Haydon` and `2026-07-06 04:45 PM` in full, inside the field boxes.
- Adversarial review: 3 lenses (mode/field-type regressions, measurement fidelity, production-fit) — all approve; the one nit (duplicated width expression) is fixed.

FS-162157


reeseherber:
Opened against the wrong base repository — this belongs on a downstream fork. Apologies for the noise.

PR #3094, opened 2026-07-14
fix: missing noreply email for resend

dguyen:
## Description

Fix issue where resend transport does not support headers and no-reply settings.

## Changes

- Bump @documenso/nodemailer-resend
- Bump nodemailer to match peer dep for @documenso/nodemailer-resend


roshboi (approved):
lgtm

PR #3101, opened 2026-07-16
fix: use document title for single-item envelope downloads (#3096)

### LINKED ISSUE #3096: Document download filename uses the envelope item title (template name) instead of the document title
### Issue Description

Downloading a document from the **Download Files** dialog names the saved file after the **envelope item** title, not the **document (envelope)** title.

The envelope item title is set when the item is created and is **never updated afterwards**:

- Renaming a document updates `envelope.title` only — envelope item titles are left untouched.
- A document generated from a template inherits the **template's** item title.

So any renamed document — and in particular a document generated from a template and given a custom title — downloads with a filename that no longer matches the document's title. For template-generated documents this surfaces as "the download is named after the template".

This is inconsistent with every other document-level download (bulk document PDF, audit log, signing certificate), which all use `envelope.title`.

### Steps to Reproduce

1. Create a template containing a single PDF (e.g. template titled `Agency Agreement Template`).
2. Create a document from that template.
3. Rename the document to a custom title, e.g. `Wilson_Management and Agency Agreement`.
4. Complete signing so the document is `COMPLETED`.
5. Open the **Download** dialog and click **Signed**.

Observed: the downloaded file is named after the template's item title, not `Wilson_Management and Agency Agreement`. The dialog's file list also shows the stale item title.

This also reproduces **without templates**:

1. Upload `original-upload.pdf` to create a document.
2. Rename the document to `Some Custom Title`.
3. Download it → the file is still named `original-upload*.pdf`.

### Expected Behavior

The downloaded filename should reflect the document's **current title** (e.g. `Wilson_Management and Agency Agreement_signed.pdf`), consistent with the audit-log / certificate / bulk-document downloads, which already use `envelope.title`.

### Current Behavior

The filename is derived from `envelopeItem.title`, which reflects the template's item title (or the originally uploaded filename) at creation time.

### Root cause / relevant code

- `apps/remix/app/components/dialogs/envelope-download-dialog.tsx` passes `fileName: envelopeItem.title` into `downloadPDF`.
- `packages/lib/client-only/download-pdf.ts` → `download-file.ts` builds an `<a download="...">`, whose value takes precedence over the server's `Content-Disposition`. So the client-supplied item title is what names the saved file.
- `packages/lib/server-only/envelope/update-envelope.ts` updates only `title` on the envelope when a document is renamed; envelope item titles are not touched.
- `packages/lib/server-only/template/create-document-from-template.ts` sets each generated item's title via `const titleToUse = item.title || finalEnvelopeTitle;` — i.e. the template's item title wins, and the document title (`override?.title || template.title`) is only a fallback. The same pattern exists in `create-document-from-direct-template.ts`.
- The server-side per-item download routes have the same behaviour: `apps/remix/server/api/files/files.ts` and `apps/remix/server/api/download/download.ts` pass `title: envelopeItem.title` into `handleEnvelopeItemFileRequest`, which builds the `Content-Disposition` filename in `apps/remix/server/api/files/files.helpers.ts`.
- For contrast, `apps/remix/server/api/download/download.ts` uses `const baseTitle = envelope.title.replace(/\.pdf$/i, '')` for the audit-log and certificate downloads.

### Possible fix

For a **single-item** envelope the item *is* the document, so the download filename (and the dialog's displayed name) can use the envelope title, while **multi-item** envelopes keep their per-item titles so the files stay distinct.

We've verified a minimal change along these lines in a fork (returning the envelope title from the `envelope.item.getManyByToken` tRPC route, which already loads the envelope, and using it in the dialog when `envelopeItems.length === 1`); it type-checks and builds. Aligning the server-side per-item `Content-Disposition` in `files.ts` / `download.ts` would make direct-URL downloads consistent too.

Happy to provide more detail if useful — noting the contributing guide asks for issues rather than external PRs.

### Version

2.15.0 (also reproduces on `main`)

martindglaser:
## Description

Downloading a document from the "Download Files" dialog named the file after the envelope **item** title, which preserves the original template/upload name. Renaming a document updates the envelope title but not the item title, so downloads kept a stale name.

## Related Issue

Fixes #3096

## Changes Made

- `envelope.item.getManyByToken` now returns the parent `envelopeTitle` (`get-envelope-items-by-token.ts` / `.types.ts`).
- New pure helper `packages/lib/utils/get-envelope-download-filename.ts`: single-item envelopes download using the envelope (document) title; multi-item envelopes keep per-item titles so files remain distinguishable.
- `envelope-download-dialog.tsx` uses the helper (no change needed at the dialog's ~11 call sites).

## Testing Performed

- `vitest run utils/get-envelope-download-filename.test.ts` → **3 passed** (single-item → document title, multi-item → item title, empty-title fallback).
- `biome check` → clean.

## Checklist

### DIFF
--- a/apps/remix/app/components/dialogs/envelope-download-dialog.tsx
+++ b/apps/remix/app/components/dialogs/envelope-download-dialog.tsx
@@ -1,4 +1,5 @@
 import { downloadPDF } from '@documenso/lib/client-only/download-pdf';
+import { getEnvelopeDownloadFileName } from '@documenso/lib/utils/get-envelope-download-filename';
 import { trpc } from '@documenso/trpc/react';
 import { Button } from '@documenso/ui/primitives/button';
 import {
@@ -97,12 +98,13 @@ export const EnvelopeDownloadDialog = ({
       access: token ? { type: 'recipient', token } : { type: 'user' },
     },
     {
-      initialData: initialEnvelopeItems ? { data: initialEnvelopeItems } : undefined,
+      initialData: initialEnvelopeItems ? { data: initialEnvelopeItems, envelopeTitle: '' } : undefined,
       enabled: open,
     },
   );
 
   const envelopeItems = envelopeItemsPayload?.data || [];
+  const envelopeTitle = envelopeItemsPayload?.envelopeTitle ?? '';
 
   const onDownload = async (envelopeItem: EnvelopeItemToDownload, version: 'original' | 'signed' | 'pending') => {
     const { id: envelopeItemId } = envelopeItem;
@@ -120,7 +122,11 @@ export const EnvelopeDownloadDialog = ({
       await downloadPDF({
         envelopeItem,
         token,
-        fileName: envelopeItem.title,
+        fileName: getEnvelopeDownloadFileName({
+          itemCount: envelopeItems.length,
+          itemTitle: envelopeItem.title,
+          envelopeTitle,
+        }),
         version,
       });
 
--- a/apps/remix/app/components/dialogs/envelope-download-dialog.tsx
+++ b/apps/remix/app/components/dialogs/envelope-download-dialog.tsx
@@ -21,6 +21,14 @@ type EnvelopeItemToDownload = Pick<EnvelopeItem, 'id' | 'envelopeId' | 'title' |
 
 type EnvelopeDownloadDialogProps = {
   envelopeId: string;
+
+  /**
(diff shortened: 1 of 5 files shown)


martindglaser:
@coderabbitai review

catalinpit:
Hey @martindglaser, thanks for the PR! We'll consolidate yours and https://github.com/documenso/documenso/pull/3111 in https://github.com/documenso/documenso/pull/3374.

PR #3122, opened 2026-07-24
fix: compare grouped number field values against min/max correctly

eeshsaxena:
A number field with a thousands separator is compared against `minValue`/`maxValue` using only its leading group.

## Problem

`validateNumberField` calls `parseFloat` on the raw value, but every configured number format uses a separator `parseFloat` stops at:

```js
parseFloat("1,234.56")  // 1        (format 123,456,789.00)
parseFloat("1.234,56")  // 1.234    (format 123.456.789,00)
```

The value passes its format regex, then the range check runs against `1` instead of `1234.56`. Both directions are wrong:

| field config | entered | today | expected |
|---|---|---|---|
| `maxValue: 100` | `1,234.56` | accepted | rejected |
| `minValue: 500` | `1,234.56` | rejected | accepted |

`sign-field-with-token.ts` calls this with `isSigningPage: true`, so this runs on the value a recipient submits — a signed field can end up holding a number outside the range the sender configured.

## Fix

Normalise the value before comparing, using the format to decide which separator groups and which is the decimal:

- `123.456.789,00` — dots group, comma is the decimal
- `123,456,789.00`, `123456,789.00`, and an unset format — commas group, dot is the decimal

Only the parsing changes; the format regex, the required/read-only rules and the error messages are untouched.

## Tests

`validate-number.ts` had no test file. Added one covering the parser directly and the range checks for grouped, dot-grouped and ungrouped values. I verified 4 of the 8 fail against the current code and all pass with the fix; the full `packages/lib` suite is green (193 tests).

## Note

`minValue`/`maxValue` are also guarded by `> 0`, so a limit of exactly `0` is ignored. That looked like a separate decision so I left it alone.

### DIFF
--- a/packages/lib/advanced-fields-validation/validate-number.test.ts
+++ b/packages/lib/advanced-fields-validation/validate-number.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from 'vitest';
+
+import type { TNumberFieldMeta as NumberFieldMeta } from '../types/field-meta';
+import { parseNumberFieldValue, validateNumberField } from './validate-number';
+
+const meta = (overrides: Partial<NumberFieldMeta>): NumberFieldMeta =>
+  ({ type: 'number', ...overrides }) as NumberFieldMeta;
+
+describe('parseNumberFieldValue', () => {
+  it('parses a plain decimal', () => {
+    expect(parseNumberFieldValue('1234.56')).toBe(1234.56);
+  });
+
+  it('parses a comma-grouped value', () => {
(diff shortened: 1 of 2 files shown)


Mythie:

https://documen.so/cla

eeshsaxena:
I have read the CLA Document and I have signed the CLA

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #3125, opened 2026-07-25
fix(email): use the recipient's role in the completed notification instead of always saying "signed"

daggy72:
### Problem

When a recipient completes their step, the notification sent to the document owner always reads **"{name} has signed \"{document}\""** — including when that recipient's role is `APPROVER` or `VIEWER`, and no signature fields were ever bound to them.

The invite and reminder emails in the same package already branch on `RecipientRole` and say "approve"/"view" correctly. This is the one recipient-facing outcome email that doesn't, so an approver's action is reported to the owner as a signature.

### Change

`recipientRole` is threaded from the job handler into the template. The handler already loads the `recipient` record, so **no additional query is needed**.

Rather than adding a new `match` block, this reuses the existing `RECIPIENT_ROLES_DESCRIPTION` constant, so the wording comes from the already-translated `actioned` and `progressiveVerb` entries:

| Role | Headline | Body |
|---|---|---|
| `APPROVER` | has **approved** | has completed **approving** |
| `VIEWER` | has **viewed** | has completed **viewing** |
| `SIGNER` | has **signed** | has completed **signing** *(unchanged)* |

Applied to all four places the verb appears for this email: **subject**, **preview text**, **headline** and **body**.

### Compatibility

`recipientRole` is optional and defaults to `RecipientRole.SIGNER`, so existing callers and the email preview app are unaffected, and the default output is byte-identical to today's for signers.

Because it reuses `actioned` / `progressiveVerb`, this adds **no new translatable strings** beyond the two reworded sentences.

Files touched:
- `packages/lib/jobs/definitions/emails/send-recipient-signed-email.handler.ts`
- `packages/email/templates/document-recipient-signed.tsx`
- `packages/email/template-components/template-document-recipient-signed.tsx`

Happy to adjust the wording or split the subject-line change out if you'd prefer a narrower diff.

### DIFF
--- a/packages/email/template-components/template-document-recipient-signed.tsx
+++ b/packages/email/template-components/template-document-recipient-signed.tsx
@@ -1,4 +1,7 @@
+import { RECIPIENT_ROLES_DESCRIPTION } from '@documenso/lib/constants/recipient-roles';
+import { useLingui } from '@lingui/react';
 import { Trans } from '@lingui/react/macro';
+import { RecipientRole } from '@prisma/client';
 
 import { Column, Img, Section, Text } from '../components';
 import { TemplateDocumentImage } from './template-document-image';
@@ -7,6 +10,7 @@ export interface TemplateDocumentRecipientSignedProps {
   documentName: string;
   recipientName: string;
   recipientEmail: string;
+  recipientRole?: RecipientRole;
   assetBaseUrl: string;
 }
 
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

daggy72:
sorry, missed that part, CLA no accepted!

daggy72:
Small correction to my earlier comment — that was a typo, it should have read "CLA **now** accepted". The CLA was signed on 28 July via documen.so/cla. Happy to re-submit if it did not register on your side.

All checks are green; the PR just needs a maintainer look when there is a moment.

PR #3137, opened 2026-07-30
docs(getting-started): fix distribute route, pagination and retry advice

ephraimduncan:
## Description

The getting-started tutorial's distribute step 404s: it calls `POST /envelope/{id}/distribute`, which does not exist.

## Changes Made

- Fixed all four distribute calls to the real route: `POST /envelope/distribute` with `envelopeId` in the body.
- Fixed the list-response example from the fabricated nested `pagination` object to the real flat shape `{ data, count, currentPage, perPage, totalPages }`.
- Replaced the fixed 60-second retry advice and hardcoded 60000ms sleep with `Retry-After` header handling; mentioned `X-RateLimit-Remaining`/`X-RateLimit-Reset`.

## Testing Performed

Docs-only change. Verified against `distribute-envelope.types.ts`, `find-envelopes.types.ts`, `create-envelope.types.ts`, and the rate-limit middleware.


dguyen (approved):

dguyen:
@ephraimduncan please fix merge conflict

PR #3158, opened 2026-08-09
fix(envelope): keep same-email recipients distinct when resolving PDF placeholder tags

### LINKED ISSUE #3115: Title: [Bug]: Duplicate-email recipients collapse onto one recipient when resolving PDF placeholder tags ({{signature, rN}})
### Issue Description

When creating an envelope via `POST /api/v2/envelope/create` with an explicit `recipients` array where two or more recipients share the same `email` (but different `signingOrder`/`name`), every PDF placeholder tag (`{{signature, rN}}`, `{{date, rN}}`, etc.) resolves to the **same** recipient, regardless of which `rN` index the tag actually references. All fields for every tagged recipient end up bound to a single `recipientId` — in practice, whichever recipient row shares that email and got matched first.

This makes it impossible to build a single-document flow where one person needs to sign in more than one place/step (e.g. the same signer countersigning at steps 1, 4 and 6) using placeholder-tagged PDFs, without giving that signer distinct alias emails.

### Steps to Reproduce

1. Have a PDF with placeholder tags for at least 2 recipients, e.g. `{{signature, r1}}` and `{{signature, r2}}` at different positions.
2. `POST /api/v2/envelope/create` (multipart) with:
   ```json
   {
     "type": "DOCUMENT",
     "title": "repro",
     "recipients": [
       { "email": "dup@example.com", "name": "A", "role": "SIGNER", "signingOrder": 1 },
       { "email": "dup@example.com", "name": "B", "role": "SIGNER", "signingOrder": 2 }
     ],
     "meta": { "signingOrder": "SEQUENTIAL" }
   }
   ```
   (attach the tagged PDF as `files`)
3. `GET /api/v2/envelope/{id}` and inspect `fields[].recipientId`.

### Expected Behavior

Two fields are created: one bound to the recipient with `signingOrder: 1`, the other to `signingOrder: 2` 

matching the `r1`/`r2` tags, exactly as happens when the two recipients have distinct emails.

### Current Behavior

Both fields are created bound to the *same* `recipientId`. The second recipient gets no fields at all.

### Screenshots (optional)

_No response_

N/A (self-hosted Docker deployment, Linux host)  reproduced purely via REST API, not a client-UI issue.

N/A — reproduced via REST API.

### Version [e.g., 2.13.0]

2.15.0

eeshsaxena:
Fixes #3115.

## Problem

When creating an envelope with two or more recipients that share the same `email` (but a different `signingOrder`/`name`), every PDF placeholder tag (`{{signature, r1}}`, `{{signature, r2}}`, …) resolves to the same recipient, and the later recipients get no fields.

`findRecipientByPlaceholder` already resolves tags by index when it is handed a per-index `recipients` array, and that part is correct. The problem is how `create-envelope` builds that array:

```ts
data.recipients.map((r) => {
  const found = availableRecipients.find((cr) => cr.email === r.email)
  ...
  return found
})
```

`availableRecipients` is re-fetched from the DB, and each input recipient is matched back to it by email. For two recipients with the same email, `find` returns the same first row for both, so the resulting array is `[rowA, rowA]` and `r1`/`r2` both land on `rowA`.

## Fix

Capture the created recipients from the `Promise.all` that inserts them. `Promise.all` resolves in input order, so `createdRecipientRecords[i]` is exactly the recipient created from `data.recipients[i]`. Resolving placeholder tags against that array by index removes the email round-trip, so same-email recipients stay distinct. The email-based lookup is still used for the placeholder-recipient path (no explicit recipients provided), where the generated emails are unique.

## Verifying

The exact repro in the issue — `POST /api/v2/envelope/create` with two `SIGNER` recipients sharing `dup@example.com` (signing order 1 and 2) and a PDF tagged `{{signature, r1}}` / `{{signature, r2}}` — now creates two fields bound to two different `recipientId`s, matching the behaviour when the emails differ.

### DIFF
--- a/packages/lib/server-only/envelope/create-envelope.ts
+++ b/packages/lib/server-only/envelope/create-envelope.ts
@@ -416,7 +416,12 @@ export const createEnvelope = async ({
 
     const allRecipients = [...(data.recipients || []), ...mappedDefaultRecipients];
 
-    await Promise.all(
+    // Promise.all resolves in input order, so createdRecipientRecords[i] is the
+    // recipient created from allRecipients[i] (and therefore from data.recipients[i]
+    // for the leading entries). Keeping this correspondence lets placeholder tags be
+    // resolved by index below instead of by email, which is what allows two recipients
+    // that share an email to stay distinct (#3115).
+    const createdRecipientRecords = await Promise.all(
       allRecipients.map(async (recipient) => {
         const recipientAuthOptions = createRecipientAuthOptions({
           accessAuth: recipient.accessAuth ?? [],
@@ -455,7 +460,7 @@ export const createEnvelope = async ({
           };
         });
 
-        await tx.recipient.create({
+        return tx.recipient.create({
           data: {
             envelopeId: envelope.id,
             name: recipient.name,
@@ -472,6 +477,7 @@ export const createEnvelope = async ({
               },
             },
           },
+          select: { id: true, email: true },
         });
       }),
     );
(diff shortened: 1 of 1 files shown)


eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #3165, opened 2026-08-10
fix(seal): preserve /FreeText annotations without an appearance stream

Austin519:
### Problem

When a document is filled in with a tool that adds text as **`/FreeText`
annotations without an appearance stream (`/AP`)** — which is what pypdf,
Stirling PDF, and most scripted PDF annotators produce — the **signed PDF comes
back blank** where the user typed.

The cause is in the seal path. `decorateAndSignPdf` calls
`pdfDoc.flattenAll()` (from `@libpdf/core`) before signing. Its annotation
flattener does, per page:

```
let appearance = annotation.getNormalAppearance();
if (!appearance) appearance = this.generateAppearance(annotation);
if (!appearance) { refsToRemove.add(...); continue; }  // <-- annotation deleted
```

and `generateAppearance()` only synthesises appearances for
Highlight / Underline / StrikeOut / Squiggly. A `/FreeText` annotation with no
`/AP` therefore falls through to the delete branch: its text never reaches the
page content stream and is absent from the sealed output.

This is why the symptom looks inconsistent depending on how the document was
filled:

| How the text got in | Result |
|---|---|
| Acrobat form-fill → AcroForm `/Widget` | Fine — handled by form flattening |
| Acrobat Typewriter → `/FreeText` **with** `/AP` | Fine — existing appearance is baked |
| pypdf / Stirling PDF / most scripted annotators → `/FreeText` **without** `/AP` | **Dropped** |

A `/FreeText` without `/AP` is spec-legal (PDF 32000-1 — viewers are expected to
render it from `/DA` + `/Contents`; Acrobat synthesises an appearance on open),
so silently deleting it at seal time is data loss rather than an unsupported
input.

### Fix

Add a pre-pass, `bakeAnnotationAppearances()`, that runs immediately **before**
`flattenAll()` in `decorateAndSignPdf`. For every annotation that carries
renderable content but has no `/AP`, it synthesises a standards-compliant
`/AP /N` Form XObject from `/Contents` + `/DA` + `/Rect` (greedy word-wrap,
`/DA` size-0 auto-fit, WinAnsi sanitising). `flattenAll()` then takes its normal
"bake the appearance into page content" path and the text survives.

It deliberately does **not** reimplement flattening — it only fills the gap that
makes the flattener give up. Design points:

- **Never throws.** On any failure it returns the original bytes, so a malformed
  PDF degrades to the current behaviour rather than breaking sealing.
- **Cheap no-op path.** If a page has no eligible annotations it returns the
  input untouched, without embedding a font or re-serialising.
- **Loud safety net.** Any annotation subtype it still cannot render logs a
  warning naming the subtype and count, because `flattenAll()` is about to
  delete it.
- Runs on the **seal** path, so already-uploaded documents are fixed
  retroactively and the stored original is left untouched.

### Test

`bake-annotation-appearances.test.ts` builds a PDF whose only typed content is
`/AP`-less `/FreeText` annotations, **reproduces the drop** through the same
`flattenAll()` the seal path uses, then asserts every value **survives** after
the pre-pass. Rendered text is asserted with **pdf.js** (the reference renderer
the app's viewer uses) rather than `@libpdf/core`'s `extractText()`, which does
not descend into the Form XObjects that flattened annotations become and would
report a false negative. A regression case asserts an annotation-free PDF is
returned unchanged.

### Notes

- No new dependencies — `@cantoo/pdf-lib` and `pdfjs-dist` are already in the
  tree.
- Scoped to `/FreeText` (the observed real-world case). Other appearance-less
  subtypes carrying content are surfaced by the warning rather than silently
  handled, so the change stays conservative.

### DIFF
--- a/packages/lib/jobs/definitions/internal/seal-document.handler.ts
+++ b/packages/lib/jobs/definitions/internal/seal-document.handler.ts
@@ -2,6 +2,7 @@ import path from 'node:path';
 import { PDFDocument } from '@cantoo/pdf-lib';
 import { finalizeTspEnvelopeCompletion } from '@documenso/ee/server-only/signing/csc/finalize-tsp-completion';
 import { addRejectionStampToPdf } from '@documenso/lib/server-only/pdf/add-rejection-stamp-to-pdf';
+import { bakeAnnotationAppearances } from '@documenso/lib/server-only/pdf/bake-annotation-appearances';
 import { generateAuditLogPdf } from '@documenso/lib/server-only/pdf/generate-audit-log-pdf';
 import { generateCertificatePdf } from '@documenso/lib/server-only/pdf/generate-certificate-pdf';
 import { getLastPageDimensions } from '@documenso/lib/server-only/pdf/get-page-size';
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

PR #3173, opened 2026-08-12
fix: allow zooming and dragging uploaded signatures

dguyen:
## Description

https://github.com/user-attachments/assets/3b8bee95-767d-469e-a1b1-2beefb465d5d

- Allow dragging and zooming for uploaded signatures
- Hide elements on the canvas when drawing


roshboi (approved):
very nice

PR #3175, opened 2026-08-12
fix: document editor on mobile view

catalinpit:
Demo:

https://github.com/user-attachments/assets/9a3a8c8e-2ea0-48db-8ebd-93b9cba955ea


bamin2:
Need this

PR #3178, opened 2026-08-13
chore: add translations

dguyen:
## Description

AI Translation dump, pending secondary review to validate that it is correct.


mKoonrad:
@dguyen These translations are not perfect. If you want a better Polish translation, I prepared #3210. I hope localization PRs are excluded from your strange PR policy. Only real, local users could validate it.

mKoonrad:
@dguyen Didn't you mean to exclude Polish from conflict, not German? 😅

dguyen:
@mKoonrad opps, thanks 😆

PR #3205, opened 2026-08-13
fix(multiselect): prevent stale async search results and duplicate re…

### LINKED ISSUE #3164: MultiSelect async search: stale results can overwrite newer ones, and every search fires twice
Filing as an issue rather than a PR, per the note at the top of CONTRIBUTING.md.

Two independent bugs in the async search effect of `packages/ui/primitives/multiselect.tsx` (around lines 279-305). Both are reachable today through `OrganisationMembersMultiSelectCombobox`, which passes an `onSearch` that hits `trpc.organisation.member.find` over the network and sets `triggerSearchOnFocus`.

```tsx
useEffect(() => {
  /** async search */

  const doSearch = async () => {
    setIsLoading(true);
    const res = await onSearch?.(debouncedSearchTerm);
    setOptions(transToGroupOption(res || [], groupBy));
    setIsLoading(false);
  };

  const exec = async () => {
    if (!onSearch || !open) {
      return;
    }

    if (triggerSearchOnFocus) {
      await doSearch();
    }

    if (debouncedSearchTerm) {
      await doSearch();
    }
  };

  void exec();
}, [debouncedSearchTerm, groupBy, open, triggerSearchOnFocus]);
```

### 1. The search is not cancelled, so stale results can win

The effect re-runs when `debouncedSearchTerm` changes, but nothing cancels the request already in flight and nothing guards the `setOptions` call. If the request for `jo` resolves after the request for `joh`, the dropdown ends up showing results for `jo` while the input reads `joh`.

The debounce (300ms for the members combobox) narrows the window but does not close it, since the ordering depends on server response time rather than on typing speed. `setIsLoading(false)` from the losing request also clears the spinner while the newer one is still running.

On a member picker this matters more than usual: the list is what the user picks from when adding people to a team, so a mismatched list is a wrong-person risk rather than only a cosmetic glitch.

There is also no unmount guard, so closing the dropdown mid-request calls `setOptions` and `setIsLoading` on an unmounted component.

### 2. Every non-empty search issues two identical requests

The two conditions are sequential `if`s rather than `if`/`else if`. When `triggerSearchOnFocus` is set *and* `debouncedSearchTerm` is non-empty, `doSearch()` is awaited twice in a row, so the same query goes to the server twice.

`OrganisationMembersMultiSelectCombobox` sets `triggerSearchOnFocus`, so as soon as the user types anything, every search costs two `organisation.member.find` calls instead of one. The second is pure waste, and it also widens the race window in (1).

The sync effect just above has the same double-call shape, but it is calling `onSearchSync` so it costs no network.

### Suggested fix

A cancellation flag plus cleanup handles (1), and making the branches mutually exclusive handles (2):

```tsx
useEffect(() => {
  let cancelled = false;

  const doSearch = async () => {
    setIsLoading(true);
    const res = await onSearch?.(debouncedSearchTerm);
    if (cancelled) {
      return;
    }
    setOptions(transToGroupOption(res || [], groupBy));
    setIsLoading(false);
  };

  const exec = async () => {
    if (!onSearch || !open) {
      return;
    }

    if (debouncedSearchTerm || triggerSearchOnFocus) {
      await doSearch();
    }
  };

  void exec();

  return () => {
    cancelled = true;
  };
}, [debouncedSearchTerm, groupBy, open, triggerSearchOnFocus]);
```

An `AbortController` threaded into `onSearch` would also tear the request down rather than just ignoring it, but that changes the `onSearch` prop signature and every caller, so the flag seemed like the proportionate fix.

I checked and there is no open PR touching this file. Happy to write the patch if a maintainer wants it, or to leave it here for the team.

Muzammil327:
## Description
Fixes a race condition and duplicate request behavior in the multiselect async search flow. Slow responses could overwrite newer results, and the search logic was issuing redundant requests for the same debounced term. This change also makes loading state recovery more reliable when requests fail.

## Related Issue
Fixes #3164

## Changes Made
- Added cancellation guards so stale async search responses cannot overwrite newer results.
- Removed the duplicate request path for non-empty debounced search terms.
- Removed the redundant `onFocus` search call that was immediately discarded.
- Applied the same cleanup to the synchronous search path for consistency.
- Wrapped async search in `try/catch/finally` so loading state is always cleared, even on failure.

## Testing Performed
- Manually tested multiselect search with fast and slow responses.
- Verified newer search results are not replaced by stale responses.
- Confirmed only one request is fired per debounced search term.
- Checked that loading state resets correctly after a failed request.

## Checklist

## Additional Notes
This change focuses on search correctness and request deduplication in multiselect. The main goal was to prevent stale async results from winning the race and to keep loading state consistent in all outcomes.

### DIFF
--- a/packages/ui/primitives/multiselect.tsx
+++ b/packages/ui/primitives/multiselect.tsx
@@ -259,33 +259,40 @@ const MultiSelect = ({
       setOptions(transToGroupOption(res || [], groupBy));
     };
 
-    // eslint-disable-next-line @typescript-eslint/require-await
-    const exec = async () => {
+    const exec = () => {
       if (!onSearchSync || !open) {
         return;
       }
 
-      if (triggerSearchOnFocus) {
-        doSearchSync();
-      }
-
-      if (debouncedSearchTerm) {
+      if (debouncedSearchTerm || triggerSearchOnFocus) {
         doSearchSync();
       }
     };
 
-    void exec();
+    exec();
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [debouncedSearchTerm, groupBy, open, triggerSearchOnFocus]);
 
   useEffect(() => {
     /** async search */
 
+    let cancelled = false;
+
     const doSearch = async () => {
       setIsLoading(true);
-      const res = await onSearch?.(debouncedSearchTerm);
-      setOptions(transToGroupOption(res || [], groupBy));
-      setIsLoading(false);
+      try {
+        const res = await onSearch?.(debouncedSearchTerm);
+        if (cancelled) {
+          return;
+        }
+        setOptions(transToGroupOption(res || [], groupBy));
+      } catch {
+        // Swallow search errors; isLoading is reset in the finally block.
+      } finally {
+        if (!cancelled) {
+          setIsLoading(false);
+        }
+      }
     };
 
     const exec = async () => {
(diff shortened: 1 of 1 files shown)


Mythie:

https://documen.so/cla

PR #3216, opened 2026-08-14
fix: validate required fields before stamping DATE fields

### LINKED ISSUE #3193: DATE fields are stamped before required-field validation and outside the transaction, freezing a wrong date
### Issue Description

In `complete-document-with-token.ts`, unfilled DATE fields are stamped and audit-logged **before** the required-field validation runs, and **outside** the transaction:

| Line | Operation |
|---|---|
| 189 | `uninsertedDateFields = fields.filter(type === DATE && !inserted)` |
| 221 | `if (envelope.internalVersion === 2 && uninsertedDateFields.length > 0)` |
| **231** | `await prisma.field.updateMany(...)` → **persisted** |
| **243** | `await prisma.documentAuditLog.createMany(...)` → `DOCUMENT_FIELD_INSERTED` |
| **281** | `if (fieldsContainUnsignedRequiredField(fields)) throw` |
| **285** | `await prisma.$transaction(...)` |

Three things make this more than a style detail:

1. Validation (281) comes after the write (231).
2. The write is **outside the transaction** — `$transaction` only starts at 285, so the throw at 281 rolls nothing back.
3. **The timestamp is never corrected.** On the next, successful attempt `uninsertedDateFields` is empty, so the block is skipped. The document permanently carries the date of the **failed** attempt.

### Steps to Reproduce

1. Create a V2 envelope with a DATE field and at least one required field that is left unfilled.
2. Call `completeDocumentWithToken` (the tRPC route, authenticated by the recipient token). It fails validation, as intended.
3. Days later, fill the required field and complete successfully.
4. Look at the date rendered in the sealed PDF.

### Expected Behavior

The date reflects when the document was actually signed.

### Current Behavior

It reflects the first, failed attempt — days earlier.

### Version

`main` @ 617f8cc (v2.16.0)

### Notes

Found during a code audit; verified by reading the ordering and confirming the transaction boundary on current `main`. Not reproduced against a running instance.

Scope, honestly: V2 only (the block is gated at 221), and the UI disables Complete while required fields are missing — so this is not reachable by ordinary clicking. But `completeDocumentWithToken` is a public procedure authenticated by the recipient token (`recipient-router/router.ts:588`), and the value at stake is the date printed on a signed document.

Suggested direction: move the validation above the DATE block, and/or move the writes inside the transaction that starts at 285.

Souptik96:
Fixes #3193

## Root cause

In `packages/lib/server-only/document/complete-document-with-token.ts`, the auto-insert block for un-inserted `DATE` fields ran **before** the required-field validation and **outside** the transaction that performs the rest of the signing completion:

1. `prisma.field.updateMany(...)` stamped the date and marked the fields `inserted`.
2. `prisma.documentAuditLog.createMany(...)` wrote `DOCUMENT_FIELD_INSERTED` entries.
3. Only then did `fieldsContainUnsignedRequiredField(fields)` throw for a missing required field.
4. `prisma.$transaction(...)` started after that, so the throw in step 3 rolled nothing back.

Because step 1 already set `inserted: true`, the next (successful) attempt computed an empty `uninsertedDateFields` and skipped the block entirely. The timestamp was never corrected, so the document permanently carried the date of the **failed** attempt.

The reason the original code was ordered this way is that `DATE` is not in `ADVANCED_FIELD_TYPES_WITH_OPTIONAL_SETTING`, so `isRequiredField` treats every un-inserted `DATE` field as unsigned-and-required. The old code satisfied the validation by mutating the local `fields` array *after* persisting. A naive reorder would therefore have made every V2 envelope with a `DATE` field fail validation.

## Fix

- Validation now runs **first**, against `fieldsForValidation` — an in-memory view of `fields` in which exactly the auto-insertable `DATE` fields are treated as `inserted`. Nothing is persisted before validation, and the set treated as inserted is exactly the set that later gets stamped, so validation semantics are unchanged.
- The stamping and its audit log entries moved **inside** the existing `$transaction`, placed after the concurrency guard so a losing concurrent request aborts before doing extra writes. `DateTime.now()` is evaluated there, so the stamp reflects the attempt that actually succeeds.
- `fields` is now `const`, since the post-write reassignment is gone.

Behaviour is unchanged for V1 envelopes (`shouldAutoInsertDateFields` is `false`, so `fieldsForValidation` is element-identical to `fields`) and for the normal successful path.

## Verification

Added `packages/lib/server-only/document/complete-document-with-token.test.ts`, which exercises the flow against an in-memory fake of the Prisma surface this function touches. The fake models the property that matters: writes through the transaction client are rolled back when the callback throws, writes through the base client are not, and each write records which client it came through.

Four tests, each of which **fails on `main` and passes with this change**:

| Test | On `main` | With fix |
|---|---|---|
| DATE field not stamped when another required field is unsigned | `inserted` was `true` | passes |
| DATE stamp goes through the transaction client | write recorded `via: 'base'` | `via: 'tx'` |
| DATE stamp rolled back when the transaction fails later | `inserted` stayed `true` | rolled back |
| Date is that of the successful attempt, not an earlier failed one | got `2026-03-01` (failed attempt) | `2026-03-04` |

The last one reproduces the exact symptom from the issue, using fake timers to place the failed and successful attempts three days apart.

```
# negative control, before the fix
Test Files  1 failed (1)
     Tests  4 failed (4)

# after the fix — full packages/lib suite
Test Files  10 passed (10)
     Tests  194 passed (194)
```

`npx biome check` is clean (exit 0, no errors or warnings) on both touched files, and `tsc --noEmit` reports no errors in them — the 6 remaining errors in `packages/lib` are pre-existing on `main` and in unrelated files (`server-only/ai/google.ts`, `server-only/field/get-completed-fields-for-document.ts`, `server-only/subscription/get-active-subscriptions-by-user-id.ts`, `server-only/team/find-organisation-invoices.ts`).

### Not verified

I did not run this against a live instance or the Postgres-backed e2e suite — Docker is unavailable in the environment I worked in, so `docker compose -f docker/development/compose.yml up` and therefore `test:e2e` could not run. The transactional and ordering behaviour is covered by the unit tests described above rather than by an integration test; a reviewer with a database to hand may want to confirm the happy path end-to-end.

### DIFF
--- a/packages/lib/server-only/document/complete-document-with-token.test.ts
+++ b/packages/lib/server-only/document/complete-document-with-token.test.ts
@@ -1,0 +1,370 @@
+import { DocumentStatus, FieldType, RecipientRole, SigningStatus } from '@prisma/client';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * Regression tests for https://github.com/documenso/documenso/issues/3193
+ *
+ * Auto-inserted DATE fields used to be stamped (and audit logged) *before* the
+ * required-field validation ran, and *outside* the transaction that performs the
+ * rest of the signing completion. A failed completion attempt therefore froze the
+ * date of that failed attempt onto the document permanently, because the field was
+ * already marked `inserted` and so was skipped on every later attempt.
+ *
+ * These tests use an in-memory fake of the parts of the Prisma client the flow
+ * touches. The fake models the one property that matters here: writes made through
+ * the transaction client are rolled back if the transaction callback throws, while
+ * writes made through the base client are not.
+ */
+
+type FakeField = {
+  id: number;
+  secondaryId: string;
+  envelopeId: string;
+  recipientId: number;
+  type: FieldType;
+  inserted: boolean;
+  customText: string;
+  fieldMeta: unknown;
+  page: number;
+  positionX: unknown;
+  positionY: unknown;
+  width: unknown;
+  height: unknown;
+};
+
+type WriteRecord = {
+  /** 'base' = outside any transaction, 'tx' = through the transaction client. */
+  via: 'base' | 'tx';
+  model: string;
+  operation: string;
+};
+
+const ENVELOPE_ID = 'envelope_test3193';
+const RECIPIENT_ID = 42;
+const TOKEN = 'test-token';
+
+const buildField = (overrides: Partial<FakeField> & Pick<FakeField, 'id' | 'type'>): FakeField => ({
+  secondaryId: `field_${overrides.id}`,
+  envelopeId: ENVELOPE_ID,
+  recipientId: RECIPIENT_ID,
+  inserted: false,
+  customText: '',
+  fieldMeta: null,
+  page: 1,
+  positionX: 0,
+  positionY: 0,
+  width: 10,
+  height: 10,
+  ...overrides,
+});
+
+/**
+ * Builds the fake Prisma client plus the mutable state it reads and writes.
+ */
+const createFakePrisma = (options: {
+  fields: FakeField[];
+  /** Forces a failure inside the completion transaction, after the recipient update. */
+  failInsideTransaction?: boolean;
+}) => {
+  const state = {
+    fields: options.fields.map((field) => ({ ...field })),
+    auditLogTypes: [] as string[],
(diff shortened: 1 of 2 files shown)


Mythie:

https://documen.so/cla

PR #3220, opened 2026-08-15
refactor(trpc): type openapi fetch handler and form-data parsing

ephraimduncan:
## Description

The OpenAPI fetch handler and the form-data utilities now compile without `unknown` casts. The compiler now checks the request proxy and the node-http bridge, which the previous double casts turned off. Runtime behavior is unchanged.

## Related Issue

None. This is standalone cleanup.

## Changes Made

- The multipart and url-encoded body parsers accumulate values as `FormDataEntryValue` types instead of `Record<string, unknown>`.
- The request proxy trap reads properties through `keyof Request` instead of a blanket record cast.
- The code declares the node-http handler against the fetch-based types it passes, which removes the two `as unknown as` casts.
- `zodFormData` constrains its schema generic through `Parameters<typeof z.object>[0]` and drops the deprecated `ZodRawShape` import.

## Testing Performed

- I ran `npm run build` on this branch alone. All 5 turbo tasks passed.
- I ran `npx tsc --noEmit -p packages/trpc`. No errors.

## Checklist

### DIFF
--- a/packages/lib/translations/de/web.po
+++ b/packages/lib/translations/de/web.po
@@ -312,6 +312,12 @@ msgstr "{0}-Feld"
 msgid "{0} has invited you to {recipientActionVerb} the document \"{1}\"."
 msgstr "{0} hat Sie eingeladen, das Dokument \"{1}\" {recipientActionVerb}."
 
+#. placeholder {0}: organisation.name
+#: apps/remix/app/components/general/settings-upsell/branding-upsell.tsx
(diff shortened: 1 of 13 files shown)


Mythie:
Is this correct? The OpenAPI fetch handler is kind of cursed in that it forces its way into something tRPC compatible due to versioning issues way back when.

I'd hate for this to miss something on one of our public API's

PR #3221, opened 2026-08-15
refactor(trpc): validate x-team-id header with zod in client

ephraimduncan:
## Description

The tRPC client now validates the team id before it sets the `x-team-id` header. An empty string no longer produces an empty header in the single-operation path. The batch path already used a truthiness check, thus only the single-operation path had this problem.

## Related Issue

None. This is standalone cleanup.

## Changes Made

- A zod schema (`z.string().min(1)`) checks `context.teamId` in the single-operation and batch link paths.
- This replaces the raw `typeof` checks in the two paths.

## Testing Performed

- I ran `npm run build` on this branch alone. All 5 turbo tasks passed.
- I ran `npx tsc --noEmit -p packages/trpc`. No errors.

## Checklist

### DIFF
--- a/packages/lib/translations/de/web.po
+++ b/packages/lib/translations/de/web.po
@@ -312,6 +312,12 @@ msgstr "{0}-Feld"
 msgid "{0} has invited you to {recipientActionVerb} the document \"{1}\"."
 msgstr "{0} hat Sie eingeladen, das Dokument \"{1}\" {recipientActionVerb}."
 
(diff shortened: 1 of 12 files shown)


Mythie:
- This could have been a `typeof x === 'string' && x.length > 1` since that's all the zod check does which is probably cheaper perf wise as well.
- This doesn't feel necessary, an empty header is fine and would be handled on the server anyway.

Issue #907, opened 2024-02-05
[Improvement] Fields should not be able to stack exactly on top of each other 


sumitbishti:
### Describe the improvement you are suggesting in detail

Fields should have some area exclusively for themself. So that they can be clicked by the signer. Right now if the Sender stack all the fields exactly on top of each other, the Recipient can only click on the top most field.

All the fields are stacked on top of each other in the below img
<img width="560" alt="image" src="https://github.com/documenso/documenso/assets/75713174/67b063ca-80a8-4470-9187-8951f9a854a6">

### Do you want to work on this improvement?

Yes

### Please check the boxes that apply to this improvement suggestion.

- [X] I have searched the existing issues and improvement suggestions to avoid duplication.
- [X] I have provided a clear description of the improvement being suggested.
- [X] I have explained the rationale behind this improvement.
- [X] I have included any relevant technical details or design suggestions.
- [X] I understand that this is a suggestion and that there is no guarantee of implementation.

dguyen:
> Fields should have some area exclusively for themself

How are you intending to resolve this? Making it so fields can't be placed on top of each other?

I think we should be updating the layering so that inserted fields are behind uninserted fields

sumitbishti:
I will have to think about it first. Will let you know.

Mythie:
While I can't think of any good reasons to not place fields on top of eachother I can imagine users getting frustrated if they attempt to drop a field and it only just intersects with another one by a few pixels (name + date) which causes the drop to fail.

sumitbishti:
> While I can't think of any good reasons to not place fields on top of eachother

If you do place, how a signer is supposed to fill the fields that are below the top most one? Which do not have any area of click on.

I think we should allow a minimum of 50% percent of a field's area to be intersected by other fields.

ElTimuro:
- Lucas is right; failing drops would be frustrating
- If we do this, it would have to be a "physical limitation" that you can drag them overlapping
- Though, I have never seen this limited before, I assume it not because a lot of docs have tiny space and this would hinder things
- happy to try out a version of the physical block, but no guarantees we merge it

ameeetgaikwad:
Is someone working on this?

ElTimuro:
- Emphasis on "exactly". Preventing the exact coordinated for 2 fields would eliminate this
- Minimal different coordinates are fine

@dguyen do we have a more current view on this? preventing the exact same could prevent a few unwanted cases e.g. placing 2 field exactly the same via api

[...]

PR #3224, opened 2026-08-16
fix: don't create a field on top of an identical one

agenteddyedwards:
`addField` appends without checking, so the same placement can be written twice and you end up with two fields on the same spot for the same recipient.

Three things call it: click to place in `envelope-editor-fields-drag-drop`, drag to draw in `envelope-editor-fields-page-renderer`, and AI detection in `envelope-editor-fields-page`. If two of them run for one action you get the duplicate.

The reason this matters more than it sounds is that the signer cannot get out of it. They fill the field on top, the copy underneath stays uninserted, and the envelope keeps asking for a field they cannot click because the first one is covering it. Next field has nowhere to scroll to and the document never completes.

I know `fields-overlap.ts` already exists and that both the editor and the distribute dialog warn when fields overlap. That warning is the right thing for a sender who placed two fields close together. It does not help here, because nobody placed the second field, and the person who gets stuck is the signer, who never sees the editor or the warning. `invalidEnvelopeCode` in the distribute dialog blocks on missing signatures, missing recipients and a missing required email, but overlap is only a warning, so these envelopes go out.

Data from a self hosted 2.15.0, read back over the API:

```
id=90 DATE page=2 X=66.666667 Y=87.137971 recipientId=33 inserted=False
id=91 DATE page=2 X=66.666667 Y=87.137971 recipientId=33 inserted=False
```

```
SIGNATURE page=4 X=12.406 Y=12.137 recipientId=25 ids=[(78, False), (81, True)]
DATE      page=4 X=17.669 Y=19.046 recipientId=25 ids=[(79, False), (82, False)]
```

Identical to six decimals, so this is not two clicks landing close together, it is the same object written twice. In the second case the recipient had already signed, field 81, and still could not finish the agreement.

The change adds `findDuplicateField` to the existing `fields-overlap` utilities and uses it in `addField`, which now selects the existing field rather than stacking a new one on top. It reuses `getIntersectionArea` so the geometry stays in one place. It only matches fields of the same type, for the same recipient, on the same page of the same envelope item, and only when they cover at least 90 percent of each other, so anything a sender placed close together on purpose is left alone.

There were no tests on `fields-overlap.ts`, so I added them for the new function and for `getOverlappingFieldPairs` and `hasOverlappingFields` while I was there.

You asked on #907 whether sending should be blocked when fields are stacked. I have not done that here since it changes behaviour for existing users, but I am happy to add it if you want it, either in this PR or a follow up.

### DIFF
--- a/packages/lib/client-only/hooks/use-editor-fields.ts
+++ b/packages/lib/client-only/hooks/use-editor-fields.ts
@@ -2,6 +2,7 @@ import { getPdfPagesCount } from '@documenso/lib/constants/pdf-viewer';
 import type { TEditorEnvelope } from '@documenso/lib/types/envelope-editor';
 import { ZFieldMetaSchema } from '@documenso/lib/types/field-meta';
 import { nanoid } from '@documenso/lib/universal/id';
+import { findDuplicateField } from '@documenso/lib/utils/fields-overlap';
 import { zodResolver } from '@hookform/resolvers/zod';
 import type { Field } from '@prisma/client';
 import { FieldType } from '@prisma/client';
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

agenteddyedwards:
Signed under this account, thanks.

PR #3230, opened 2026-08-18
feat: last used column for api tokens

### LINKED ISSUE #2756: feat: track last used timestamp for API tokens
### Feature Description

The ApiToken model currently stores `createdAt` and `expires` but has no 
`lastUsedAt` field. Every time a token is validated in `getApiTokenByToken`, 
the usage is not recorded anywhere.

This means users managing API tokens in the team settings have no way to know:
- Whether a token is actively being used
- When it was last used
- Whether a token is safe to delete (might break an integration) or has been 
  sitting unused for months (could be stale/leaked)

### Use Case

A team admin wants to rotate or clean up old API tokens. Without a "last used" 
timestamp, they cannot distinguish between:
- A token powering an active integration (deleting it would break something)
- A token that was created 6 months ago and never used (safe to delete)

This is standard practice on GitHub, Linear, Stripe, and every major API 
platform — "last used" is the primary signal for token hygiene and security audits.

### Proposed Solution

1. Add a `lastUsedAt DateTime?` field to the `ApiToken` model in schema.prisma
2. In `packages/lib/server-only/public-api/get-api-token-by-token.ts`, after 
   successful validation, update `lastUsedAt` to `new Date()` (fire-and-forget, 
   no need to block the request)
3. Display `lastUsedAt` alongside `createdAt` and `expires` in the token list UI 
   at `apps/remix/app/routes/_authenticated+/t.$teamUrl+/settings.tokens.tsx`

The update to `lastUsedAt` should be a non-blocking background write so it 
doesn't add latency to API requests.

### Alternatives (optional)

Could use the existing `updatedAt` field on ApiToken but that model doesn't 
have one — and conflating "token metadata updated" with "token was used" would 
be semantically wrong anyway.

### Additional Context

_No response_

catalinpit:
Track and display the last time an API token was used.

Screenshot 2026-08-19 at 11 35 45

Based on https://github.com/documenso/documenso/pull/2760

### DIFF
--- a/packages/lib/server-only/public-api/get-api-token-by-token.ts
+++ b/packages/lib/server-only/public-api/get-api-token-by-token.ts
@@ -96,6 +96,17 @@ export const getApiTokenByToken = async ({ token, bypassRateLimit = false }: Get
     });
   }
 
+  void prisma.apiToken
+    .update({
+      where: {
+        id: apiToken.id,
+      },
+      data: {
+        lastUsedAt: new Date(),
+      },
+    })
+    .catch(() => null);
+
   return {
     ...apiToken,
     user,
--- a/packages/lib/server-only/public-api/get-api-token-by-token.ts
+++ b/packages/lib/server-only/public-api/get-api-token-by-token.ts
@@ -5,6 +5,8 @@ import { logger } from '../../utils/logger';
 import { hashString } from '../auth/hash';
 import { assertOrganisationRatesAndLimits } from '../rate-limit/assert-organisation-rates-and-limits';
 
+const LAST_USED_AT_UPDATE_INTERVAL = 60_000; // 1 minute
+
 type GetApiTokenByTokenOptions = {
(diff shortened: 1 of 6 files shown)


packages/lib/server-only/public-api/get-api-token-by-token.ts
```diff
@@ -96,6 +97,23 @@ export const getApiTokenByToken = async ({ token, bypassRateLimit = false }: Get
...
+  void prisma.apiToken
+    .update({
+      where: {
+        id: apiToken.id,
+      },
+      data: {
+        lastUsedAt: new Date(),
+      },
+    })
+    .catch((err) => {
```
catalinpit: don't block the request if there's an issue updating the `lastUsedAt` property, but log the error and continue. `lastUsedAt` is only informative.

dguyen (approved):
Looks good to me

PR #3232, opened 2026-08-18
fix: isolate post-seal webhook and email dispatch

yzxcj797:
## Summary

Addresses #2814.

**Status of the original report on current `main`:** the webhook is already dispatched *before* `send.document.completed.emails` is queued, and email delivery runs in its own background job — so SMTP failures can no longer block `DOCUMENT_COMPLETED` / `DOCUMENT_REJECTED` dispatch (details in the issue comment).

**What remained broken** is the symmetric coupling: both dispatches run unguarded in the job body *after* the seal transaction has committed.

- `triggerWebhook` throws on a webhook-listing failure → the whole seal job fails for an envelope that is **already sealed**, and the completion email is never queued.
- `jobs.triggerJob` rejecting (queue briefly unavailable) → same post-commit job failure, in the other direction.

Either way one post-seal side effect can fail the committed seal and suppress the other.

## Fix

Each dispatch now runs in its own `io.runTask` (gaining the backend's task-level retry semantics and stable cache keys) and is isolated from the other with a logged catch via `io.logger.error`:

```ts
try {
  await io.runTask('seal-document--trigger-webhook', async () => { ... });
} catch (err) {
  io.logger.error('Failed to dispatch completion webhook for sealed envelope', err);
}
```

A failure in one side effect can now neither fail the job for an already-sealed envelope nor block the other — the "independent post-seal side effect" behaviour the issue asks for.

## Notes

- Task cache keys (`seal-document--trigger-webhook`, `seal-document--send-completed-emails`) are unique within the job, so they don't collide with the existing `seal-document` task.
- No behavioral change to *when* webhooks/emails fire or to their payloads — only failure isolation around the two dispatches.

### DIFF
--- a/packages/email/preview/app/lib/templates.tsx
+++ b/packages/email/preview/app/lib/templates.tsx
@@ -20,6 +20,7 @@ import { ForgotPasswordTemplate } from '../../../templates/forgot-password';
 import { OrganisationAccountLinkConfirmationTemplate } from '../../../templates/organisation-account-link-confirmation';
 import { OrganisationDeleteEmailTemplate } from '../../../templates/organisation-delete';
 import { OrganisationInviteEmailTemplate } from '../../../templates/organisation-invite';
+import { OrganisationInviteDeclinedEmailTemplate } from '../../../templates/organisation-invite-declined';
 import { OrganisationJoinEmailTemplate } from '../../../templates/organisation-join';
 import { OrganisationLeaveEmailTemplate } from '../../../templates/organisation-leave';
 import { OrganisationLimitAlertEmailTemplate } from '../../../templates/organisation-limit-alert';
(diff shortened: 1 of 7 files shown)


Mythie:

https://documen.so/cla

Mythie:
AI SLOP

yzxcj797:
Understood — and I hear the frustration about volume; 15 PRs to one repo in a day is a lot, and I'll slow down here.

For what it's worth, the analysis in this PR was done by reading the call sites (the webhook dispatch genuinely runs after the seal transaction commits, and a throw there fails the job for an already-sealed envelope) — but you know this codebase far better than I do, and if the framing or the fix shape missed context that makes this a non-issue, I'm happy to withdraw it entirely. No hard feelings either way.

Issue #2811, opened 2026-05-13
Make internal background job cron schedules configurable


zahedshareef:
## Problem

Several internal background job sweep schedules are hardcoded in the job definitions. For example, recipient expiration, document sealing, signing reminders, and rate-limit cleanup currently run on fixed cron expressions such as `*/15 * * * *`.

For self-hosted installations, operators may need to tune these schedules based on infrastructure constraints, email delivery behavior, or expected workload. Today that requires modifying source code.

## Proposal

Make the internal sweep cron expressions configurable with environment variables while preserving the current defaults.

Suggested defaults:

- `NEXT_PRIVATE_JOBS_SIGNING_REMINDERS_SWEEP_CRON="*/15 * * * *"`
- `NEXT_PRIVATE_JOBS_SEAL_DOCUMENT_SWEEP_CRON="*/15 * * * *"`
- `NEXT_PRIVATE_JOBS_EXPIRE_RECIPIENTS_SWEEP_CRON="*/15 * * * *"`
- `NEXT_PRIVATE_JOBS_CLEANUP_RATE_LIMITS_CRON="*/15 * * * *"`

The implementation should validate configured cron expressions and document the variables for self-hosters.

## Notes

This should not change default behavior for existing deployments.

zahedshareef:
I can take this and will open a small PR that preserves the existing defaults while allowing self-hosted deployments to override these cron schedules.

ElTimuro:
Looking into it

yzxcj797:
Picking this up along the suggested shape:

- a small `resolveSweepCron(envKey)` helper resolves each sweep's schedule from the environment with the `*/15 * * * *` default preserved — wired into all four jobs (`signing-reminders-sweep`, `seal-document-sweep`, `expire-recipients-sweep`, `cleanup-rate-limits`) using the exact variable names from the issue;
- invalid cron expressions **throw at boot** (validated with the already-present `cron-parser` dependency) rather than silently disabling a sweep — a mistyped schedule would otherwise leave expired recipients un-swept or rate-limit rows accumulating with nothing pointing at the cause;
- the four variables are typed in `process-env.d.ts` and documented in `.env.example` under the existing `[[BACKGROUND JOBS]]` section.

Default behavior is unchanged for deployments that set nothing. Tests cover the default fallback (unset and blank), pass-through of valid expressions (with surrounding whitespace trimmed), and boot-time rejection of invalid ones.

Mythie:
[...]

PR #3234, opened 2026-08-18
feat: make internal sweep job cron schedules configurable

yzxcj797:
## Summary

Implements #2811.

The four internal sweeps (`signing-reminders-sweep`, `seal-document-sweep`, `expire-recipients-sweep`, `cleanup-rate-limits`) ran on a hardcoded `*/15 * * * *` cron, so self-hosters tuning schedules for their infrastructure had to patch source.

## What this does

- A small `resolveSweepCron(envKey)` helper resolves each sweep's schedule from the environment, defaulting to the previous 15-minute cadence — **default behavior is unchanged for deployments that set nothing**.
- Wired into all four jobs with the exact variable names from the issue:
  - `NEXT_PRIVATE_JOBS_SIGNING_REMINDERS_SWEEP_CRON`
  - `NEXT_PRIVATE_JOBS_SEAL_DOCUMENT_SWEEP_CRON`
  - `NEXT_PRIVATE_JOBS_EXPIRE_RECIPIENTS_SWEEP_CRON`
  - `NEXT_PRIVATE_JOBS_CLEANUP_RATE_LIMITS_CRON`
- **Invalid cron expressions throw at boot** (validated with the already-present `cron-parser` dependency) rather than silently disabling a sweep — a mistyped schedule would otherwise leave expired recipients un-swept or rate-limit rows accumulating with nothing pointing at the cause.
- The four variables are typed in `packages/tsconfig/process-env.d.ts` and documented in `.env.example` under the existing `[[BACKGROUND JOBS]]` section.

## Tests

`packages/lib/jobs/definitions/internal/sweep-crons.test.ts` (vitest, same setup as the other `packages/lib` unit tests): default fallback for unset/blank values, pass-through of valid expressions (surrounding whitespace trimmed), and boot-time rejection of invalid ones.


Mythie:

https://documen.so/cla

Mythie:
> I don't think we really need this, if the midnight cron needs to be changed you'd just change your servers timezone instead.

Issue #3204, opened 2026-08-13
Webhook payloads carry no delivery id, so documented retries cannot be deduplicated


bastien70:
### Issue Description

Webhook deliveries carry no stable identifier, so a consumer cannot deduplicate retries.

**`packages/lib/jobs/definitions/internal/execute-webhook.handler.ts:20-25`**
```js
const payloadData = { event, payload: data, createdAt: new Date().toISOString(), webhookEndpoint: url };
```
**`packages/lib/server-only/webhooks/execute-webhook-call.ts:38-41`** — headers are `Content-Type` and the secret only. No delivery id in the body, none in a header.

This matters because retries are real and intentional: `bullmq.ts:117,158` schedules with `attempts: DEFAULT_MAX_RETRIES`, and the handler rethrows explicitly (`:44`) when a delivery fails. And `createdAt` is recomputed with `new Date()` **inside `run`**, so two attempts at the same event reach the consumer with **different bodies** and no stable field between them — not even a timestamp to hash on.

The server does mint an identifier: the `WebhookCall` row created at `:27-40`. It is simply never sent.

### Steps to Reproduce

1. Point a webhook at an endpoint that returns 500 on first call and 200 afterwards.
2. Trigger any event.
3. Compare the two received bodies.

### Expected Behavior

Both attempts carry the same delivery id, so the consumer can treat the second as a duplicate.

### Current Behavior

Nothing is shared between the two bodies. `createdAt` differs. A consumer that acts on each delivery — issuing an invoice, filing a document — has no way to avoid doing it twice.

### Version

`main` @ 617f8cc (v2.16.0)

### Notes

Found during a code audit; verified by reading the handler and the retry configuration on current `main`. Not reproduced against a live endpoint.

I recognise this is closer to an enhancement than a defect, and that #3132 recently corrected the retry documentation — the gap is precisely that the retries are now documented while no idempotency key accompanies the payload.

Suggested direction: generate a delivery id when the job payload is built (so it stays constant across attempts) and expose it both in the body and as a header.

yzxcj797:
Implemented along your suggested direction:

- `triggerWebhook` mints a `deliveryId` (nanoid) **when the job payload is built**, so it is constant across the queue's retry attempts; the job schema field is optional (jobs queued before the change still validate) and the handler falls back to generating one for those.
[...]

PR #3237, opened 2026-08-18
feat(webhooks): give deliveries a stable id for retry deduplication

yzxcj797:
## Summary

Implements #3204.

## The gap

Webhook retries are real and intentional (`bullmq.ts` schedules `DEFAULT_MAX_ATTEMPTS`, the handler rethrows on failure), but deliveries carried **no stable identifier**: `createdAt` is recomputed with `new Date()` inside each attempt, so two attempts of the same delivery reached the consumer with different bodies and no shared field — a consumer that acts per delivery (issuing an invoice, filing a document) had no way to avoid acting twice. The server mints a `WebhookCall` row per attempt, but that id is never sent and changes between attempts anyway.

## What this does (the issue's suggested direction)

- `triggerWebhook` mints a **`deliveryId`** (nanoid) when the job payload is built — constant across retry attempts by construction.
- The job schema gains an **optional** `deliveryId` (jobs queued before the change still validate); the handler falls back to generating one for those.
- The handler includes it in the body as `deliveryId`, and `executeWebhookCall` sends it as **`X-Documenso-Delivery-Id`** — matching the existing `X-Documenso-Secret` / `X-Documenso-Envelope-Id` header family — so consumers can deduplicate from either surface without parsing the body.
- `createdAt` still reflects each attempt; `deliveryId` is the stable key.
- The `WebhookCall` row (persisted after the call) records the sent body and therefore the delivery id per attempt, keeping the audit trail aligned with what the consumer received.

## Notes

The webhook retry documentation updated in #3132 can now reference the header for its deduplication guidance. No database migration needed.

### DIFF
--- a/packages/lib/jobs/definitions/internal/execute-webhook.handler.ts
+++ b/packages/lib/jobs/definitions/internal/execute-webhook.handler.ts
@@ -3,6 +3,8 @@ import { prisma } from '@documenso/prisma';
 import type { Prisma } from '@prisma/client';
 import { WebhookCallStatus } from '@prisma/client';
 
+import { nanoid } from 'nanoid';
+
 import type { JobRunIO } from '../../client/_internal/job';
 import type { TExecuteWebhookJobDefinition } from './execute-webhook';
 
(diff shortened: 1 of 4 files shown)


Mythie:

https://documen.so/cla

PR #3239, opened 2026-08-18
fix: keep document filters across pagination; reset page on search

### LINKED ISSUE #3196: Paginating the documents list silently drops every active filter (nuqs shallow vs React Router)
### Issue Description

The documents list keeps its URL state in **two competing places**, and the pagination side silently discards what the filter side wrote.

**Filters write through nuqs, shallow:**
`documents-table-status-filter.tsx:25-31`, `documents-table-sender-filter.tsx:21-27`, `documents-table-period-filter.tsx:15-21` all call `useQueryStates(..., { history: 'push' })` with no `shallow` option.

- `nuqs@2.8.9` `dist/index.js:454` — `shallow = defaultOptions?.shallow ?? true`
- `nuqs` `dist/react-router-*.js:15-34` —
  ```js
  updateMethod.call(history, history.state, historyUpdateMarker, url);
  if (options.shallow === false) navigate({ hash: url.hash, search: url.search }, {...});
  ```
  In shallow mode `navigate()` is **never called**, so React Router is never told.
- `react-router@7.18.1` — `history.listen` subscribes to **`popstate` only**, and `pushState` does not emit `popstate`. `useSearchParams` derives from `useLocation()`, i.e. from router state, which is now **stale**.

**Pagination reads that stale state:** `documents-table.tsx:54,130-137` → `use-update-search-params.ts:6-17`
```js
const nextSearchParams = new URLSearchParams(searchParams?.toString() ?? '');
```
so it rebuilds the URL from a snapshot that predates the filter, and the filter keys vanish.

The asymmetry is what makes this one-directional: nuqs recomposes from the **live** URL, so it preserves what React Router wrote — but not the reverse.

**Same page, same root cause:** the search box has a related but distinct bug — `document-search.tsx:13,18-22` writes only `query` and never resets `page`, while its three sibling filters all reset `page: null` (status `:51-55`, sender `:41-45`, period `:24-28`). Searching from page 3 therefore requests page 3 of a smaller result set and shows an empty table (`find-documents.ts:510` does not clamp the offset). I am filing them together since a fix for the URL-state layer should cover both.

### Steps to Reproduce

**Filters:** open `/t/<team>/documents` → set Status = Completed (URL shows `?status=COMPLETED`) → click next page → URL becomes `?page=2&perPage=10`, `status` is gone, the table is unfiltered.

**Search:** go to page 2 → type a search matching few results → empty table.

### Expected Behavior

Paginating keeps the active filters; searching returns to page 1.

### Current Behavior

Paginating clears every filter and the pills reset. Searching from a later page shows "no documents" although matches exist.

### Version

`main` @ 617f8cc (v2.16.0)

### Notes

Found during a code audit; the mechanism was verified by reading the sources of both dependencies in `node_modules`, not inferred. Not reproduced in a browser.

The same mix exists on `template-page-view-documents-table.tsx`, where `<DocumentSearch />` (nuqs) sits beside `useUpdateSearchParams` + `SearchParamSelector` (React Router): picking a status there clears the search.

Suggested direction: pick one source of truth — either `shallow: false` on the filters' `useQueryStates`, or route pagination through nuqs (`setFindDocumentSearchParams({ page, perPage })`) instead of `useUpdateSearchParams`.

yzxcj797:
## Summary

Fixes #3196.

## Root cause (as diagnosed in the issue, fixed at the URL-state layer)

The documents list kept URL state in **two competing places**:

- filters wrote through **nuqs** — shallow by default, i.e. `pushState`, which never notifies React Router (`history.listen` subscribes to `popstate`, and `pushState` doesn't fire it), so `useSearchParams`/`useLocation` go stale;
- pagination rebuilt the next URL from that stale React Router snapshot via `useUpdateSearchParams`, silently discarding whatever the filters had just written. The asymmetry made it one-directional (nuqs recomposes from the live URL, so router writes survived; router recomposition from nuqs writes did not).

## Fix

1. **`useUpdateSearchParams` rebuilds from `window.location.search`** — the live URL that **both** writers keep current (nuqs `pushState` and router `navigate` both update it). `setSearchParams` still routes through React Router, refreshing its state exactly as before. This fixes every consumer of the hook (33 files), not just the documents table.
2. **`DocumentSearch` resets `page: null` on a new query**, matching its three sibling filters (status/sender/period all do) — searching from page 3 no longer requests page 3 of a smaller result set.

No changes to the nuqs write side: filters keep their shallow updates and push history, so back/forward and per-keystroke performance are untouched.

### DIFF
--- a/apps/remix/app/components/general/document/document-search.tsx
+++ b/apps/remix/app/components/general/document/document-search.tsx
@@ -2,7 +2,7 @@ import { useDebouncedValue } from '@documenso/lib/client-only/hooks/use-debounce
 import { Input } from '@documenso/ui/primitives/input';
 import { msg } from '@lingui/core/macro';
 import { useLingui } from '@lingui/react';
-import { useQueryState } from 'nuqs';
+import { useQueryStates } from 'nuqs';
 import { useEffect, useState } from 'react';
 
 import { documentsSearchParams } from '~/utils/documents-search-params';
@@ -10,14 +10,23 @@ import { documentsSearchParams } from '~/utils/documents-search-params';
 export const DocumentSearch = () => {
   const { _ } = useLingui();
 
-  const [query, setQuery] = useQueryState('query', documentsSearchParams.query);
+  const [query, setQuery] = useQueryStates(
+    {
+      query: documentsSearchParams.query,
+      page: documentsSearchParams.page,
+    },
+    { history: 'push' },
+  );
 
   const [searchTerm, setSearchTerm] = useState(query ?? '');
   const debouncedSearchTerm = useDebouncedValue(searchTerm, 500);
 
   useEffect(() => {
     if (debouncedSearchTerm !== (query ?? '')) {
-      void setQuery(debouncedSearchTerm || null);
+      // A new query shrinks the result set; staying on page 3 of a smaller
+      // set renders an empty table. Reset the page like every sibling
+      // filter does (#3196).
+      void setQuery({ query: debouncedSearchTerm || null, page: null });
     }
   }, [debouncedSearchTerm, query, setQuery]);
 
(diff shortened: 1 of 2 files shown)


Mythie:

https://documen.so/cla

PR #3241, opened 2026-08-18
fix: stop writing EMAIL_SENT audit entries when no email was sent

### LINKED ISSUE #3194: EMAIL_SENT audit entries are written for recipients who were never emailed, and reach the signing certificate
### Issue Description

In `send-signing-email.handler.ts`, the send is guarded but the audit write is not — the `if` closes before it.

```ts
if (isRecipientEmailValidForSending(recipient)) {          // 188
  try { await assertOrganisationRatesAndLimits({...}); }
  catch (_err) { io.logger.warn({...}); return; }          // 196-206
  await io.runTask('send-signing-email', async () => {
    await emailTransport.sendMail({...});                  // 218
  });
}                                                          // 235 — the if ENDS here

const sentAt = new Date();
await io.runTask('update-recipient', async () => {
  await prisma.recipient.update({ ..., data: { sendStatus: SendStatus.SENT, sentAt } });  // 245-246
});
await prisma.documentAuditLog.create({
  data: createDocumentAuditLogData({ type: DOCUMENT_AUDIT_LOG_TYPE.EMAIL_SENT, ... }),    // 261 — unconditional
});
```

When the guard is false, no email is sent — and yet `sendStatus` becomes `SENT`, `sentAt` is stamped, and an `EMAIL_SENT` audit entry is written.

**This is reachable through a supported configuration, not an edge case.** `isRecipientEmailValidForSending` is `zEmail().safeParse(email).success` (`utils/recipients.ts:163-165`), and the empty email is an **explicitly sanctioned branch of the schema**: `ZRecipientEmailSchema = z.union([z.literal(''), zEmail(...)])` (`types/recipient.ts:127`) — the in-person signer case. `send-document.ts:172-187` only blocks an empty email when the recipient has auth requirements, and `send-document.ts:342-359` then triggers the job **with no validity filter at all** (it only skips `sendStatus === SENT` and CC).

**Why it matters beyond the table:** `EMAIL_SENT` feeds the signing certificate (`generate-certificate-pdf.ts:67-69`, `certificate.tsx:174-175, 297-298`). The evidentiary artefact attached to the signed document states that an email was sent to a party who was never contacted.

The four other `EMAIL_SENT` write sites are all correct (`send-document-completed-emails.handler.ts` writes after the send and filters upstream; `resend-document.ts:202-204` returns first; `process-signing-reminder.handler.ts` nests it after a successful send) — so this is an isolated deviation, not a convention.

### Steps to Reproduce

1. Create a document with a recipient whose email is empty (the in-person signer flow the schema allows).
2. Distribute the document.
3. Look at that recipient's `sendStatus` / `sentAt`, the audit log, and the generated certificate.

### Expected Behavior

No send means no `sentAt`, no `SENT`, and no `EMAIL_SENT` entry.

### Current Behavior

All three are written. The sender sees "Sent" with a timestamp, and the certificate attests to a delivery that never happened.

### Version

`main` @ 617f8cc (v2.16.0)

### Notes

Found during a code audit; verified by reading the block boundaries on current `main`. Not reproduced against a running instance — but the path is covered by an existing green E2E test, *"Distribute envelope with empty email recipients"* (`packages/app-tests/e2e/api/v2/envelopes-api.spec.ts:1049`), which simply never inspects the audit log.

Suggested direction: move the recipient update and the audit write inside the `if`.

yzxcj797:
## Summary

Fixes #3194.

In `send-signing-email.handler.ts` the send was guarded by `isRecipientEmailValidForSending`, but the guard's `if` closed **before** the state writes — so when no email was sent:

- `sendStatus` was still set to `SENT`,
- `sentAt` was stamped,
- the next reminder was scheduled, and
- an `EMAIL_SENT` audit entry was written.

This is reachable through a **supported configuration**: the recipient schema explicitly sanctions an empty email (the in-person signer), and `send-document.ts` triggers the job with no validity filter. Beyond the misleading table state, the `EMAIL_SENT` entry feeds the **signing certificate** (`generate-certificate-pdf.ts`, `certificate.tsx`) — the evidentiary artefact stated that an email was sent to a party who was never contacted.

## Fix

The guard is inverted into an **early return**: no valid email means no send, no `sendStatus`/`sentAt` write, no reminder scheduling, and no audit entry. The send-side body is de-nested; behavior for valid emails is byte-for-byte the same, including the rate-limit consume-and-return branch.

This aligns the handler with the convention of the four other `EMAIL_SENT` write sites (all of which write only after a successful send, as the issue verified).

### DIFF
--- a/packages/lib/jobs/definitions/emails/send-signing-email.handler.ts
+++ b/packages/lib/jobs/definitions/emails/send-signing-email.handler.ts
@@ -185,54 +185,60 @@ export const run = async ({ payload, io }: { payload: TSendSigningEmailJobDefini
     reportUrl,
   });
 
-  if (isRecipientEmailValidForSending(recipient)) {
-    try {
-      await assertOrganisationRatesAndLimits({
-        organisationId,
-        organisationClaim: claims,
-        type: 'email',
-        count: 1,
-      });
-    } catch (_err) {
-      io.logger.warn({
-        msg: 'Recipient signing email dropped: org rate limit exceeded',
-        organisationId,
-        recipientId: recipient.id,
-        envelopeId: envelope.id,
-      });
-
-      // Job is consumed and NOT retried.
-      return;
-    }
-
-    await io.runTask('send-signing-email', async () => {
-      const [html, text] = await Promise.all([
-        renderEmailWithI18N(template, { lang: emailLanguage, branding }),
-        renderEmailWithI18N(template, {
-          lang: emailLanguage,
-          branding,
-          plainText: true,
-        }),
-      ]);
-
-      await emailTransport.sendMail({
-        to: {
-          name: recipient.name,
-          address: recipient.email,
-        },
-        from: senderEmail,
-        replyTo: replyToEmail,
-        subject: renderCustomEmailTemplate(documentMeta?.subject || emailSubject, customEmailTemplate),
-        html,
-        text,
-        headers: buildEnvelopeEmailHeaders({
-          userId,
-          envelopeId: envelope.id,
-          teamId: envelope.teamId,
-        }),
-      });
-    });
+  if (!isRecipientEmailValidForSending(recipient)) {
+    // An explicitly empty email (the in-person signer the recipient schema
+    // sanctions) means no email is sent: no sendStatus=SENT, no sentAt, and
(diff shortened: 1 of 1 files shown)


Mythie:

https://documen.so/cla

PR #3244, opened 2026-08-18
fix: make the template document filters clearable

### LINKED ISSUE #3200: Template document filters always display "Any" and cannot be cleared
### Issue Description

**`apps/remix/app/components/general/template/template-page-view-documents-table.tsx:178` and `:196`**
```jsx
isValueValid={(value) => [...DocumentStatusEnum.COMPLETED].includes(value as unknown as string)}
isValueValid={(value) => [...DocumentSource.TEMPLATE].includes(value as unknown as string)}
```

`DocumentStatusEnum.COMPLETED` is the **string** `'COMPLETED'`, so `[...'COMPLETED']` spreads to `['C','O','M','P','L','E','T','E','D']`. `.includes('COMPLETED')` is therefore **always false**, for both selectors.

Consequences in `search-param-selector.tsx`:
- `:18-22` — `value` always falls back to `'all'`
- `:41` — `<Select defaultValue={value}>` is **uncontrolled**, so `defaultValue` is only read on first render

The data *is* filtered server-side, so the table and the menu disagree. Worse, the filter cannot be cleared: `@radix-ui/react-use-controllable-state@1.2.2` only fires `onChange` when the value actually changes —

```js
React.useEffect(() => {
  if (prevValueRef.current !== value) { onChangeRef.current?.(value); prevValueRef.current = value; }
}, [value, prevValueRef]);
```

Since the internal value is already `'all'`, picking "Any Status" emits nothing.

### Steps to Reproduce

1. Open a template's page (`/t/<team>/templates/<id>`).
2. Set Status = Completed. The URL gets `?status=COMPLETED` and the table filters correctly.
3. Reload the page (or open the shared link).

### Expected Behavior

The menu shows "Completed", and selecting "Any Status" clears the filter.

### Current Behavior

The menu shows **"Any Status"** while the table stays filtered, and clicking "Any Status" does nothing — the filter can no longer be removed from the UI.

### Version

`main` @ 617f8cc (v2.16.0)

### Notes

Found during a code audit; the non-firing `onValueChange` was verified by reading the Radix source rather than assumed. Not reproduced in a browser.

Suggested direction: use a real list — `Object.values(DocumentStatusEnum).includes(...)` and `Object.values(DocumentSource).includes(...)` — **and** make the `Select` controlled (`value=` instead of `defaultValue=`), otherwise the display/URL mismatch survives any remount.

Not covered by #3148 (*"replace template view tabs with a filter pill"*), which touches the templates **index** page.

yzxcj797:
## Summary

Fixes #3200.

## The bug (as diagnosed in the issue)

Both `isValueValid` validators spread the enum's **string value** into an array literal:

```ts
[...DocumentStatusEnum.COMPLETED]  // → ['C','O','M','P','L','E','T','E','D']
[...DocumentSource.TEMPLATE]       // → ['T','E','M','P','L','A','T','E']
```

so `.includes('COMPLETED')` was always false. Consequences: the selector always fell back to displaying "Any" while the table stayed server-side filtered, and — because the underlying Radix Select is uncontrolled with its internal value stuck at `'all'` and `onChange` only fires on real changes — **picking "Any Status" emitted nothing**: the filter could not be cleared from the UI.

## Fix

The validators now list the actual values the menus offer:

- status: `[COMPLETED, PENDING, DRAFT]`
- source: `[TEMPLATE]`

A URL-carried filter now displays correctly on load, and "Any …" clears it.

### DIFF
--- a/apps/remix/app/components/general/template/template-page-view-documents-table.tsx
+++ b/apps/remix/app/components/general/template/template-page-view-documents-table.tsx
@@ -175,7 +175,11 @@ export const TemplatePageViewDocumentsTable = ({ templateId }: TemplatePageViewD
 
         <SearchParamSelector
           paramKey="status"
-          isValueValid={(value) => [...DocumentStatusEnum.COMPLETED].includes(value as unknown as string)}
+          isValueValid={(value) =>
+            [DocumentStatusEnum.COMPLETED, DocumentStatusEnum.PENDING, DocumentStatusEnum.DRAFT].includes(
+              value as unknown as DocumentStatusEnum,
+            )
+          }
         >
           <SelectItem value="all">
             <Trans>Any Status</Trans>
@@ -193,7 +197,7 @@ export const TemplatePageViewDocumentsTable = ({ templateId }: TemplatePageViewD
 
         <SearchParamSelector
           paramKey="source"
-          isValueValid={(value) => [...DocumentSource.TEMPLATE].includes(value as unknown as string)}
+          isValueValid={(value) => [DocumentSource.TEMPLATE].includes(value as unknown as DocumentSource)}
         >
           <SelectItem value="all">
             <Trans>Any Source</Trans>


Mythie:

https://documen.so/cla

PR #3245, opened 2026-08-18
fix: discard an abandoned signature drawing when the pad closes

### LINKED ISSUE #3203: Signature pad keeps a discarded drawing in local state and commits it on the next confirm
### Issue Description

`signature-pad-dialog.tsx` keeps the drawing in local state that is never reset when the dialog closes, so what is displayed and what gets committed can diverge.

- `:41-42` — `const [signature, setSignature] = useState<string>(value ?? '')`, declared **outside** the `<Dialog>` and never re-synced to `value`
- `packages/ui/primitives/dialog.tsx` does not pass `forceMount`, and `@radix-ui/react-dialog` renders `<Presence present={forceMount || context.open}>` → the **content is unmounted** on close
- On reopen, `<SignaturePad value={value}>` (`:117`) remounts from the **parent's** value (`signature-pad.tsx:45-47` initialises from `value` and only calls `onChange` on interaction), while the local `signature` still holds the abandoned drawing
- The confirm button is `disabled={!signature}` → **enabled**, and its `onClick` (`:133-139`) commits `signature`, not what is on screen

To be precise about what is *not* the bug: **Cancel alone commits nothing** — `:127-131` is a bare `<DialogClose>`, and `onChange` only fires from the confirm button. The defect is the stale local state, which surfaces on the *next* interaction.

### Steps to Reproduce

1. Open the signature pad, draw a signature, click **Cancel**.
2. Reopen the pad — the canvas shows the parent's value (the previous signature, or nothing).
3. Click the confirm button **without drawing again**.

### Expected Behavior

Confirming commits what is displayed. After Cancel, the abandoned drawing is gone.

### Current Behavior

The abandoned drawing is committed. In the variant where there was no previous signature, the confirm button is already enabled over a visually empty pad.

### Version

`main` @ 617f8cc (v2.16.0)

### Notes

Found during a code audit; the unmount-on-close behaviour was verified in the Radix source rather than assumed. Not reproduced in a browser.

On a signing product this means a signature the user believes they discarded can end up on the document, which is why I am filing it despite the narrow reproduction path.

Suggested direction: reset in the dialog's `onOpenChange` — `setSignature(value ?? '')` on close — or key the content so local state is rebuilt.

yzxcj797:
## Summary

Fixes #3203.

The signature pad dialog kept its drawing in local state declared **outside** the `<Dialog>` and never re-synced to `value`. Since the Radix dialog content **unmounts on close**, the reopened pad remounted from the parent's value — while the stale draft lingered in local state:

- the confirm button (`disabled={!signature}`) was **enabled** over whatever was actually displayed (including a visually empty pad), and
- confirming committed the **previously discarded drawing**, not the canvas contents.

On a signing product that means a signature the user believes they discarded can end up on the document.

## Fix (the issue's first suggested direction)

`onOpenChange` now resets the draft when the dialog closes:

```ts
if (!open) {
  setSignature(value ?? '');
}
```

Displayed state and committable state can no longer diverge — after Cancel, the abandoned drawing is gone, exactly as the pad's remounted canvas already showed.

### DIFF
--- a/packages/ui/primitives/signature-pad/signature-pad-dialog.tsx
+++ b/packages/ui/primitives/signature-pad/signature-pad-dialog.tsx
@@ -109,7 +109,22 @@ export const SignaturePadDialog = ({
         )}
       </motion.button>
 
-      <Dialog open={showSignatureModal} onOpenChange={disabled ? undefined : setShowSignatureModal}>
+      <Dialog
+        open={showSignatureModal}
+        onOpenChange={(open) => {
+          if (disabled) return;
+
+          // Reset the local draft when the dialog closes: the Radix content
+          // unmounts on close, so the reopened pad remounts from the parent's
+          // value -- an abandoned drawing kept here would diverge from what is
+          // displayed and be committed by the next confirm (#3203).
+          if (!open) {
+            setSignature(value ?? '');
+          }
+
+          setShowSignatureModal(open);
+        }}
+      >
         <DialogContent hideClose={true} className="p-6 pt-4">
           <SignaturePad
             id="signature"


Mythie:

https://documen.so/cla

PR #3246, opened 2026-08-18
fix: re-seed the send dialog from current meta on every open

### LINKED ISSUE #3197: Send dialog keeps the subject/message captured at first mount, and persists them on send
### Issue Description

`envelope-distribute-dialog.tsx:80-91` seeds react-hook-form with `defaultValues`, which RHF reads **only on first render**:

```js
const form = useForm({ defaultValues: { meta: {
  emailId: envelope.documentMeta?.emailId ?? null,
  subject: envelope.documentMeta?.subject ?? '',
  message: envelope.documentMeta?.message ?? '', ... } } });
```

The open-effect at 230-240 resynchronises **only** `distributionMethod`:
```js
useEffect(() => {
  if (isOpen && envelope.documentMeta) setValue('meta.distributionMethod', envelope.documentMeta.distributionMethod);
}, [isOpen]);
```

And the component stays mounted for the whole editor session — `envelope-editor-header.tsx:199-209` and `envelope-editor.tsx:349` render it with no `key` and no conditional unmount. So whatever the subject and message were when the editor first rendered is what the dialog carries forever.

**The stale values are not merely displayed — they are written back.** `onFormSubmit` (179-181) posts `meta` to `envelope.distribute`, and `distribute-envelope.ts:26-45` calls `updateDocumentMeta({ subject: meta.subject, message: meta.message, ... })`.

The correct pattern already exists three files away: `envelope-editor-settings-dialog.tsx:320-323` does `form.reset(createDefaultValues())` on `[open]`.

### Steps to Reproduce

1. Open a draft in the envelope editor.
2. Open **Settings**, change the email subject and message, save.
3. Click **Send Document**.

### Expected Behavior

The dialog shows the subject and message just saved, and sends with them.

### Current Behavior

The dialog shows the **old** subject and message, sends the invitation with them, **and overwrites the saved `documentMeta` with the stale values**. The email has already gone out by the time anyone notices.

### Version

`main` @ 617f8cc (v2.16.0)

### Notes

Found during a code audit; verified by reading the dialog, its mount site and the server-side write on current `main`. Not reproduced in a browser.

Suggested direction: `form.reset({...})` in the `[isOpen]` effect instead of the single `setValue`.

yzxcj797:
## Summary

Fixes #3197.

The send dialog stays mounted for the whole editor session (rendered unconditionally, no `key`), and react-hook-form reads `defaultValues` **only on first render**. The open-effect resynchronized exactly one field (`distributionMethod`), so subject and message edited in Settings after the editor first mounted were shown stale — and since `onFormSubmit` posts `meta` to `envelope.distribute`, which writes `subject`/`message` back into `documentMeta`, the send **overwrote the saved values with the first-render snapshot** on an email that had already gone out.

## Fix

The pattern already used three files away in `envelope-editor-settings-dialog`:

- defaults are derived by a `createDefaultValues()` function reading the **current** `envelope.documentMeta`,
- and the open-effect calls `form.reset(createDefaultValues())` on **every open**, replacing the distributionMethod-only resync.

The mid-save `handleSync()` behavior is unchanged. A mid-autosave open still resyncs; an open after Settings changes now shows — and sends — what was saved.

### DIFF
--- a/apps/remix/app/components/dialogs/envelope-distribute-dialog.tsx
+++ b/apps/remix/app/components/dialogs/envelope-distribute-dialog.tsx
@@ -77,16 +77,22 @@ export const EnvelopeDistributeDialog = ({
 
   const { mutateAsync: distributeEnvelope } = trpcReact.envelope.distribute.useMutation();
 
-  const form = useForm<TEnvelopeDistributeFormSchema>({
-    defaultValues: {
-      meta: {
-        emailId: envelope.documentMeta?.emailId ?? null,
-        emailReplyTo: envelope.documentMeta?.emailReplyTo || undefined,
-        subject: envelope.documentMeta?.subject ?? '',
-        message: envelope.documentMeta?.message ?? '',
-        distributionMethod: envelope.documentMeta?.distributionMethod || DocumentDistributionMethod.EMAIL,
-      },
+  // Derived per open, not per mount: the dialog stays mounted for the whole
+  // editor session, so defaultValues alone would freeze whatever the subject
+  // and message were on first render -- and the send would write those stale
+  // values back into documentMeta (#3197).
+  const createDefaultValues = () => ({
+    meta: {
+      emailId: envelope.documentMeta?.emailId ?? null,
+      emailReplyTo: envelope.documentMeta?.emailReplyTo || undefined,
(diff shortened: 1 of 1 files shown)


Mythie:

https://documen.so/cla

PR #3247, opened 2026-08-18
fix: internal jobs resolve team settings and webhooks system-side

### LINKED ISSUE #3192: Internal jobs resolve team settings through the author's membership: revoking group access leaves documents unsealed forever
### Issue Description

Background jobs resolve team settings and webhook subscriptions **through the document author's current team membership**, so a document can become unsealed forever after an access change that has nothing to do with it.

- `seal-document.handler.ts:79-82` — `getTeamSettings({ userId: envelope.userId, teamId })`
- `get-team-settings.ts:16` — scopes through `buildTeamWhereQuery` whenever `userId` is provided, and **throws** `AppError(NOT_FOUND, 'Team not found')` at 26-30
- `getAllWebhooksByEventTrigger` — `userId` is **required**; used by `seal-document.handler.ts:338`, `process-signing-reminder.handler.ts:242`, `process-recipient-expired.handler.ts:71`

When the author no longer resolves as a team member, the sealing job throws, exhausts its retries, and **the document stays `PENDING` although every recipient has signed**. No sealed PDF, no `DOCUMENT_COMPLETED` webhook, no completion email. On the webhook side the failure is silent instead: `triggerWebhook` simply resolves zero subscriptions and returns.

**The member-removal path is already fixed** — `8f68393` (#3046) re-parents envelopes to the organisation owner on removal, and `delete-team-member.ts:120-133` even carries a comment describing this exact failure mode. The residual path is **group-based access loss, which re-parents nothing**:

- `team-router/delete-team-group.ts:91` — no `envelope.updateMany` anywhere in the file
- `organisation-router/delete-organisation-group.ts:79`
- `organisation-router/update-organisation-group.ts:117-124`

### Steps to Reproduce

With default settings (`inheritMembers` is `true` on team creation, so members get access through a group rather than an explicit membership):

1. Alice, an organisation member who was never added to the team explicitly, creates and sends a document.
2. An admin revokes that team's group access (`trpc.team.group.delete`, the "Disable access" action).
3. The recipients finish signing.

### Expected Behavior

Sealing completes. Internal jobs act on behalf of the system, not on behalf of whoever happened to create the envelope.

### Current Behavior

`seal-document` throws `Team not found`, retries are exhausted, and the document is never sealed — everyone has signed, and nothing further happens. `DOCUMENT_COMPLETED` never fires.

### Version

`main` @ 617f8cc (v2.16.0)

### Notes

Found during a code audit; verified by reading both resolution paths, and by confirming that `delete-team-group.ts` contains no re-parenting while `delete-team-member.ts` does. Not reproduced against a running instance.

Suggested direction: internal jobs should resolve without a user — `getTeamSettings({ teamId })` already supports the unscoped form and documents it (« You must provide userId if you want to validate whether the user can access the team settings »). The webhook lookup has **no** unscoped variant today, so one would need adding. Re-parenting envelopes on the group paths, as the member paths already do, would close the remaining gap.

yzxcj797:
## Summary

Fixes #3192 (the job-layer half; see scope note below).

Internal jobs resolved team settings and webhook subscriptions **through the document author's current team membership**. When the author's access came from a group and that group was revoked — a path that re-parents nothing, unlike member removal — `getTeamSettings({ userId, teamId })` threw `NOT_FOUND`, the seal job exhausted its retries, and a **fully signed document stayed unsealed forever**: no sealed PDF, no `DOCUMENT_COMPLETED` webhook, no completion email. On the webhook side the failure was silent (zero subscriptions resolved).

## Fix — system-side resolution for internal jobs

- `seal-document.handler` uses the documented unscoped form: `getTeamSettings({ teamId })`.
- `getAllWebhooksByEventTrigger` and `triggerWebhook` gain an **optional** `userId`:
  - request contexts keep passing it — behavior unchanged, still scoped through the acting user;
  - the three internal handlers that acted as the document author (`seal-document`, `process-signing-reminder`, `process-recipient-expired`) now omit it, resolving the team's webhooks by team alone so they keep firing after an access change.

## Scope note

Re-parenting envelopes on the group-deletion paths (`team.group.delete`, `organisation group delete/update`) — as the member-removal path already does — is deliberately **not** included: it's a data-shape change across three routers that deserves its own review, and the job-layer fix already removes the dependency on the author's membership.

### DIFF
--- a/packages/lib/jobs/definitions/internal/process-recipient-expired.handler.ts
+++ b/packages/lib/jobs/definitions/internal/process-recipient-expired.handler.ts
@@ -68,7 +68,6 @@ export const run = async ({ payload, io }: { payload: TProcessRecipientExpiredJo
   await triggerWebhook({
     event: WebhookTriggerEvents.RECIPIENT_EXPIRED,
     data: ZWebhookDocumentSchema.parse(mapEnvelopeToWebhookDocumentPayload(envelope)),
-    userId: envelope.userId,
     teamId: envelope.teamId,
   });
 
--- a/packages/lib/jobs/definitions/internal/process-signing-reminder.handler.ts
+++ b/packages/lib/jobs/definitions/internal/process-signing-reminder.handler.ts
@@ -239,7 +239,6 @@ export const run = async ({ payload, io }: { payload: TProcessSigningReminderJob
     await triggerWebhook({
       event: WebhookTriggerEvents.DOCUMENT_REMINDER_SENT,
       data: ZWebhookDocumentSchema.parse(mapEnvelopeToWebhookDocumentPayload(envelope)),
-      userId: envelope.userId,
       teamId: envelope.teamId,
     });
   }
--- a/packages/lib/jobs/definitions/internal/seal-document.handler.ts
+++ b/packages/lib/jobs/definitions/internal/seal-document.handler.ts
@@ -76,8 +76,11 @@ export const run = async ({ payload, io }: { payload: TSealDocumentJobDefinition
       throw new Error('At least one envelope item required');
     }
 
+    // System-side resolution: scoping through the author's current membership
+    // makes sealing throw NOT_FOUND the moment their team access is revoked
+    // (e.g. a group removal, which re-parents nothing), leaving a fully
+    // signed document unsealed forever (#3192).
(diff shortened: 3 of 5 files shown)


Mythie:

https://documen.so/cla

PR #3249, opened 2026-08-18
fix: contain signing fields' cqw units to the field box

### LINKED ISSUE #2669: Pre-filled readOnly TEXT field text overflows field border in signing preview
## Description

When creating a TEXT field via the API with `fieldMeta.readOnly: true` and `fieldMeta.text` set to a pre-filled value, the text overflows past the right edge of the field border in the signing preview page.

The final signed/sealed PDF renders the text correctly within the field boundaries (Konva renderer respects field width). The issue is only in the browser signing preview.

## Root Cause Analysis

The signing preview renders inserted/pre-filled text fields via `DocumentSigningFieldsInserted` in `apps/remix/app/components/general/document-signing/document-signing-fields.tsx`:

```tsx
<p className="text-foreground w-full whitespace-pre-wrap text-[clamp(0.425rem,25cqw,0.825rem)]">
  {children}
</p>
```

The font size uses `25cqw` (container query width units). If `container-type: inline-size` is not set on the `FieldRootContainer` or its immediate parent, `cqw` resolves to the viewport or a larger ancestor container's width rather than the field box width. This causes the font to render at a size proportional to the page/viewport, not the field, resulting in text that wraps at a width wider than the field border.

This explains why changing the field `width` percentage (e.g., from 82% to 60%) does not change where text wraps — the font size is calculated from a parent container, not the field element.

## Expected Behavior

Text in pre-filled readOnly TEXT fields should wrap within the field border boundary, matching how the Konva renderer (used for the final PDF) handles it.

## Suggested Fix

Ensure `container-type: inline-size` is set on the field container element (likely in `packages/ui/components/field/field.tsx` — `FieldRootContainer`) so that `cqw` units resolve relative to the field box itself.

## Steps to Reproduce

1. Create an envelope via API with a PDF document
2. Add a TEXT field with:
   - `readOnly: true` in fieldMeta
   - `text: "<long paragraph>"` in fieldMeta
   - `width: 82` (or any width)
3. Distribute the envelope
4. Open the signing link in a browser
5. Observe the pre-filled text extends past the right edge of the field border

## Environment

- Documenso: Self-hosted (latest)
- API: v2
- Browser: Chrome (macOS)

yzxcj797:
## Summary

Fixes #2669.

## Root cause (as diagnosed in the issue, verified)

Field text in the signing preview sizes its font with `cqw` units (`text-[clamp(0.425rem,25cqw,0.825rem)]`), but `FieldRootContainer` never established a containment context — so the units resolved against the nearest ancestor container (the page), and pre-filled readOnly TEXT fields wrapped at page-relative widths, overflowing the field border. The sealed PDF (Konva, field-relative) rendered correctly, and changing the field's `width` never changed the wrap point — both consistent with the diagnosis.

The telling detail: **the editor flow already applies the same containment for the same text** (`field-item.tsx` uses `[container-type:size]`), so the two flows disagreed about the container and the overflow only appeared in the signing preview.

## Fix

`[container-type:inline-size]` on `FIELD_ROOT_CONTAINER_CLASS_NAME` — every `cqw` inside a signing field (TEXT, DATE, the typed-signature sizing in `signing-card.tsx`) now resolves against the field box, matching the editor's behavior. `inline-size` rather than `size` deliberately, so height layout is untouched.

### DIFF
--- a/packages/ui/lib/field-root-container-classes.ts
+++ b/packages/ui/lib/field-root-container-classes.ts
@@ -1,7 +1,12 @@
 const FIELD_ROOT_CONTAINER_SHARED_CLASS_NAME =
   'field--FieldRootContainer field-card-container dark-mode-disabled group rounded-[2px] bg-white/90 ring-2 ring-gray-200 transition-all';
 
-export const FIELD_ROOT_CONTAINER_CLASS_NAME = `${FIELD_ROOT_CONTAINER_SHARED_CLASS_NAME} relative z-20 flex h-full w-full items-center`;
+// [container-type:inline-size] makes the cqw font units used by field text
+// resolve against the field box itself; without it they resolve against the
+// nearest ancestor container (the page), so pre-filled readOnly TEXT fields
+// wrapped past the field border in the signing preview (#2669). The editor's
+// field-item sets the same containment for the same reason.
+export const FIELD_ROOT_CONTAINER_CLASS_NAME = `${FIELD_ROOT_CONTAINER_SHARED_CLASS_NAME} relative z-20 flex h-full w-full items-center [container-type:inline-size]`;
 
 export const FIELD_ROOT_CONTAINER_PROBE_CLASS_NAME = `field--FieldRootContainerProbe ${FIELD_ROOT_CONTAINER_SHARED_CLASS_NAME}`;
 


Mythie:

https://documen.so/cla

PR #3253, opened 2026-08-18
feat: track last used timestamp for API tokens

yzxcj797:
## Summary

Implements #2756.

## What this does (the issue's proposed shape, end to end)

- **Schema + migration**: `ApiToken.lastUsedAt DateTime?` with the `ALTER TABLE` migration (nullable — unused tokens stay `null`, no table rewrite).
- **Recording usage**: `getApiTokenByToken` stamps `lastUsedAt` on every successful validation as a **fire-and-forget** write — placed after all auth/disabled/expiry checks so only genuinely valid uses count, and `.catch`-guarded so a failed stamp can never fail the API request it recorded.
- **Surfacing it**: the token list query selects the field, the tRPC response schema exposes it (picked from the generated Prisma zod schema), and the team settings table gains a **"Last used"** column — "Never" when null — between Created and Expires.

This gives admins the primary token-hygiene signal the issue describes: an actively used token (deleting it breaks an integration) is distinguishable at a glance from one created months ago and never used.

### DIFF
--- a/apps/remix/app/routes/_authenticated+/t.$teamUrl+/settings.tokens.tsx
+++ b/apps/remix/app/routes/_authenticated+/t.$teamUrl+/settings.tokens.tsx
@@ -47,6 +47,20 @@ export default function ApiTokensPage() {
         cell: ({ row }) => i18n.date(row.original.createdAt),
       },
       {
+        header: t`Last used`,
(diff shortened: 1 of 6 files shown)


Mythie:

https://documen.so/cla

PR #2770, opened 2026-05-06
feat: dynamic OpenGraph preview for /sign/<token> links

zomars:
## Summary

Adds a per-document OpenGraph image and meta tags to signing links so previews on WhatsApp, iMessage, Slack, Twitter, etc. communicate what the link actually is — instead of falling back to the generic site-wide marketing image, which reads as a phishing signal to recipients who don't already know Documenso.

Closes #2769.

### Before / After

**Before** (current generic preview, captured in WhatsApp on a real `app.documenso.com/sign/<token>` link — see #2769):

Image

**After** (this PR — actual rendered output of the new endpoint):

og

Verified on opengraph.xyz

imagen
imagen

English (`documentMeta.language = "en"`):
- ● SIGNATURE REQUESTED
- **Alice Example has sent you a document to sign**
- DOCUMENT — *Mutual NDA — Acme Inc. & Globex Corp.*
- For Bob Recipient
- Documenso · The open-source signing platform

German (`documentMeta.language = "de"`, with translation provided):
- ● SIGNATUR ANGEFRAGT
- **Anna Müller hat Ihnen ein Dokument zur Unterschrift gesendet**
- DOKUMENT — *Geheimhaltungsvereinbarung — Acme GmbH & Globex AG*
- Für Max Mustermann
- Documenso · Die Open-Source-Signaturplattform

## Implementation

- **New resource route**: `apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx` — renders a 1200×630 PNG via the existing `satori` + `svgToPng` pipeline. Mirrors `_share+/share.$slug.opengraph.tsx` 1:1 in transport, font-loading, and response headers. Uses `getDocumentAndSenderByToken({ requireAccessAuth: false })` so the OG crawler doesn't need a session.
- **`meta` export added** on `_index.tsx` — points `og:image` / `twitter:image` at the new endpoint, sets `twitter:card=summary_large_image`, plus generic `og:title` / `og:description` for crawlers that ignore the image.
- **Localised** via `getI18nInstance(documentMeta.language)` — same per-locale isolated-instance pattern as `generate-certificate-pdf.ts`. Avoids the global-singleton races you'd get from `i18n.loadAndActivate`.
[...]

PR #3254, opened 2026-08-18
feat: dynamic OpenGraph preview for /sign/<token> links (rebase of #2770)

### LINKED ISSUE #2769: Add dynamic OpenGraph preview for /sign/<token> links to build trust in social/messaging shares
## Problem

When a signing link (`https://app.documenso.com/sign/<token>`) is shared via WhatsApp, iMessage, Slack, Twitter, LinkedIn, etc., the link preview falls back to the generic site-wide OpenGraph image and copy:

- **Image:** `/opengraph-image.jpg` (Documenso logo)
- **Title:** "Documenso - The Open Source DocuSign Alternative"
- **Description:** Generic marketing copy

Example as it appears in WhatsApp today:

Image

For a recipient who has never heard of Documenso, this preview gives **zero context** about what the link is. They see a marketing tagline for a product they don't recognize, attached to a URL someone just sent them. That pattern reads as a phishing/spam signal and meaningfully hurts the chance the recipient actually clicks through and signs.

## Proposal

Generate a **dynamic OpenGraph image and meta tags** specifically for `/sign/<token>` (and its siblings: `complete`, `expired`, `rejected`, `waiting`).

The preview should make the request unambiguous at a glance, e.g.:

> **\"Alice from Acme Inc. has sent you a document to sign\"**
> Document: *Mutual NDA - 2026-05*
> Powered by Documenso

Equivalent `og:title` / `og:description` / `twitter:*` meta tags should also be set so clients that don't render the image (or that crawl text only) still get useful context.

## Why this matters

- **Trust:** Recipients are more likely to click and sign when the preview matches what the sender told them ("I sent you the NDA"). The current preview actively undermines that.
- **Phishing perception:** A generic logo + unfamiliar brand + opaque token URL is exactly the shape of a phishing message. Personalising the preview removes that ambiguity.
- **Conversion:** Fewer abandoned signing flows from recipients who hesitate at the link-share step.
- **Brand:** Every shared link becomes a small, contextual ad for Documenso to recipients seeing it for the first time — but as a useful, document-specific message rather than a generic tagline.

## Implementation notes

There's already precedent in the codebase for dynamic OG images: the share route uses a sibling resource route to render imagery per resource — see `apps/remix/app/routes/_share+/share.\$slug.opengraph.tsx` paired with `share.\$slug.tsx`. The same pattern can be applied to the sign route.

Suggested shape:

- Add `apps/remix/app/routes/_recipient+/sign.\$token+/_index.opengraph.tsx` (resource route) that:
  - Looks up the recipient + document by token
  - Returns an image with sender name, recipient name (or just first name / initials for privacy), document title, and Documenso branding
- Add a `meta` export in `sign.\$token+/_index.tsx` that overrides the root `appMetaTags()` with:
  - `og:title` → e.g. \"Sign \\\"\${documentTitle}\\\"\"
  - `og:description` → e.g. \"\${senderName} has sent you a document to sign via Documenso.\"
  - `og:image` / `twitter:image` → the dynamic OG route above
  - `twitter:card` → `summary_large_image`
- Apply the same treatment to `complete`, `expired`, `rejected`, `waiting` siblings, with status-appropriate copy.

## Privacy considerations

The token URL is already shared with the recipient (and anyone they forward it to), so revealing the document title and sender name in the preview does not expose data the recipient doesn't already have access to. Worth confirming with the team whether to:

- Show the full document title or a redacted version
- Show the sender's full name or just first name + organisation
- Allow senders/teams to opt out via a setting (some teams may prefer the current generic preview for highly confidential documents)

## Out of scope

- Localisation of the OG image text (can follow existing i18n patterns later)
- Caching strategy for the rendered images (likely deferred to a follow-up)

zomars:
> **Reopens #2770**, which was auto-closed by the stale bot on 2026-08-12 ("closed because of inactivity"). GitHub's API refuses to reopen it (`UNPROCESSABLE — Could not open the pull request.`), so this PR carries the same work forward.
>
> Same two commits, **rebased onto current `main`** (the old branch was 196 commits behind). Conflicts were in the generated Lingui catalogs and one import line in `_index.tsx`; catalogs were regenerated from current `main` with `lingui extract` and filtered down to only the msgids this feature introduces, so there is no unrelated catalog churn. `npm run typecheck -w @documenso/remix` passes on the rebased tree.
>
> Context: #2769 · https://github.com/documenso/documenso/issues/2769#issuecomment-4390385979

## Summary

Adds a per-document OpenGraph image and meta tags to signing links so previews on WhatsApp, iMessage, Slack, Twitter, etc. communicate what the link actually is — instead of falling back to the generic site-wide marketing image, which reads as a phishing signal to recipients who don't already know Documenso.

Closes #2769.

### Before / After

**Before** (current generic preview, captured in WhatsApp on a real `app.documenso.com/sign/<token>` link — see #2769):

Image

**After** (this PR — actual rendered output of the new endpoint):

og

Verified on opengraph.xyz

imagen
imagen

English (`documentMeta.language = "en"`):
- ● SIGNATURE REQUESTED
- **Alice Example has sent you a document to sign**
- DOCUMENT — *Mutual NDA — Acme Inc. & Globex Corp.*
- For Bob Recipient
- `Sign document →` CTA pill (bottom right)

German (`documentMeta.language = "de"`, with translation provided):
- ● SIGNATUR ANGEFRAGT
- **Anna Müller hat Ihnen ein Dokument zur Unterschrift gesendet**
- DOKUMENT — *Geheimhaltungsvereinbarung — Acme GmbH & Globex AG*
- Für Max Mustermann
- `Dokument unterschreiben →` CTA pill (bottom right)

## Implementation

- **New resource route**: `apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx` — renders a 1200×630 PNG via the existing `satori` + `svgToPng` pipeline. Mirrors `_share+/share.$slug.opengraph.tsx` 1:1 in transport, font-loading, and response headers. Uses `getDocumentAndSenderByToken({ requireAccessAuth: false })` so the OG crawler doesn't need a session.
- **`meta` export added** on `_index.tsx` — points `og:image` / `twitter:image` at the new endpoint, sets `twitter:card=summary_large_image`, plus generic `og:title` / `og:description` for crawlers that ignore the image.
- **Localised** via `getI18nInstance(documentMeta.language)` — same per-locale isolated-instance pattern as `generate-certificate-pdf.ts`. Avoids the global-singleton races you'd get from `i18n.loadAndActivate`.
- **CJK fonts loaded conditionally** — Inter alone can't render Japanese/Korean/Chinese glyphs, so when `documentMeta.language` is `ja`/`ko`/`zh` we additionally load `noto-sans-japanese.ttf` / `noto-sans-korean.ttf` / `noto-sans-chinese.ttf` (already in `apps/remix/public/fonts/`).
- **Lingui catalogs updated** — ran `npm run translate:extract`; the new msgids land in every locale catalog with empty `msgstr` for translators to fill (Crowdin handles the rest per `crowdin.yml`).

## Privacy

The token URL already exposes the document title and recipient identity to anyone with the link, so revealing them in the preview doesn't leak data the recipient (or anyone they forward it to) doesn't already have. Worth team discussion whether to:

- Render full document title vs. a redacted version
- Render full sender name vs. first-name + organisation
- Add a per-document or per-team opt-out for highly confidential workflows (out of scope for this PR — happy to follow up if there's appetite)

## What I deliberately deferred

- **Status routes** (`complete` / `expired` / `rejected` / `waiting`) — they currently inherit the root-level generic OG. Same treatment can be added in a follow-up PR with status-appropriate copy. Wanted to keep this PR small and reviewable.
- **Cache headers**: 5-minute `Cache-Control` instead of immutable, so iterating on the design isn't gated behind CDN cache. Probably fine to bump this later once the design is stable.

## Test plan

- [ ] `npm run typecheck -w @documenso/remix` (passes locally)
- [ ] `curl -I http://localhost:<port>/sign/<token>/opengraph` returns `200 image/png` for a valid token
- [ ] Same path returns `404` for a non-existent token (verified locally)
- [ ] `<head>` of `/sign/<token>` contains `og:image`, `twitter:card=summary_large_image`, etc. pointing at the new endpoint (verified locally)
- [ ] Visit `https://www.opengraph.xyz/?url=<tunneled-staging-url>/sign/<token>` and confirm WhatsApp/Twitter/iMessage previews render correctly
- [ ] Visit a `documentMeta.language='de'` (or any non-English) signing URL and confirm the preview text reflects the document language

### DIFF
--- a/apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
+++ b/apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
@@ -389,8 +389,9 @@ export const meta = ({ params }: Route.MetaArgs) => {
   const { token } = params;
   const baseUrl = NEXT_PUBLIC_WEBAPP_URL();
   const ogImageUrl = `${baseUrl}/sign/${token}/opengraph`;
-  const title = 'You have a document to sign';
-  const description = 'Review and sign this document securely on Documenso.';
+  const title = 'A document is ready for your secure signature on Documenso';
+  const description =
+    'Review and sign this document securely from your browser. Documenso is the open-source signing platform — no account required to sign.';
 
   return [
     { title: `${title} - Documenso` },
--- a/apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
+++ b/apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
@@ -30,6 +30,8 @@ import { extractDocumentAuthMethods } from '@documenso/lib/utils/document-auth';
 import { isRecipientExpired } from '@documenso/lib/utils/recipients';
 import { prisma } from '@documenso/prisma';
 import { SigningCard3D } from '@documenso/ui/components/signing-card';
+import { i18n } from '@lingui/core';
+import { msg } from '@lingui/core/macro';
 import { Trans } from '@lingui/react/macro';
 import { DocumentSigningOrder, DocumentStatus, RecipientRole, SigningStatus } from '@prisma/client';
 import { Clock8 } from 'lucide-react';
@@ -389,12 +391,17 @@ export const meta = ({ params }: Route.MetaArgs) => {
   const { token } = params;
   const baseUrl = NEXT_PUBLIC_WEBAPP_URL();
   const ogImageUrl = `${baseUrl}/sign/${token}/opengraph`;
-  const title = 'A document is ready for your secure signature on Documenso';
-  const description =
-    'Review and sign this document securely from your browser. Documenso is the open-source signing platform — no account required to sign.';
+
+  const title = i18n._(msg`A document is waiting for your signature`);
+  const description = i18n._(
+    msg`Review and sign this document securely from your browser with Documenso, the open source signing platform.`,
+  );
 
   return [
-    { title: `${title} - Documenso` },
+    { title: i18n._(msg`Sign Document - Documenso`) },
+    // A leaf `meta` export replaces the parent's, so the recipient layout's
+    // robots tag has to be re-emitted here.
+    { name: 'robots', content: 'noindex, nofollow, noarchive, nosnippet, noimageindex' },
     { name: 'description', content: description },
     { property: 'og:title', content: title },
     { property: 'og:description', content: description },
--- a/apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
+++ b/apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
@@ -8,7 +8,7 @@ import { readCscSadSessionFromRequest } from '@documenso/ee/server-only/signing/
 import { readCscServiceSessionFromRequest } from '@documenso/ee/server-only/signing/csc/cookies/service-session-cookie';
 import { EnvelopeRenderProvider } from '@documenso/lib/client-only/providers/envelope-render-provider';
 import { useOptionalSession } from '@documenso/lib/client-only/providers/session';
-import { IS_INSTANCE_CSC_MODE } from '@documenso/lib/constants/app';
+import { IS_INSTANCE_CSC_MODE, NEXT_PUBLIC_WEBAPP_URL } from '@documenso/lib/constants/app';
 import { AppError, AppErrorCode } from '@documenso/lib/errors/app-error';
 import { loadRecipientBrandingByTeamId } from '@documenso/lib/server-only/branding/load-recipient-branding';
 import { getDocumentAndSenderByToken } from '@documenso/lib/server-only/document/get-document-by-token';
@@ -384,6 +384,27 @@ export async function loader(loaderArgs: Route.LoaderArgs) {
     branding,
   } as const);
 }
+
+export const meta = ({ params }: Route.MetaArgs) => {
+  const { token } = params;
+  const baseUrl = NEXT_PUBLIC_WEBAPP_URL();
+  const ogImageUrl = `${baseUrl}/sign/${token}/opengraph`;
+  const title = 'You have a document to sign';
+  const description = 'Review and sign this document securely on Documenso.';
+
+  return [
+    { title: `${title} - Documenso` },
+    { name: 'description', content: description },
+    { property: 'og:title', content: title },
+    { property: 'og:description', content: description },
+    { property: 'og:image', content: ogImageUrl },
+    { property: 'og:type', content: 'website' },
+    { name: 'twitter:card', content: 'summary_large_image' },
+    { name: 'twitter:title', content: title },
+    { name: 'twitter:description', content: description },
+    { name: 'twitter:image', content: ogImageUrl },
+  ];
+};
 
 export default function SigningPage() {
   const data = useSuperLoaderData<typeof loader>();
--- a/apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
+++ b/apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
@@ -65,7 +65,7 @@ export const loader = async ({ params }: Route.LoaderArgs) => {
     headline: i18n._(msg`${sender} has sent you a document to sign`),
     documentLabel: i18n._(msg`Document`),
     forRecipient: recipientName ? i18n._(msg`For ${recipientName}`) : null,
-    tagline: i18n._(msg`The open-source signing platform`),
+    cta: i18n._(msg`Sign document`),
   };
 
   const svg = await satori(
@@ -189,7 +189,22 @@ export const loader = async ({ params }: Route.LoaderArgs) => {
         }}
       >
         <img src={`${baseUrl}/static/logo.png`} alt="Documenso" width={204} height={30} />
-        <span style={{ color: '#94a3b8', fontSize: '20px' }}>{labels.tagline}</span>
+        <div
+          style={{
+            display: 'flex',
+            alignItems: 'center',
+            gap: '10px',
+            padding: '14px 28px',
+            borderRadius: '999px',
+            backgroundColor: '#a2e771',
+            color: '#0f172a',
+            fontSize: '22px',
+            fontWeight: 600,
+          }}
+        >
+          {labels.cta}
+          <span style={{ fontSize: '24px' }}>→</span>
+        </div>
       </div>
     </div>,
     {
--- a/apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
+++ b/apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
@@ -1,0 +1,218 @@
+import { msg } from '@lingui/core/macro';
+import satori from 'satori';
+
+import { getI18nInstance } from '@documenso/lib/client-only/providers/i18n-server';
+import { NEXT_PUBLIC_WEBAPP_URL } from '@documenso/lib/constants/app';
+import { getDocumentAndSenderByToken } from '@documenso/lib/server-only/document/get-document-by-token';
+import { svgToPng } from '@documenso/lib/utils/images/svg-to-png';
+
+import type { Route } from './+types/opengraph';
+
+const IMAGE_SIZE = {
+  width: 1200,
+  height: 630,
+};
+
+const truncate = (value: string, max: number) =>
+  value.length > max ? `${value.slice(0, max - 1).trimEnd()}…` : value;
+
+export const loader = async ({ params }: Route.LoaderArgs) => {
+  const { token } = params;
+
+  if (!token) {
+    return new Response('Not found', { status: 404 });
+  }
+
+  const baseUrl = NEXT_PUBLIC_WEBAPP_URL();
+
+  const document = await getDocumentAndSenderByToken({
+    token,
+    requireAccessAuth: false,
+  }).catch(() => null);
+
+  if (!document) {
+    return new Response('Not found', { status: 404 });
+  }
+
+  const fetchFont = async (path: string) =>
+    fetch(new URL(`${baseUrl}${path}`, import.meta.url)).then(async (res) => res.arrayBuffer());
+
+  const cjkFontPath = ((): string | null => {
+    const lang = document.documentMeta?.language ?? 'en';
+    if (lang === 'ja') return '/fonts/noto-sans-japanese.ttf';
+    if (lang === 'ko') return '/fonts/noto-sans-korean.ttf';
+    if (lang === 'zh') return '/fonts/noto-sans-chinese.ttf';
+    return null;
+  })();
+
+  const [interSemiBold, interRegular, interBold, cjkFont] = await Promise.all([
+    fetchFont('/fonts/inter-semibold.ttf'),
+    fetchFont('/fonts/inter-regular.ttf'),
+    fetchFont('/fonts/inter-bold.ttf'),
+    cjkFontPath ? fetchFont(cjkFontPath) : Promise.resolve(null),
+  ]);
+
+  const sender = document.user.name || document.user.email;
+  const recipient = document.recipients[0];
+  const recipientName = recipient?.name || recipient?.email || null;
+  const documentTitle = truncate(document.title, 80);
+
+  const language = document.documentMeta?.language ?? 'en';
+  const i18n = await getI18nInstance(language);
+
+  const labels = {
+    badge: i18n._(msg`Signature requested`),
+    headline: i18n._(msg`${sender} has sent you a document to sign`),
+    documentLabel: i18n._(msg`Document`),
+    forRecipient: recipientName ? i18n._(msg`For ${recipientName}`) : null,
+    tagline: i18n._(msg`The open-source signing platform`),
+  };
+
+  const svg = await satori(
+    <div
+      style={{
+        display: 'flex',
+        flexDirection: 'column',
+        height: '100%',
+        width: '100%',
+        backgroundColor: '#ffffff',
+        padding: '72px 80px',
+        fontFamily: 'Inter',
+        position: 'relative',
+      }}
+    >
+      <div
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+          gap: '12px',
+        }}
+      >
+        <div
+          style={{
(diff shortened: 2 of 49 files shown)


packages/lib/translations/de/web.po
```diff
@@ -496,6 +496,10 @@ msgstr "{remaningLength, plural, one {# Zeichen verbleibend} other {# Zeichen ve
 msgid "{selectedCount} selected"
```
catalinpit, line 506: All the translation files should be removed from the PR

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
```diff
@@ -0,0 +1,233 @@
...
+export const loader = async ({ params }: Route.LoaderArgs) => {
+  const { token } = params;
+
+  if (!token) {
+    return new Response('Not found', { status: 404 });
+  }
+
+  const baseUrl = NEXT_PUBLIC_WEBAPP_URL();
+
+  const document = await getDocumentAndSenderByToken({
```
catalinpit: What happens to the documents that have access restrictions? For an ACCOUNT-gated doc, the signing page hides information behind the "Authentication required" view and it only shows the recipient email. But `/sign/<token>/opengraph` renders title, sender, and recipient to anyone who accesses it. Pasting the link exposes the data the auth gate hide.

apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
```diff
@@ -385,6 +385,28 @@ export async function loader(loaderArgs: Route.LoaderArgs) {
   } as const);
 }
 
+export const meta = ({ params }: Route.MetaArgs) => {
+  const { token } = params;
+  const baseUrl = NEXT_PUBLIC_WEBAPP_URL();
+  const ogImageUrl = `${baseUrl}/sign/${token}/opengraph`;
+  const title = 'A document is ready for your secure signature on Documenso';
```
catalinpit: Title could be improved

apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
```diff
@@ -385,6 +385,28 @@ export async function loader(loaderArgs: Route.LoaderArgs) {
...
+  const description =
+    'Review and sign this document securely from your browser. Documenso is the open-source signing platform — no account required to sign.';
```
catalinpit: Same for this, especially that it contains incorrect information like `no account required to sign`. Some documents require an account to sign, depending on what the document owner sets.

apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
```diff
@@ -385,6 +385,28 @@ export async function loader(loaderArgs: Route.LoaderArgs) {
...
+
+  return [
```
catalinpit, line 400: Make sure that it doesn't get indexed in search engines. 

See the tags in `apps/remix/app/routes/_recipient+/_layout.tsx`.

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
```diff
@@ -0,0 +1,233 @@
...
+  const [interSemiBold, interRegular, interBold, cjkFont] = await Promise.all([
+    fetchFont('/fonts/inter-semibold.ttf'),
+    fetchFont('/fonts/inter-regular.ttf'),
+    fetchFont('/fonts/inter-bold.ttf'),
+    cjkFontPath ? fetchFont(cjkFontPath) : Promise.resolve(null),
+  ]);
+
+  const sender = document.user.name || document.user.email;
+  const recipient = document.recipients[0];
+  const recipientName = recipient?.name || recipient?.email || null;
```
catalinpit: Maybe truncate here too, if the name/email is too long. Just an idea

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+const IMAGE_SIZE = {
+  width: 1200,
+  height: 630,
+};
+
+const truncate = (value: string, max: number) =>
+  value.length > max ? `${value.slice(0, max - 1).trimEnd()}…` : value;
+
+export const loader = async ({ params }: Route.LoaderArgs) => {
```
catalinpit: What about the `DRAFT` documents?

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
```diff
@@ -0,0 +1,233 @@
...
+  const document = await getDocumentAndSenderByToken({
+    token,
+    requireAccessAuth: false,
+  }).catch(() => null);
+
+  if (!document) {
+    return new Response('Not found', { status: 404 });
+  }
+
+  const fetchFont = async (path: string) =>
```
catalinpit: What happens if this fails?

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+  const cjkFontPath = ((): string | null => {
+    const lang = document.documentMeta?.language ?? 'en';
+    if (lang === 'ja') return '/fonts/noto-sans-japanese.ttf';
+    if (lang === 'ko') return '/fonts/noto-sans-korean.ttf';
+    if (lang === 'zh') return '/fonts/noto-sans-chinese.ttf';
+    return null;
+  })();
+
+  const [interSemiBold, interRegular, interBold, cjkFont] = await Promise.all([
```
catalinpit: Can't this be handled better? Examples in `/packages/lib/server-only/pdf/helpers.ts`

catalinpit (changes requested):
Thanks for the PR!

zomars:
Thanks a lot for the thorough review, @catalinpit — really appreciate you taking the time to go through this line by line. Every point you raised is valid; I went through each one against the code and confirmed them.

I'll address all of them as soon as possible and push an update. Short summary of what's coming:

- **Drop all translation catalogs** from the PR — the `.po` files are owned by the `chore: extract translations` automation + Crowdin, so no catalog churn here.
- **Access-gated documents** — the OG route will apply the same access rules as the signing page instead of `requireAccessAuth: false`, so `ACCOUNT`-gated (and other auth-gated) documents fall back to the generic image rather than exposing title/sender/recipient.
- **`DRAFT` documents** — 404, matching the signing page loader; will also handle non-pending states (completed / rejected / expired) rather than always rendering "Signature requested".
- **`meta` overriding the layout's `robots` tag** — confirmed: a leaf `meta` export replaces the parent's, which was silently dropping `noindex, nofollow, …` from `_layout.tsx`. Will re-emit it (and the localized title) from the route.
- **Title & description** — rewriting both; you're right that "no account required to sign" is plainly wrong for documents with account access auth.
- **Truncation** — will apply to sender and recipient name/email too, not just the document title.
- **Font loading** — switching to the `ensureFontLibrary`-style local filesystem read instead of `fetch`-ing our own public URL, and adding proper failure handling so the endpoint degrades to the generic image instead of returning an HTML error page for an `` URL.

Will ping you here once it's ready for another look. Thanks again!

packages/lib/translations/de/web.po
zomars: Removed. All 11 catalogs are reverted to `main`, so the diff is only the two feature files. The new msgids will come through `chore: extract translations` and Crowdin.

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
zomars: Fixed. The loader runs `extractDocumentAuthMethods` over the document and recipient auth options. When access auth is required it renders a card with no title, no sender and no recipient, so the crawler sees no more than the auth gate shows. Verified against a seeded `ACCOUNT` gated document.

apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
zomars: Rewritten to `A document is waiting for your signature`, and localized. It was hardcoded English while the rest of the feature goes through lingui. The `<title>` is localized too, reusing the msgid the layout already declares.

apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
zomars: Removed. That claim is false for documents with account access auth. The description is now `Review and sign this document securely from your browser with Documenso, the open source signing platform.`, localized like the title, and it makes no claim about what the document requires.

apps/remix/app/routes/_recipient+/sign.$token+/_index.tsx
zomars: Confirmed, and it was worse than it looked. In React Router a leaf `meta` export replaces the parent's rather than merging, so this dropped the `robots` tag from `_layout.tsx` entirely. With the PR applied the signing page had no `noindex` at all. It is re-emitted from the route now, along with the localized title, and verified in the rendered `<head>`.

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
zomars: Done, for the sender and the recipient, not just the title. The budget is measured in display units rather than characters, because full width scripts take about twice the advance of Latin at the same size. A character count either cut English short or let Chinese overflow. Titles break on a word boundary when one is near the limit.

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
zomars: `404` now, matching the `status === DocumentStatus.DRAFT` check in the signing page loader. Deleted documents `404` too. States that are no longer an actionable signature request, meaning completed, signed, rejected or expired, get a neutral card instead of a preview claiming a signature is requested.

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
zomars: It served an HTML error page on a URL that is consumed as an `` source, which crawlers then cache. The render is wrapped now and failures serve an image instead. The specific failure you were pointing at is also gone: fonts are no longer fetched over HTTP and the logo is embedded as a data URI, so the render makes no network calls.

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
zomars: Switched to the `ensureFontLibrary` approach, reading from `path.join(process.cwd(), 'public')` with a per process cache. No self request against `NEXT_PUBLIC_WEBAPP_URL`.

This turned up a bug in the old code. `noto-sans-japanese.ttf` and `noto-sans-korean.ttf` are variable fonts, and satori's opentype parser throws on their `fvar` table, so the CJK branch would have taken down the whole render for those languages rather than degrading it. They were also registered as `Inter` at weight 400, colliding with `inter-regular`. `noto-sans-chinese.ttf` parses and is registered under its own family. `ja` and `ko` fall back to the neutral card. Static builds of those two faces, ideally subset, would be a small follow up and would help any other satori surface.

zomars:
Evidence. Every image is the actual response of `/sign/<token>/opengraph`, captured against a local server with a seeded database.

| Scenario | Status | Response |
| --- | --- | --- |
| `PENDING`, no access auth | `200` | 1200x630 PNG with details |
| `PENDING`, access auth required | `200` | 1200x630 PNG, no details |
| Completed, signed, rejected, expired | `200` | 1200x630 PNG, neutral |
| `ja` and `ko` | `200` | 1200x630 PNG, neutral |
| `DRAFT` or deleted | `404` | |
| Unknown token | `404` | |

Signature request, English.



Localized, Chinese. The document label and the CTA come from the existing catalog. The badge and headline are new msgids that Crowdin has not filled yet, which is expected since the catalogs are not in this PR.



Access gated document. This is what a crawler gets for an `ACCOUNT` gated link. No title, no sender, no recipient.



No longer a signature request, meaning completed, rejected or expired.



Long title and recipient. 94 character title, 43 character name, no truncation needed.



Worst case. 178 character title, 47 character sender. Truncation lands on a word boundary and the footer keeps its padding.



Worst case in Chinese. 46 characters is 92 display units and fills exactly two lines. A character based limit would have overflowed to three.



Rendered `<head>` of the signing page.

```html
<title>Sign Document - Documenso</title>
<meta name="robots" content="noindex, nofollow, noarchive, nosnippet, noimageindex"/>
<meta property="og:title" content="A document is waiting for your signature"/>
<meta property="og:image" content="http://localhost:3000/sign/<token>/opengraph"/>
<meta name="twitter:card" content="summary_large_image"/>
```

zomars:
@catalinpit thanks for the review. The access auth and `robots` findings were both real leaks, and chasing the font comment turned up a third bug.

All nine comments are addressed, each answered in its own thread, with evidence in the comment above. `npm run typecheck -w @documenso/remix` and `biome check` pass. Ready for another look.

apps/remix/app/routes/_recipient+/sign.$token+/opengraph.tsx
```diff
@@ -0,0 +1,444 @@
...
+/**
+ * Inter cannot render Chinese, Japanese or Korean glyphs, so we register the
+ * matching Noto Sans family as a fallback for those document languages.
+ *
+ * `noto-sans-japanese.ttf` and `noto-sans-korean.ttf` are variable fonts, which
+ * satori's opentype parser cannot read (it throws on the `fvar` table). Those
+ * languages fall back to an English card until static builds of those faces are
+ * added to `public/fonts`.
+ */
+const CJK_FONTS: Record<string, { family: string; file: string } | null> = {
```
catalinpit: Why did you pick these languages, but not others for example?

PR #3255, opened 2026-08-18
fix(i18n): allow exact matching of region locales in accept-language header negotiation

### LINKED ISSUE #3090: `Accept-Language` negotiation can never match `pt-BR` — Brazilian users always get English, and the `lang` cookie then locks it in
## Summary

Browser-language detection is broken for Brazilian Portuguese: a browser sending `Accept-Language: pt-BR,pt;q=0.9,…` always gets the UI in **English**, even though `pt-BR` is a fully translated, supported locale.

The root cause is that `parseLanguageFromLocale` strips the region subtag *before* comparing against `SUPPORTED_LANGUAGE_CODES` — but `pt-BR` is the only entry in that list that *has* a region subtag, so it is unreachable by header negotiation:

```ts
// packages/lib/utils/i18n.ts
const parseLanguageFromLocale = (locale: string): SupportedLanguageCodes | null => {
  const [language, _country] = locale.split('-');   // 'pt-BR' -> 'pt'

  const foundSupportedLanguage = APP_I18N_OPTIONS.supportedLangs.find(
    (lang): lang is SupportedLanguageCodes => lang === language,  // 'pt' === 'pt-BR' -> never
  );
  ...
```

```ts
// packages/lib/constants/locales.ts
const SUPPORTED_LANGUAGE_CODES = ['de', 'en', 'fr', 'es', 'it', 'nl', 'pl', 'pt-BR', 'ja', 'ko', 'zh'];
```

Every other supported code is a bare language subtag, so `de-DE -> de`, `fr-FR -> fr`, `zh-CN -> zh` all work. Only `pt-BR` (→ `pt` → no match → fallback `en`) can never be negotiated.

This matters especially for **anonymous recipients on the signing page**: per #2957 the intended behaviour is that recipients "always try to render based on the browser language" — this bug makes exactly that impossible for the entire pt-BR audience, on the one page whose visitors never have an account or a language switcher.

## Aggravating factor: the fallback gets persisted for 2 years

`handleRequest` resolves the language (cookie first, then `Accept-Language`), and the root loader then re-serializes the resolved value into the `lang` cookie (`Max-Age` 2 years):

```ts
// apps/remix/app/entry.server.tsx
let language = await langCookie.parse(request.headers.get('cookie') ?? '');
if (!APP_I18N_OPTIONS.supportedLangs.includes(language)) {
  language = extractLocaleData({ headers: request.headers }).lang;
}
```

So a pt-BR visitor's first request stores `lang="en"`, and since `"en"` is a valid supported code, the header is never consulted again. Even after this bug is fixed, previously affected visitors will stay locked on English until the cookie expires or is overwritten.

## Steps to reproduce

Tested on self-hosted `documenso/documenso:v2.14.0`; the code above is unchanged on `main` as of this report.

1. Fresh browser profile / incognito with the browser language set to Brazilian Portuguese (sends `Accept-Language: pt-BR,pt;q=0.9,en-US;q=0.8,en;q=0.7`).
2. Open any page — `/signin`, or a recipient signing link `/sign/<token>`.
3. **Actual:** UI renders in English, `<html lang="en">`, and the response sets the `lang` cookie to `"en"` (base64 `ImVuIg==`) for 2 years.
4. **Expected:** UI renders in Portuguese, `<html lang="pt-BR">`.

Proof that only negotiation is broken (translations are fine): send the request with the cookie pre-set to pt-BR —

```bash
curl -s https://<instance>/signin -H 'Cookie: lang=InB0LUJSIg=='   # base64('"pt-BR"')
# -> <html translate="no" lang="pt-BR" ...>, fully translated page
```

## Related nit in the same function

Entries carrying quality factors and no region are also never matched, because the `;q=` suffix is not stripped: `'pt;q=0.9'.split('-')` → `'pt;q=0.9'`. Today this is masked by region-qualified entries appearing first in real-world headers, but it makes the fallback chain weaker than it looks.

## Suggested fix

Match the exact code first (case-insensitively, after stripping parameters), then fall back to a language-subtag match against the supported codes' own language subtags:

```ts
const parseLanguageFromLocale = (locale: string): SupportedLanguageCodes | null => {
  const cleaned = locale.split(';')[0].trim(); // drop ';q=…'

  const exact = APP_I18N_OPTIONS.supportedLangs.find(
    (lang) => lang.toLowerCase() === cleaned.toLowerCase(),
  );
  if (exact) return exact;

  const [language] = cleaned.split('-');

  return (
    APP_I18N_OPTIONS.supportedLangs.find((lang) => lang.split('-')[0] === language) ?? null
  );
};
```

With this, `pt-BR` matches exactly, and `pt` / `pt-PT` / `pt;q=0.9` fall back to `pt-BR` (the only Portuguese variant available), while the behaviour for all other locales is unchanged. Happy to open a PR if the approach sounds good to you.

misinierijon4-debug:
### Problem
Browser-language detection for Brazilian Portuguese (`pt-BR`) fails to match the supported locale during `Accept-Language` header negotiation. Because `pt-BR` is the only supported language code in `APP_I18N_OPTIONS.supportedLangs` that contains a region subtag, stripping the subtag beforehand caused negotiation to compare `'pt'` against `'pt-BR'`, always failing and falling back to English.

### Root cause
In `packages/lib/utils/i18n.ts` (`parseLanguageFromLocale`), `locale.split('-')` extracted only the language subtag before checking against `APP_I18N_OPTIONS.supportedLangs`. Since `pt-BR` is stored with its full tag, `pt` never matched `pt-BR`. Additionally, quality parameters (e.g. `;q=0.9`) were not stripped prior to matching.

### Change
Updated `parseLanguageFromLocale` in `packages/lib/utils/i18n.ts` to:
1. Strip quality parameters (`;q=...`) from the locale string.
2. Check for an exact case-insensitive match against supported language codes (enabling `pt-BR` and `de` to match directly).
3. Fall back to matching the base language subtag against the supported codes' language subtags (allowing generic `pt` or `pt-PT` to resolve to `pt-BR`).

#### Test Red (Before Fix)
```
Buggy parse for 'pt-BR': null
Buggy parse for 'pt': null
```

#### Test Green (After Fix)
```
PS C:\Users\erijo\Downloads\claude> node --test test_reproduce_issue_3090.mjs
▶ Documenso Issue #3090: Brazilian Portuguese Locale Negotiation
  ✔ reproduces bug: pt-BR and pt headers fail to match pt-BR with buggy parser (1.9369ms)
  ✔ verified fix: pt-BR, pt, and quality factor headers correctly resolve to pt-BR (0.4368ms)
✔ Documenso Issue #3090: Brazilian Portuguese Locale Negotiation (3.8337ms)
ℹ tests 2
ℹ suites 1
ℹ pass 2
ℹ fail 0
```

### Reference
Fixes #3090

### DIFF
--- a/packages/lib/utils/i18n.ts
+++ b/packages/lib/utils/i18n.ts
@@ -21,17 +21,24 @@ export async function dynamicActivate(locale: string) {
 }
 
 const parseLanguageFromLocale = (locale: string): SupportedLanguageCodes | null => {
-  const [language, _country] = locale.split('-');
+  const cleaned = locale.split(';')[0].trim();
 
-  const foundSupportedLanguage = APP_I18N_OPTIONS.supportedLangs.find(
-    (lang): lang is SupportedLanguageCodes => lang === language,
+  const exactMatch = APP_I18N_OPTIONS.supportedLangs.find(
+    (lang): lang is SupportedLanguageCodes => lang.toLowerCase() === cleaned.toLowerCase(),
   );
 
-  if (!foundSupportedLanguage) {
-    return null;
+  if (exactMatch) {
+    return exactMatch;
   }
 
-  return foundSupportedLanguage;
+  const [language] = cleaned.split('-');
+
+  const languageMatch = APP_I18N_OPTIONS.supportedLangs.find(
+    (lang): lang is SupportedLanguageCodes =>
+      lang.split('-')[0].toLowerCase() === language.toLowerCase(),
+  );
+
+  return languageMatch ?? null;
 };
 
 /**


Mythie:

https://documen.so/cla

PR #3263, opened 2026-08-19
feat(webhooks): attach stable deliveryId and header for retry deduplication (#3204)

### LINKED ISSUE #3204: Webhook payloads carry no delivery id, so documented retries cannot be deduplicated

Rodrigoue9:
# Title
feat(webhooks): attach stable deliveryId and header for retry deduplication (#3204)

## Description
- Adds `deliveryId` and stable `createdAt` attributes to `TExecuteWebhookJobDefinition`.
- Embeds `deliveryId` in the webhook JSON payload and passes `X-Documenso-Delivery-Id` in the HTTP request headers.
- Allows webhook consumers to reliably deduplicate retry attempts across network reconnects and server failures.

Closes #3204

### DIFF
--- a/packages/lib/jobs/definitions/internal/execute-webhook.handler.ts
+++ b/packages/lib/jobs/definitions/internal/execute-webhook.handler.ts
@@ -17,14 +17,18 @@ export const run = async ({ payload, io: _io }: { payload: TExecuteWebhookJobDef
 
   const { webhookUrl: url, secret } = webhook;
 
+  const deliveryId = payload.deliveryId ?? crypto.randomUUID();
+  const createdAt = payload.createdAt ?? new Date().toISOString();
+
   const payloadData = {
     event,
+    deliveryId,
     payload: data,
-    createdAt: new Date().toISOString(),
+    createdAt,
     webhookEndpoint: url,
   };
 
-  const result = await executeWebhookCall({ url, body: payloadData, secret });
+  const result = await executeWebhookCall({ url, body: payloadData, secret, deliveryId });
 
   await prisma.webhookCall.create({
     data: {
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

PR #3271, opened 2026-08-19
fix(pdf): resolve inherited MediaBox and CropBox (#3267)

### LINKED ISSUE #3267: [Bug]: Inherited /MediaBox ignored — page size falls back to US Letter, misplacing placeholder fields on A4
## Summary

Documenso does not resolve an **inherited `/MediaBox`**. When a PDF declares its page size only on the `/Pages` node — which is valid per PDF 32000-1:2008 §7.7.3.4, where `MediaBox` is an inheritable page attribute — the page size falls back to US Letter. On A4 documents every placeholder-resolved field is then misplaced, and the offset grows with the distance from the page origin.

This affects `POST /envelope/field/create-many` with `placeholder`, and by the same mechanism anything else that maps coordinates onto the page.

## Evidence

Same page content, same placeholders, only the location of the `/MediaBox` differs:

| `/MediaBox` location | resolved column positions (mm) | row spacing (mm) |
|---|---|---|
| on `/Pages` only (inherited) | 15.6 · 54.5 · 93.4 · 132.3 · 166.3 | 26.5 |
| on each `/Page` (explicit) | 16.0 · 56.0 · 96.0 · 136.0 · 171.0 | 24.9 |
| **expected** | **16.0 · 56.0 · 96.0 · 136.0 · 171.0** | **24.9** |

The deviation in the inherited case is exactly the A4/Letter ratio:

```
x: measured 0.97250 … 0.97321   A4/Letter width  210 / 215.9 = 0.97267
y: measured 1.06292             A4/Letter height 297 / 279.4 = 1.06299
```

The source file is unambiguous — `pdfinfo` reports `Page size: 595.28 x 841.89 pts (A4)`, and the box is present, just one level up:

```
1 0 obj  /Type /Pages   /MediaBox [0 0 595.28 841.89]   ← inherited from here
3 0 obj  /Type /Page    (no /MediaBox)
```

## Steps to reproduce

1. Generate an A4 PDF whose `/MediaBox` sits only on the `/Pages` node — fpdf2 does this by default when all pages share one size; other producers do the same.
2. Place `{{signature}}` at known positions, e.g. a grid at x = 15/55/95/135/170 mm, y = 20/45/…/195 mm.
3. `POST /document/create` (multipart) with one `SIGNER` recipient.
4. `POST /envelope/field/create-many`:
   ```json
   { "envelopeId": "…", "data": [
       { "recipientId": 1, "envelopeItemId": "…", "type": "SIGNATURE",
         "placeholder": "{{signature}}", "matchAll": true } ] }
   ```
5. Compare the resulting `positionX` / `positionY` with the placeholder positions from `pdftotext -bbox`.
6. Repeat with the same PDF after copying the `/MediaBox` onto each page — the offsets disappear.

## Side effect: fields are silently dropped

Of 50 placeholders in the grid, only **43** produced fields in the inherited case. The missing ones are the lowest rows: after the 1.063 stretch they exceed 279.4 mm and fall outside the assumed page. The API returns success, the fields are simply absent — so a document can be distributed with fewer signature fields than intended.

## Environment

- Documenso v2.16.0, self-hosted, `NEXT_PUBLIC_UPLOAD_TRANSPORT=database`
- Documents created via `POST /api/v2/document/create`, fields via `POST /api/v2/envelope/field/create-many`
- Source PDFs produced by fpdf2 2.8.x, A4, embedded TrueType subset

## Expected behaviour

When a page has no own `/MediaBox`, walk up the page tree and use the inherited one, as required by the specification. Falling back to US Letter silently produces wrong geometry for every A4 document, which is the default format outside North America.

## Note

Explicit coordinates (`page` / `positionX` / `positionY` / `width` / `height`) are placed correctly on the same file, so the regression is confined to the path that derives geometry from the page size.

## Workaround for producers

Write the `/MediaBox` into every `/Page` object instead of relying on inheritance. In fpdf2 this can be done before serialising:

```python
from fpdf.output import _dimensions_to_mediabox

for page in pdf.pages.values():
    if not page.media_box:
        page.media_box = _dimensions_to_mediabox(page.dimensions())
```

Rodrigoue9:
# Title
fix(pdf): resolve inherited MediaBox and CropBox from parent page tree (#3267)

## Description
- Updates `getPageSize` to walk up parent `/Pages` nodes when a `/Page` node lacks its own explicit `/MediaBox` or `/CropBox`.
- Complies with ISO 32000-1 §7.7.3.4 (inheritable page attributes), preventing A4 documents from erroneously falling back to US Letter and misplacing placeholder fields.

Closes #3267

### DIFF
--- a/packages/lib/server-only/pdf/get-page-size.ts
+++ b/packages/lib/server-only/pdf/get-page-size.ts
@@ -18,12 +18,34 @@ export const getPageSize = (page: PDFPage) => {
 
   try {
     mediaBox = page.getMediaBox();
+    if (!mediaBox) {
+      let parent = (page.node as any).Parent?.();
+      while (parent && !mediaBox) {
+        const parentMediaBox = parent.MediaBox?.();
+        if (parentMediaBox) {
+          mediaBox = parentMediaBox.asRectangle?.() ?? parentMediaBox;
+          break;
+        }
+        parent = parent.Parent?.();
+      }
+    }
   } catch {
     // MediaBox lookup can fail for malformed PDFs where the entry is not a valid PDFArray.
   }
 
   try {
     cropBox = page.getCropBox();
+    if (!cropBox) {
+      let parent = (page.node as any).Parent?.();
+      while (parent && !cropBox) {
+        const parentCropBox = parent.CropBox?.();
+        if (parentCropBox) {
+          cropBox = parentCropBox.asRectangle?.() ?? parentCropBox;
+          break;
+        }
+        parent = parent.Parent?.();
+      }
+    }
   } catch {
     // CropBox lookup can fail for malformed PDFs where the entry is not a valid PDFArray.
   }


Mythie:

https://documen.so/cla

PR #3277, opened 2026-08-20
feat: allow admin recipient search

dguyen:
## Description

- Allow searching documents via recipient attributes in admin panel
- Add "View all results" in the admin command bar for recipients


roshboi (approved):
lgtm

PR #3278, opened 2026-08-20
fix(fields): resolve inherited /MediaBox for placeholder coordinates

### LINKED ISSUE #3267: [Bug]: Inherited /MediaBox ignored — page size falls back to US Letter, misplacing placeholder fields on A4

yzxcj797:
Fixes #3267.

## Root cause — verified down to the library line

The placeholder flow divides match coordinates by `@libpdf/core`'s `page.width`/`page.height`. In `@libpdf/core`'s `pdf-page.ts`, `getBox()` reads **the page's own dictionary entry only** — the inheritance its `getMediaBox()` docstring promises ("accounting for inheritance from parent pages") is not implemented — and a missing MediaBox returns the **US Letter default** (`{ 612, 792 }`). So a MediaBox declared only on the `/Pages` node (legal per PDF 32000-1:2008 §7.7.3.4; fpdf2's default for same-size pages) silently measures every page as Letter:

```
x: 210/215.9 = 0.97267  ← the issue's measured 0.97250…0.97321
y: 297/279.4 = 1.06299  ← the issue's measured 1.06292
```

I reproduced that ratio exactly from the math before touching any code.

## Fix

Resolve the effective page size through **pdf-lib**, whose MediaBox lookup walks `/Parent` (verified in `@cantoo/pdf-lib`'s `PDFPageLeaf.MediaBox()` → `getInheritableAttribute`):

- `resolveEffectivePageSizes(bytes)` — one pdf-lib parse per envelope item, mapping each page through the existing `getPageSize` helper (CropBox/MediaBox semantics preserved).
- `pickPageSize(resolved, libpdfSize)` — prefer the resolved size **only when it differs** from the text layer's own value, so rotated pages (where `@libpdf/core` swaps width/height) and explicit-MediaBox documents keep exactly their current path.
- `mapBoundingBoxToFieldPosition(bbox, pageSize, fieldSize)` — the point→percent math moved unchanged into a pure function, making the A4/Letter regression unit-testable without the envelope/prisma harness.
- Resolution is advisory: a PDF pdf-lib can't parse for sizing falls back to the text-layer dimensions (pre-fix behavior) rather than failing the field creation.

## Tests (`placeholder-page-size.test.ts`)

- **Inherited-MediaBox fixture built in-test**: a pdf-lib page whose MediaBox is moved up to the `/Pages` node (deleting the per-page entry) — resolves as A4, not Letter. **This is the document class from the issue.**
- Explicit per-page MediaBox resolves identically (no behavior change for the common case).
- `mapBoundingBoxToFieldPosition`: 16mm on A4 → 7.619% (the issue's expected value); a Letter regression guard pins the historical math; explicit field width/height overrides bbox-derived percentages.
- `pickPageSize`: prefers the resolved size on mismatch, keeps the text-layer size on agreement (rotation) and on missing resolution.

(The silent field-drop side effect — placeholders pushed past 279.4mm by the stretch — disappears with the size fix: A4 coordinates no longer exceed the A4 page.)

Upstream note: the inheritance walk belongs in `@libpdf/core`'s `getBox()`; this PR is the defensive fix for documents already in the wild and can slim down once the upstream lands.

### DIFF
--- a/packages/lib/server-only/field/create-envelope-fields.ts
+++ b/packages/lib/server-only/field/create-envelope-fields.ts
@@ -15,6 +15,11 @@ import { canRecipientFieldsBeModified } from '../../utils/recipients';
 import { assertEnvelopeMutable } from '../envelope/assert-envelope-mutable';
 import { getEnvelopeWhereInput } from '../envelope/get-envelope-by-id';
 import { type BoundingBox, whiteoutRegions } from '../pdf/auto-place-fields';
+import {
+  mapBoundingBoxToFieldPosition,
+  pickPageSize,
+  resolveEffectivePageSizes,
+} from './placeholder-page-size';
 
 type CoordinatePosition = {
   page: number;
@@ -119,6 +124,10 @@ export const createEnvelopeFields = async ({
     over resolved placeholders before saving back.
   */
   const pdfCache = new Map<string, PDF>();
+  // Inheritance-resolved page sizes (see resolveEffectivePageSizes): the
+  // percentage math below divides by the size the viewer shows, not the
+  // US-Letter fallback a /Pages-declared MediaBox reads as (#3267).
+  const pageSizeCache = new Map<string, Awaited<ReturnType<typeof resolveEffectivePageSizes>>>();
 
   if (hasPlaceholderFields) {
     for (const item of envelope.envelopeItems) {
@@ -126,6 +135,13 @@ export const createEnvelopeFields = async ({
       const pdfDoc = await PDF.load(new Uint8Array(bytes));
 
       pdfCache.set(item.id, pdfDoc);
+
+      try {
+        pageSizeCache.set(item.id, await resolveEffectivePageSizes(new Uint8Array(bytes)));
+      } catch {
+        // Size resolution is advisory: a PDF pdf-lib cannot parse for sizing
+        // keeps @libpdf/core's page dimensions, the pre-fix behavior.
+      }
     }
   }
 
@@ -189,9 +205,14 @@ export const createEnvelopeFields = async ({
 
       const matchesToProcess = field.matchAll ? matches : [matches[0]];
       const pages = pdfDoc.getPages();
+      const effectiveSizes = pageSizeCache.get(envelopeItemId);
 
       return matchesToProcess.map((match) => {
         const page = pages[match.pageIndex];
+        const pageSize = pickPageSize(effectiveSizes?.[match.pageIndex], {
+          width: page.width,
+          height: page.height,
+        });
 
         /*
           Record this placeholder's bounding box for whiteout. The bbox is in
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

PR #3284, opened 2026-08-20
fix(uploads): make the S3 checksum mode configurable for Object Lock buckets

### LINKED ISSUE #3282: Sealing fails against S3 buckets with Object Lock enabled — requestChecksumCalculation is hardcoded to WHEN_REQUIRED
### Issue Description

When the upload bucket has S3 Object Lock enabled, document sealing fails and documents remain stuck in PENDING after all recipients have signed.

The seal job retries three times and then gives up with:

[JOBS:task-...] Task failed InvalidRequest: Content-MD5 OR x-amz-checksum- HTTP header is required
for Put Object requests with Object Lock parameters
    at AwsRestXmlProtocol.handleError (/node_modules/@aws-sdk/core/dist-cjs/submodules/protocols/index.js)
[JOBS]: Job internal.seal-document failed BackgroundTaskExceededRetriesError: Task exceeded retries

S3 requires a Content-MD5 or x-amz-checksum-* header on any PutObject request that carries Object Lock parameters. packages/lib/universal/upload/providers/s3-provider.ts sets:

ts
requestChecksumCalculation: 'WHEN_REQUIRED',
responseChecksumValidation: 'WHEN_REQUIRED',

The in-code comment explains why — S3-compatible providers such as GarageHQ, MinIO and Backblaze B2 reject the AWS SDK's default CRC32 checksums with InvalidDigest. That is a reasonable default, but it is not overridable: because the value is set explicitly on the client, the standard AWS_REQUEST_CHECKSUM_CALCULATION environment variable has no effect.

Uploads are unaffected because they use a presigned browser POST, which carries no Object Lock parameters. Only the server-side seal write fails.

### Steps to Reproduce

Configure NEXT_PUBLIC_UPLOAD_TRANSPORT=s3 against a real AWS S3 bucket
Enable Object Lock on that bucket with a default retention period
Create a document, send it, and have all recipients sign
Sealing fails; the document stays PENDING and appears under Admin → Unsealed Documents

### Expected Behavior

Sealing succeeds against buckets with Object Lock enabled, or the checksum behaviour is configurable so operators who need it can opt in without patching.

### Current Behavior

Make it configurable while keeping the current default, so third-party S3-compatible backends are unaffected:

requestChecksumCalculation:
  env('NEXT_PRIVATE_UPLOAD_CHECKSUM_CALCULATION') === 'WHEN_SUPPORTED'
    ? 'WHEN_SUPPORTED'
    : 'WHEN_REQUIRED',

The ternary preserves the literal union type so tsc stays happy — reading the env value directly widens it to string and fails typecheck.

Verified against v2.16.0: with NEXT_PRIVATE_UPLOAD_CHECKSUM_CALCULATION=WHEN_SUPPORTED, previously stuck documents seal successfully and the completion email is delivered with the signed PDF attached.

Happy to open a PR if this approach looks right.

### Screenshots (optional)

_No response_

Ubuntu 24.04.3 LTS

Chrome

### Version [e.g., 2.13.0]

2.16.0

yzxcj797:
Fixes #3282 — the reporter's exact proposed change (verified against v2.16.0 in the issue), plus the compose passthrough so Docker deployments can actually set it.

## Root cause (as reported)

S3 **requires** a `Content-MD5` or `x-amz-checksum-*` header on any PutObject request carrying Object Lock parameters. The provider pins `requestChecksumCalculation: 'WHEN_REQUIRED'` — the correct default for the S3-compatible backends the comment names (GarageHQ/MinIO/B2 reject the SDK's default CRC32 with `InvalidDigest`) — but pinning the value explicitly **also disables the SDK's standard `AWS_REQUEST_CHECKSUM_CALCULATION` env var**, leaving Object Lock operators no way to opt in. The server-side seal write is the only affected call (browser uploads ride a presigned POST with no Object Lock parameters), so documents seal-fail and stay PENDING after every recipient signed.

## The change

```ts
requestChecksumCalculation:
  env('NEXT_PRIVATE_UPLOAD_CHECKSUM_CALCULATION') === 'WHEN_SUPPORTED'
    ? 'WHEN_SUPPORTED'
    : 'WHEN_REQUIRED',
```

(the reporter's ternary verbatim — it preserves the literal union type; reading the env value directly widens to `string` and fails typecheck), applied to both the request and response modes. **The default is unchanged** — third-party backends are unaffected unless an operator explicitly opts in. `docker/production/compose.yml` passes `NEXT_PRIVATE_UPLOAD_CHECKSUM_CALCULATION` through so the documented Docker deployment path can set it.

## Test

`s3-provider.test.ts` (new) — constructs the provider with the env var unset (default path) and set to `WHEN_SUPPORTED`, asserting the var is consulted on both paths. Pre-fix the var was a hardcoded literal, never read. (A prisma-free mock keeps the module importable without a generated client.)


Mythie:

https://documen.so/cla

Issue #3191, opened 2026-08-13
Rejecting an AES/QES envelope is a dead end: status never becomes REJECTED and cannot be recovered


bastien70:
### Issue Description

`DocumentStatus.REJECTED` is written in exactly one place in production code: `seal-document.handler.ts` (`status: finalEnvelopeStatus`, ~309). But at 169-181, **before** that write:

```ts
const finalEnvelopeStatus = isRejected ? DocumentStatus.REJECTED : DocumentStatus.COMPLETED;

if (isTspEnvelope(envelope)) {
  if (isResealing) { throw new AppError(AppErrorCode.NOT_SETUP, {...}); }
  if (isRejected) {
    throw new AppError(AppErrorCode.NOT_SETUP, {
      message: 'TSP envelope rejection is not supported in V1 — rejection stamps would invalidate PAdES signatures.',
    });
  }
```

Rejection commits the recipient state first (`reject-document-with-token.ts:57-82`: `signingStatus: REJECTED` plus audit entry), **then** triggers the job (92), which throws. The result is a recipient marked `REJECTED`, cancellation emails already sent to everyone, and an envelope **frozen in `PENDING` indefinitely**.

**No guard upstream of it:** neither `reject-document-with-token.ts` nor `reject-document-on-behalf-of.ts` (API v2) checks `isTspEnvelope`. The asymmetry is the tell — the *completion* path does branch on TSP before any side effect (`recipient-router/router.ts:604-617`); the *rejection* path (664) does not.

**No recovery:** "Reseal" from admin re-enters the same handler, and since the envelope is still `PENDING`, `isResealing` is false and it lands on the same throw. Other recipients keep signing a document that can never conclude. The only way out is cancelling the document, which loses the rejection semantics.

**Reachability:** CSC mode is a plain env var — `IS_INSTANCE_CSC_MODE()` = `NEXT_PRIVATE_SIGNING_TRANSPORT === 'csc'` (`constants/app.ts:103-109`), no runtime licence check. On such an instance, envelopes are AES/QES by default. There is no test coverage: `rg 'QES|IS_INSTANCE_CSC' packages/app-tests/` returns nothing.

### Steps to Reproduce

1. Run an instance with `NEXT_PRIVATE_SIGNING_TRANSPORT=csc`.
2. Send a document to two signers.
3. Have one of them reject it.
4. Check the envelope status, and the recipients' view.

### Expected Behavior

[...]

Issue #3092, opened 2026-07-13
Sealing fails with PlaceholderError when a timestamp authority is configured (B-LTA signature exceeds @libpdf default estimatedSize of 12288 bytes)


fabrunrun:
### Issue Description

On a self-hosted instance, setting NEXT_PRIVATE_SIGNING_TIMESTAMP_AUTHORITY causes the seal-document job to fail with a PlaceholderError. As a result, documents are never sealed and the envelope stays stuck in PENDING even though all recipients have signed (this is very likely the root cause behind #2215).

Without a TSA, sealing works fine. The moment a TSA is set, signPdf() requests a B-LTA signature (timestamp + LTV + archival document timestamp) whose CMS structure is larger than the default signature placeholder reserved by @libpdf, and signing throws.

Environment:
- Documenso v2.14.0 (self-hosted, Docker)
- @libpdf/core 0.4.0 (bundled)
- Signing certificate: Certigna Identity CA (qualified EU/eIDAS CA, RSA) - full chain leaf -> intermediate -> root
- Timestamp authority: DigiCert http://timestamp.digicert.com, RSA-4096 timestamp responder
- NEXT_PUBLIC_UPLOAD_TRANSPORT=s3
Error:
```
PlaceholderError: Signature placeholder too small: need 12465 bytes, have 12288 bytes
```

Root cause:
signPdf() in packages/signing calls pdf.sign() with longTermValidation and archivalTimestamp enabled when a TSA is present, but never passes estimatedSize, so @libpdf falls back to its default of 12288 bytes (12 KB):

```js
// packages/signing/index.ts (and the bundled seal-document handler)
const { bytes } = await pdf.sign({
  signer,
  reason: 'Signed by Documenso',
  location: NEXT_PUBLIC_WEBAPP_URL(),
  contactInfo: NEXT_PUBLIC_SIGNING_CONTACT_INFO(),
  subFilter: NEXT_PRIVATE_USE_LEGACY_SIGNING_SUBFILTER() ? 'adbe.pkcs7.detached' : 'ETSI.CAdES.detached',
  timestampAuthority: tsa ?? undefined,
  longTermValidation: !!tsa,
  archivalTimestamp: !!tsa,
  // no estimatedSize -> @libpdf default 12288
});
```

[...]

Issue #2921, opened 2026-06-03
Bug: Document remains in "Pending" state when TSA is enabled behind an HTTP proxy


nobrother:
## Description

When timestamping is enabled via `NEXT_PRIVATE_SIGNING_TIMESTAMP_AUTHORITY`, document signing never completes in an environment that requires outbound internet access through an HTTP proxy.

The final signer can successfully sign the document, but the document remains in the **Pending** state indefinitely and the signed PDF is never generated.

No error messages are visible in the application logs.

## Environment

* Documenso version: v2.9.1
* Deployment method: Docker Compose
* Node.js version: 22.22.2
* Database: PostgreSQL
* Storage: Database storage (not MinIO/S3)
* Outbound internet access: HTTP proxy required

Proxy configuration:

```env
NODE_USE_ENV_PROXY=1
HTTP_PROXY=http://proxy:3128
HTTPS_PROXY=http://proxy:3128
http_proxy=http://proxy:3128
https_proxy=http://proxy:3128
```

TSA configuration:

```env
NEXT_PRIVATE_SIGNING_TIMESTAMP_AUTHORITY=https://timestamp.sectigo.com
```

## Steps to Reproduce

1. Deploy Documenso behind an HTTP proxy.
2. Configure the proxy environment variables shown above.
3. Configure `NEXT_PRIVATE_SIGNING_TIMESTAMP_AUTHORITY`.
4. Create a document and send it for signing.
5. Complete all signatures, including the final signer.

## Expected Behavior

After the final signer signs the document:

* The document should be finalized.
* The signed PDF should be generated.
* The document status should change from Pending to Completed.

## Actual Behavior

After the final signer signs the document:

* The signing UI completes successfully.
* The document remains in Pending status indefinitely.
* No signed PDF is generated.
* No relevant errors appear in the application logs.

## Additional Findings

The issue appears to be specifically related to TSA processing.

If `NEXT_PRIVATE_SIGNING_TIMESTAMP_AUTHORITY` is removed:

* Document signing completes successfully.
* The signed PDF is generated.
* The document status changes to Completed.

Proxy connectivity itself appears to be working correctly. From inside the Documenso container:

```bash
node -e "fetch('https://timestamp.sectigo.com').then(r=>console.log(r.status)).catch(console.error)"
```

returns:

```text
200
```

This suggests that general outbound HTTPS connectivity through the proxy is functional.

[...]

PR #3288, opened 2026-08-20
fix(envelope): flip the envelope to REJECTED in the rejection transaction

### LINKED ISSUE #3287: After a 200 from POST /api/v2/envelope/recipient/{id}/reject the envelope is still PENDING, and other recipients can still act on it
### Issue Description

The lifecycle page says a rejection moves the document to Rejected immediately and that other pending recipients can no longer act on it. Neither holds at the moment the rejection returns. The rejection handler commits the recipient row and an audit log and nothing else; the document-level flip to `REJECTED` is written only by the asynchronous `internal.seal-document` job, after the whole PDF pipeline has run. The very next `GET /api/v2/envelope/{id}` still reports `"status": "PENDING"`.

The stale read is the visible half. The harm is the other half: every guard that stops the rest of the signing flow keys off `envelope.status === PENDING`. `sign-field-with-token` and `complete-document-with-token` check only that plus the acting recipient's own signing status, so for the whole window the other recipients can still sign or reject a document the product has already cancelled by email, since `send.document.cancelled.emails` is queued by the rejection request itself. If the seal job fails, and it demonstrably does (see #2921, #3092, #3191), the window never closes: the sweep job retries only between 15 minutes and 6 hours after the last recipient action and gives up after that.

#### Documentation

https://docs.documenso.com/docs/concepts/document-lifecycle, "Rejected":

> If you enable document rejection in settings, recipients can reject instead of signing. When any recipient rejects:
>
> * The document immediately moves to **Rejected** state
> * Other pending recipients can no longer act on the document
> * The document owner is notified

#### Cause

`rejectDocumentOnBehalfOf` commits exactly two rows in its transaction, the recipient (`signedAt`, `signingStatus: REJECTED`, `rejectionReason`) and a `DOCUMENT_RECIPIENT_REJECTED` audit log, then fires three jobs and returns: https://github.com/documenso/documenso/blob/3cf2963cd03d8b24770b7490bdb20e596baa5d65/packages/lib/server-only/document/reject-document-on-behalf-of.ts#L115-L177. It never touches `Envelope.status`; the only `DocumentStatus` reference in the file is the precondition `if (envelope.status !== DocumentStatus.PENDING)`.

The single writer of `DocumentStatus.REJECTED` in production code is `packages/lib/jobs/definitions/internal/seal-document.handler.ts`, which computes `finalEnvelopeStatus = isRejected ? REJECTED : COMPLETED` and writes it only after loading every envelope item's PDF, stamping the rejection, generating the certificate and audit-log PDFs, signing and uploading. With the default local job provider that job is dispatched over HTTP and explicitly not awaited (`packages/lib/jobs/client/local.ts` races the fetch against a 150 ms timeout), so the mutation returns long before the status changes.

The recipient-token path `reject-document-with-token.ts` has the identical shape, so this is the design of rejection rather than an API-only slip. The public API surface came with #3007 (2026-06-22, "feat: add API endpoint to reject documents on behalf of recipients"); the async design dates to #1472 (2024-11-14, "feat: signature rejection").

Still present at today's head, `75330166cc` (v2.17.0): https://github.com/documenso/documenso/blob/75330166cc00b29c14399bc2e391e4b4d8080c00/packages/lib/server-only/document/reject-document-on-behalf-of.ts#L115-L177. Both culprit files and the doc page are byte-identical to the tested v2.16.0.

Fix shape: write `Envelope.status = REJECTED` and the `DOCUMENT_REJECTED` webhook inside the same transaction that flips the recipient, and let the seal job produce only the stamped PDF. That closes the read-after-write gap, makes the "other recipients can no longer act" guard effective immediately, and removes the dependency of a terminal legal state on a PDF pipeline that can fail. If the asynchronous design is intended instead, the docs should say the status settles after sealing and point integrators at the `DOCUMENT_REJECTED` webhook, the way `developers/examples/common-workflows` already polls for a terminal status.

One thing that already works correctly and is worth documenting either way: recipient-level truth is immediate. The same GET payload already shows the rejecting recipient as `REJECTED` with the reason, so an integration can read the rejection off the recipient array before the envelope agrees.

#### Suggested labels

`type: bug`

#### Related

- #3191, "Rejecting an AES/QES envelope is a dead end: status never becomes REJECTED and cannot be recovered" (open since 2026-08-13, no comments): names the same single-writer root cause, but is scoped to the TSP/CSC case where the seal handler throws so the flip never happens at all. The ordinary non-TSP lag reported here is not covered by it.
- #3180, "Recipients who already signed can still reject, destroying their own signature" (open): a second missing guard on the same rejection path.
- #2921 and #3092 (open): two ways the seal job dies, leaving the envelope in PENDING forever.
- #2215 (closed): the completion twin, fixed with the sweep rather than a synchronous status write.
- #2563 (merged 2026-03-05): the seal-document sweep, the project's own mitigation for this class of stall; its query explicitly matches "any recipient has REJECTED". It runs 15 minutes to 6 hours after the last recipient action, so it cannot make "immediately" true.
- #3234 (closed unmerged, 2026-08-18): would have made the sweep cadence configurable, so that window is still fixed.

### Steps to Reproduce

Build tested: v2.16.0 (tag `v2.16.0`, `3cf2963cd03d8b24770b7490bdb20e596baa5d65`), built and run from source: `npm ci`, `npx turbo run build --filter=@documenso/remix`, `npm run start -w @documenso/remix`, against Postgres 17, with `NEXT_PRIVATE_JOBS_PROVIDER=local`, `NEXT_PUBLIC_UPLOAD_TRANSPORT=database` and `NEXT_PRIVATE_SIGNING_TRANSPORT=local`. Every call carries a team API token: `Authorization: <token>`. Ids below are from the recorded run.

1. Create an envelope from a template: `POST /api/v2/envelope/use`, `Content-Type: multipart/form-data; boundary=tcguard`, payload part `{"envelopeId":"<template envelope id>","externalId":"tcref-reject-b99cc338a5"}`. 200, `{"id":"envelope_dswotmftyonlisrm","recipients":[]}`.

2. Add two signers, so the effect on a bystander is observable:

   ```
   {"envelopeId":"envelope_dswotmftyonlisrm","data":[{"email":"tcref-declines-b99cc338a5@documenso.test","name":"TCRef Decliner","role":"SIGNER"},{"email":"tcref-other-b99cc338a5@documenso.test","name":"TCRef Other","role":"SIGNER"}]}
   ```

   200, recipients `360` and `361`.

3. Give each a signature field:

   ```
   {"envelopeId":"envelope_dswotmftyonlisrm","data":[{"type":"SIGNATURE","recipientId":"360","page":1,"positionX":10,"positionY":70,"width":20,"height":5},{"type":"SIGNATURE","recipientId":"361","page":1,"positionX":10,"positionY":80,"width":20,"height":5}]}
   ```

   200.

4. Send it: `POST /api/v2/envelope/distribute` with `{"envelopeId":"envelope_dswotmftyonlisrm","meta":{"distributionMethod":"NONE"}}`. 200. (`NONE` only removes the need for a mail server.)

5. Reject on the first recipient's behalf:

   ```
   POST /api/v2/envelope/recipient/360/reject
   {"envelopeId":"envelope_dswotmftyonlisrm","recipientId":"360","reason":"tcref declined b99cc338a5"}
   ```

   200, recipient 360 comes back `"signingStatus":"REJECTED"`, `"rejectionReason":"tcref declined b99cc338a5"`, `"signedAt":"2026-08-14T20:05:31.134Z"`. The server log shows the request firing `internal.seal-document`, `send.signing.rejected.emails` and `send.document.cancelled.emails`.

6. Read the envelope back immediately: `GET /api/v2/envelope/envelope_dswotmftyonlisrm`. 200.

Re-tested live on v2.17.0 (75330166cc, 2026-08-19): still reproduces. Reject returned 200 with signingStatus REJECTED on the recipient, but the envelope still read PENDING at 12 ms and at 112 ms after that 200, flipping to REJECTED at 52 ms in one run and 638 ms in another.

### Expected Behavior

Step 6 reports `"status": "REJECTED"`, and from step 5 onward recipient 361 can no longer sign or reject.

### Current Behavior

Step 6 returns `"status": "PENDING"` with `completedAt: null`, while the embedded recipients already read `360: REJECTED` and `361: NOT_SIGNED`:

```json
{"internalVersion":2,"type":"DOCUMENT","status":"PENDING","id":"envelope_dswotmftyonlisrm","secondaryId":"document_367","externalId":"tcref-reject-b99cc338a5","completedAt":null,"documentMeta":{"signingOrder":"PARALLEL","distributionMethod":"NONE", ...},"recipients":[{"role":"SIGNER","signingStatus":"REJECTED","id":360,"email":"tcref-declines-b99cc338a5@documenso.test","signedAt":"2026-08-14T20:05:31.134Z","rejectionReason":"tcref declined b99cc338a5", ...},{"role":"SIGNER","signingStatus":"NOT_SIGNED","id":361,"email":"tcref-other-b99cc338a5@documenso.test","signedAt":null,"rejectionReason":null, ...}], ...}
```

Step 5 took 299 ms and step 6 took 205 ms, so the read happened roughly 200 to 400 ms after the rejection. The cancellation emails to the owner and to recipient 361 had already been queued at that point.

Scope of the evidence: the transcript establishes only that the flip had not happened within that window; that the write is asynchronous, and therefore that the window has no documented bound, comes from the source, cited above. The recorded run stopped at this step, so the two follow-on assertions in the same flow (a rejected envelope refusing an update, and the second recipient being unable to reject) were never executed and carry no evidence either way.

n/a (API)

### Version [e.g., 2.13.0]

2.16.0 (tested; tag `v2.16.0` = `3cf2963cd03d8b24770b7490bdb20e596baa5d65`), re-checked in source against 2.17.0 (`75330166cc00b29c14399bc2e391e4b4d8080c00`), where the culprit files and the doc page are byte-identical.

yzxcj797:
Fixes #3287 — the issue's "Fix shape" implemented: the status flip moves into the rejection's own transaction; the seal job keeps only the PDF work.

## Root cause (as diagnosed, verified)

`rejectDocumentOnBehalfOf` commits exactly two rows — the recipient (`signedAt`, `REJECTED`, `rejectionReason`) and a `DOCUMENT_RECIPIENT_REJECTED` audit log — then fires three jobs and returns. **`Envelope.status` is never touched.** The only production writer of `DocumentStatus.REJECTED` is the `internal.seal-document` job, after loading every envelope item's PDF, stamping the rejection, generating the certificate and audit-log PDFs, signing and uploading — and the local job provider dispatches it un-awaited against a 150ms timeout. Consequences, exactly as the issue lays out:

1. **Stale read**: the very next `GET /api/v2/envelope/{id}` after a 200 still says `"status": "PENDING"`.
2. **Open action window**: every `status === PENDING` guard (`sign-field-with-token`, `complete-document-with-token`, further rejections) keeps admitting the OTHER recipients — on a document the product has already cancelled, since `send.document.cancelled.emails` is queued by the rejection request itself.
3. **Permanent window on seal failure**: when the seal job fails (#2921, #3092, #3191), the sweep retries only between 15min and 6h after the last recipient action and then gives up — a terminal legal state dependent on a PDF pipeline that can fail.

## The fix

Both rejection paths — the public API's `reject-document-on-behalf-of` and the recipient-token `reject-document-with-token` (the issue notes the identical shape) — now include:

```ts
prisma.envelope.update({
  where: { id: envelope.id },
  data: { status: DocumentStatus.REJECTED },
}),
```

**inside the same `$transaction`** as the recipient flip. The documented lifecycle ("the document immediately moves to Rejected; other pending recipients can no longer act; the owner is notified") becomes true at the moment the rejection lands, and the read-after-write gap closes.

The seal job is unchanged and remains the stamped-PDF/certificate producer; its own status write is idempotent (it writes `REJECTED` again — the pre-existing `finalEnvelopeStatus` computation and the documents it produces are untouched).

## Tests

Source-contract tests (`reject-document-on-behalf-of.test.ts`) pin the transaction composition of both handlers — the `prisma.envelope.update` + `status: DocumentStatus.REJECTED` pair present in the `$transaction` of each — and that the seal job still owns the stamped-PDF path. **Both handler tests fail on `main`** (no envelope write in either transaction); the seal-job contract passes on both sides (its behavior is deliberately unchanged).

### DIFF
--- a/packages/lib/server-only/document/reject-document-on-behalf-of.test.ts
+++ b/packages/lib/server-only/document/reject-document-on-behalf-of.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from 'vitest';
+
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+/**
+ * The rejection handlers must write `Envelope.status = REJECTED` inside the
+ * SAME `$transaction` that flips the recipient — previously only the async
+ * seal job wrote it, so a 200 from the reject endpoint left the envelope
+ * PENDING and the PENDING-keyed guards kept letting other recipients act
+ * (#3287).
+ *
+ * These are source-contract tests: the transaction composition is what the
+ * fix is, and it is directly inspectable in the handler source (the module
+ * needs a generated prisma client to import, which unit CI doesn't have).
+ */
+describe('rejection writes the envelope status in the same transaction (#3287)', () => {
+  const read = (p: string) =>
+    readFileSync(resolve(__dirname, p), 'utf-8');
+
+  it('the behalf-of path includes an envelope.update to REJECTED in the transaction', () => {
+    const src = read('./reject-document-on-behalf-of.ts');
+    const tx = src.slice(src.indexOf('$transaction'), src.indexOf(']);', src.indexOf('$transaction')));
+    expect(tx).toContain('prisma.envelope.update');
+    expect(tx).toContain('status: DocumentStatus.REJECTED');
+    // The envelope flip precedes the recipient flip (same transaction; order
+    // is not semantic but keeps the diff readable).
+    expect(tx.indexOf('prisma.envelope.update')).toBeLessThan(tx.indexOf('prisma.recipient.update'));
+  });
+
+  it('the recipient-token path includes the same flip', () => {
+    const src = read('./reject-document-with-token.ts');
+    const tx = src.slice(src.indexOf('$transaction'), src.indexOf(']);', src.indexOf('$transaction')));
+    expect(tx).toContain('prisma.envelope.update');
+    expect(tx).toContain('status: DocumentStatus.REJECTED');
+  });
+
+  it('the seal job remains the stamped-PDF producer (idempotent status write)', () => {
+    const src = read('../../jobs/definitions/internal/seal-document.handler.ts');
+    expect(src).toContain('finalEnvelopeStatus');
+    expect(src).toContain('DocumentStatus.REJECTED');
+  });
+});
--- a/packages/lib/server-only/document/reject-document-on-behalf-of.ts
+++ b/packages/lib/server-only/document/reject-document-on-behalf-of.ts
@@ -113,6 +113,25 @@ export async function rejectDocumentOnBehalfOf({
   // Update the recipient status to rejected and record an external rejection
   // audit log within the same transaction.
   const [updatedRecipient] = await prisma.$transaction([
+    // Flip the envelope to REJECTED in the SAME transaction as the recipient
+    // row. The status was previously written only by the async seal job —
+    // after the whole PDF pipeline — so a 200 from this endpoint left the
+    // envelope PENDING, and every `status === PENDING` guard
+    // (sign-field-with-token, complete-document-with-token) kept letting the
+    // OTHER recipients act on a document the product had already cancelled
+    // (the cancellation emails are queued by this very request). When the
+    // seal job fails — and it demonstrably does — the window never closed.
+    // The seal job's own status write is idempotent (it writes REJECTED
+    // again) and the stamped PDF/certificate reflect the rejection either
+    // way (#3287).
+    prisma.envelope.update({
+      where: {
+        id: envelope.id,
+      },
+      data: {
+        status: DocumentStatus.REJECTED,
+      },
+    }),
     prisma.recipient.update({
       where: {
         id: recipient.id,
--- a/packages/lib/server-only/document/reject-document-with-token.ts
+++ b/packages/lib/server-only/document/reject-document-with-token.ts
@@ -55,6 +55,17 @@ export async function rejectDocumentWithToken({ token, id, reason, requestMetada
 
   // Update the recipient status to rejected
   const [updatedRecipient] = await prisma.$transaction([
+    // Same-transaction envelope flip as the API path: the document is
+    // Rejected the moment the rejection lands, and the PENDING-keyed
+    // guards close immediately (#3287).
+    prisma.envelope.update({
+      where: {
+        id: envelope.id,
+      },
+      data: {
+        status: DocumentStatus.REJECTED,
+      },
+    }),
     prisma.recipient.update({
       where: {
         id: recipient.id,


Mythie:

https://documen.so/cla

PR #3136, opened 2026-07-30
fix(trpc): v2 field position updates silently dropped; rewrite fields API docs

ephraimduncan:
## Description

The Fields API guide shows requests that the API accepts. Coordinate update requests move fields. Previously, the API returned success but did not move the fields.

## Related Issue

No linked issue.

## Changes Made

The guide uses the envelope request fields and response wrapper from the active schemas. The v2 update route maps the public coordinate names to the service names.

The guide shows Decimal coordinates as JSON strings. The metadata and sample IDs match the API output.

## Testing Performed

- The docs MDX check and type generation completed successfully.
- The Remix TypeScript check completed successfully.
- Biome found no errors in the changed TypeScript file.
- The local browser test showed no visible layout errors in the field table and code examples.
- The runtime check showed `"positionX":"12.5"` for a Prisma Decimal value.

## Checklist

- [x] I did local tests of these changes.
- [x] I updated the API guide.
- [x] I obeyed the code style.
- [x] I corrected the posted review findings.

## Additional Notes

Legacy v1 and `/document/field/*` callers still pass the service coordinate names. The v2 route changes only the envelope API boundary.

PR #3289, opened 2026-08-20
fix(api): keep a field's stored fieldMeta when update-many omits it

### LINKED ISSUE #3286: POST /api/v2/envelope/field/update-many resets a field's fieldMeta to type defaults when fieldMeta is omitted
### Issue Description

The Fields API documents `update-many` as a partial update: change one property of one field in a single request. In practice, any `update-many` call that does not resend the complete `fieldMeta` overwrites the stored configuration with the type's default metadata, behind a 200 that echoes the wiped state as though it were the update result. A TEXT field loses its label, placeholder, required flag and character limit; a RADIO field's option list becomes a single blank option, a CHECKBOX's becomes one blank unchecked option, a DROPDOWN's becomes a single "Option 1". There is no way to avoid it from the client side: `type` is the schema's union discriminator and is therefore mandatory on an update, so the default is always materialised. The documented update example, `{ id, type, pageY }`, is exactly the call that triggers it.

#### Documentation

https://docs.documenso.com/docs/developers/api/fields, "Update Fields":

> Update Fields [#update-fields]
>
> Update one or more fields in a single request.
>
> `POST /envelope/field/update-many`

and the response the page shows for that partial update:

```json
{
  "fields": [
    { "id": 101, "type": "SIGNATURE", "positionY": 85 },
    { "id": 102, "type": "DATE", "positionY": 85 }
  ]
}
```

Nothing on the page says a partial update discards the rest of the field's configuration.

#### Cause

`ZEnvelopeFieldAndMetaSchema` gives every fieldMeta-bearing member a create-time default: https://github.com/documenso/documenso/blob/3cf2963cd03d8b24770b7490bdb20e596baa5d65/packages/lib/types/field-meta.ts#L403-L448. All ten members read `fieldMeta: Z<Type>FieldMeta.optional().default(FIELD_<TYPE>_META_DEFAULT_VALUES)` (SIGNATURE, INITIALS, NAME, EMAIL, DATE, TEXT, NUMBER, RADIO, CHECKBOX, DROPDOWN). The update route consumes that schema (`packages/trpc/server/envelope-router/envelope-fields/update-envelope-fields.types.ts:18`), so an omitted `fieldMeta` arrives at the service as the full default object rather than as `undefined`, and the service writes it unconditionally inside `tx.field.update`: https://github.com/documenso/documenso/blob/75330166cc00b29c14399bc2e391e4b4d8080c00/packages/lib/server-only/field/update-envelope-fields.ts#L136.

The defaults were introduced for create by #2181 (2025-11-12, "fix: add default values for envelope field meta"), which switched the update route's types from `ZFieldAndMetaSchema` to `ZEnvelopeFieldAndMetaSchema`; they are correct for create and wrong for update. The deprecated `/document/field/update-many` path is unaffected: its `ZUpdateFieldSchema` uses the plain `ZFieldAndMetaSchema` with no defaults.

Still present at today's head, `75330166cc` (v2.17.0): both `field-meta.ts` and `update-envelope-fields.ts` are byte-identical to the tested v2.16.0.

Fix shape: give the update route a schema whose `fieldMeta` is optional with no default, so an omitted `fieldMeta` leaves the column untouched, and leave create's defaults alone.

#### Suggested labels

`type: bug`

#### Related

- PR #3136 fixes the sibling defect on the same endpoint (the route validates `page` / `positionX` / `positionY` while the service reads `pageNumber` / `pageX` / `pageY`, so a move is a silent no-op). It covers the coordinate half only. Its current diff still forwards `fieldMeta: field.fieldMeta`, the zod-defaulted value, so with #3136 merged a move would work and the field's configuration would still be wiped. Merging it makes the endpoint look fixed while this defect stays.
- The audit log written by `diffFieldChanges` records the fieldMeta reset, so the loss is visible in the audit trail even though nothing surfaces it to the caller.

Found by TrueCourse running the published API docs against a live instance; the full transcript (requests, responses, server log) is available on request.

### Steps to Reproduce

1. Create an envelope from a template: `POST /api/v2/envelope/use`, `Content-Type: multipart/form-data; boundary=tcguard`, payload part `{"envelopeId":"<template envelope id>","externalId":"tcref-fieldmove-4850ccfd5b"}`. 200, `{"id":"envelope_yroalbrelnoaxrmm","recipients":[]}`.

2. Add a recipient:

   ```
   {"envelopeId":"envelope_yroalbrelnoaxrmm","data":[{"email":"tcref-4850ccfd5b@documenso.test","name":"TCRef Signer","role":"SIGNER"}]}
   ```

   200, recipient `id: 334`.

3. Create a configured TEXT field:

   ```
   {"envelopeId":"envelope_yroalbrelnoaxrmm","data":[{"type":"TEXT","recipientId":"334","page":1,"positionX":10,"positionY":20,"width":20,"height":5,"fieldMeta":{"type":"text","label":"Job Title","placeholder":"Enter your job title","required":true,"characterLimit":40,"textAlign":"left"}}]}
   ```

   200, field `id: 300`, and the response echoes the meta back: `"fieldMeta":{"label":"Job Title","placeholder":"Enter your job title","required":true,"type":"text","characterLimit":40,"textAlign":"left"}`.

4. Update one property, sending no `fieldMeta`:

   ```
   POST /api/v2/envelope/field/update-many
   {"envelopeId":"envelope_yroalbrelnoaxrmm","data":[{"id":"300","type":"TEXT","positionY":60}]}
   ```

   200. The stored `fieldMeta` is now the TEXT defaults.

Re-tested live on v2.17.0 (75330166cc, 2026-08-19): still reproduces, unchanged. The same 200 replaced {label:'Job Title', placeholder:'Enter your job title', required:true, characterLimit:40, textAlign:'left'} with the text-type defaults.

### Expected Behavior

An `update-many` call that does not carry `fieldMeta` leaves the field's stored `fieldMeta` untouched. After step 4 the field still reads `{"label":"Job Title","placeholder":"Enter your job title","required":true,"type":"text","characterLimit":40,"textAlign":"left"}`.

### Current Behavior

Step 4 returns HTTP 200 and the stored configuration is gone:

```json
{"data":[{"envelopeId":"envelope_yroalbrelnoaxrmm","envelopeItemId":"envelope_item_iuruhhafmskdumud","type":"TEXT","id":300,"secondaryId":"cmstdmf0l0007y6r79fbzxeee","recipientId":334,"page":1,"positionX":"10","positionY":"20","width":"20","height":"5","customText":"","inserted":false,"fieldMeta":{"label":"","placeholder":"","required":false,"readOnly":false,"fontSize":12,"type":"text","text":"","textAlign":"left"},"documentId":343,"templateId":null}]}
```

That `fieldMeta` is byte-for-byte `FIELD_TEXT_META_DEFAULT_VALUES` (`packages/lib/types/field-meta.ts:302-312`). The response is built from the row after `prisma.field.update`, so it is what is persisted. The server log shows the `envelope.field.updateMany` call completing normally; stderr is empty.

The blast radius is all ten fieldMeta-bearing types. RADIO drops to `values: [{id: 1, checked: false, value: ''}]`, CHECKBOX to one blank unchecked option, DROPDOWN to `values: [{value: 'Option 1'}]` with an empty default, TEXT and NUMBER lose label, placeholder, default text or value, character limit and min/max. Every SDK generated from the v2 OpenAPI document inherits it.

The same response also shows the coordinate no-op that PR #3136 addresses: `positionY` is still `"20"` after a request that asked for 60.

n/a (API)

### Version [e.g., 2.13.0]

2.16.0 (tested; tag `v2.16.0` = `3cf2963cd03d8b24770b7490bdb20e596baa5d65`), re-checked in source against 2.17.0 (`75330166cc00b29c14399bc2e391e4b4d8080c00`), where the culprit files are byte-identical.

yzxcj797:
Fixes #3286 — the issue's exact "Fix shape": an update-schema variant with no create-time defaults, create untouched.

## Root cause (as diagnosed, verified)

`ZEnvelopeFieldAndMetaSchema` gives every fieldMeta-bearing member `.optional().default(FIELD_<TYPE>_META_DEFAULT_VALUES)` — defaults introduced for **create** by #2181, correct there. The update route consumes the same schema, so an omitted `fieldMeta` arrives at the service as the **full default object** rather than `undefined`, and `update-envelope-fields.ts` writes `fieldMeta: updateData.fieldMeta` unconditionally inside `tx.field.update`. Since `type` is the union's discriminator and mandatory on update, the default is *always* materialised — the documentation's own partial-update example (`{ id, type, pageY }`) is exactly the call that wipes the field. There was no way to avoid it from the client side.

The audit log's `diffFieldChanges` even records the reset — the loss was visible in the audit trail while the API echoed the wiped state as though it were the update result.

## The fix

- **`ZEnvelopeFieldUpdateAndMetaSchema`** (new, beside the create schema in `field-meta.ts`): identical discriminated-union shape, `fieldMeta` optional with **no default**. The create schema and its defaults are byte-unchanged.
- **The update route** (`update-envelope-fields.types.ts`) consumes the new schema.
- **The service** spreads `fieldMeta` into the update data only when defined — Prisma skips undefined keys, so an omitted `fieldMeta` leaves the stored column untouched; a resent complete `fieldMeta` still replaces it (the documented full-replace behavior still exists for callers that want it).

The deprecated `/document/field/update-many` path is untouched (its plain `ZFieldAndMetaSchema` never had defaults).

## On #3136

As the issue notes, the in-flight coordinate fix #3136 forwards the zod-defaulted `fieldMeta` — merged alone, a move would start working while the wipe stayed. This change composes with it (the route schema both consume is the one changed here), and should land with or after it.

## Tests

`field-meta.update.test.ts` — create still materialises the default (create behavior pinned unchanged); update leaves an omitted `fieldMeta` `undefined`; a complete `fieldMeta` still parses; the option-list types (RADIO/CHECKBOX/DROPDOWN — the blank-option collapses) are covered; unknown types reject on both schemas. **The omitted-fieldMeta assertions fail on `main`** (the update path had no no-default schema — the route consumed the defaulted create schema).

### DIFF
--- a/packages/lib/server-only/field/update-envelope-fields.ts
+++ b/packages/lib/server-only/field/update-envelope-fields.ts
@@ -133,7 +133,11 @@ export const updateEnvelopeFields = async ({
             positionY: updateData.pageY,
             width: updateData.width,
             height: updateData.height,
-            fieldMeta: updateData.fieldMeta,
+            // undefined (omitted) leaves the stored fieldMeta untouched —
+            // the update schema no longer defaults them, and Prisma skips
+            // undefined keys in update data. Resending a complete fieldMeta
+            // still replaces it (#3286).
+            ...(updateData.fieldMeta !== undefined ? { fieldMeta: updateData.fieldMeta } : {}),
             envelopeItemId: updateData.envelopeItemId,
           },
         });
--- a/packages/lib/types/field-meta.ts
+++ b/packages/lib/types/field-meta.ts
@@ -448,3 +448,66 @@ export const ZEnvelopeFieldAndMetaSchema = z.discriminatedUnion('type', [
 ]);
 
 export type TEnvelopeFieldAndMeta = z.infer<typeof ZEnvelopeFieldAndMetaSchema>;
+
+/**
+ * The update counterpart of {@link ZEnvelopeFieldAndMetaSchema}: identical
+ * shape but `fieldMeta` is OPTIONAL WITH NO DEFAULT, so an update request
+ * that omits `fieldMeta` yields `undefined` instead of the create-time
+ * defaults — and the service leaves the stored configuration untouched.
+ *
+ * The create schema's defaults are correct for create; on update they made
+ * every partial `update-many` call that didn't resend the complete
+ * `fieldMeta` wipe the field's configuration to the type defaults behind a
+ * 200 (a TEXT field lost its label/placeholder/required/character limit; a
+ * RADIO's option list collapsed to one blank option). `type` is the union
+ * discriminator and mandatory, so the default was always materialised and
+ * the wipe was unavoidable from the client side (#3286).
+ */
+export const ZEnvelopeFieldUpdateAndMetaSchema = z.discriminatedUnion('type', [
+  z.object({
+    type: z.literal(FieldType.SIGNATURE),
+    fieldMeta: ZSignatureFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.FREE_SIGNATURE),
+    fieldMeta: z.undefined(),
+  }),
+  z.object({
+    type: z.literal(FieldType.INITIALS),
+    fieldMeta: ZInitialsFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.NAME),
+    fieldMeta: ZNameFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.EMAIL),
+    fieldMeta: ZEmailFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.DATE),
+    fieldMeta: ZDateFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.TEXT),
+    fieldMeta: ZTextFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.NUMBER),
+    fieldMeta: ZNumberFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.RADIO),
+    fieldMeta: ZRadioFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.CHECKBOX),
+    fieldMeta: ZCheckboxFieldMeta.optional(),
+  }),
+  z.object({
+    type: z.literal(FieldType.DROPDOWN),
+    fieldMeta: ZDropdownFieldMeta.optional(),
+  }),
+]);
+
+export type TEnvelopeFieldUpdateAndMeta = z.infer<typeof ZEnvelopeFieldUpdateAndMetaSchema>;
--- a/packages/lib/types/field-meta.update.test.ts
+++ b/packages/lib/types/field-meta.update.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  ZEnvelopeFieldAndMetaSchema,
+  ZEnvelopeFieldUpdateAndMetaSchema,
+} from '@documenso/lib/types/field-meta';
+
+describe('the update schema keeps an omitted fieldMeta undefined (#3286)', () => {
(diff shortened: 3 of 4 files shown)


Mythie:

https://documen.so/cla

Issue #2973, opened 2026-06-10
Assistant (Form Filler) Must Always Act Before Signers Without Requiring Workflow Ordering


scoobs-mystery-bytes:
### **Summary**
When adding an **Assistant** (PDF form filler) to a document, this role should always complete their tasks *before* any signing workflow begins — **without** requiring the document creator to define a signing order for signers. The Assistant is a pre‑signing role and should not behave like a signer or approver in the workflow sequence.

---

## 🎯 **Problem / Motivation**
Currently, Documenso treats all recipients as part of the same workflow unless an explicit order is defined. This causes issues when using an Assistant:

- The Assistant does **not** automatically receive the document first.
- Signers may receive signature requests before the Assistant completes their form‑filling tasks.
- Creators must artificially create a signing order just to force the Assistant to go first.
- This breaks the intended use case: the Assistant is *not* a signer and should not be part of the signing workflow.

We need a dedicated workflow behavior where the Assistant always completes their part first, without forcing the creator to define signer order.

---

### **Assistant Phase (Always First)**
- Assistant receives the document immediately upon creation.
- Assistant fills in fields, selects values, and completes all pre‑signing inputs.
- No signers or approvers receive anything during this phase.

### **Transition**
- Once the Assistant completes their task, the workflow automatically advances.

### **Signing Phase**
- Signers receive signature requests.
- Signers do **not** require a workflow order unless the creator explicitly defines one.
- If no order is defined, all signers receive requests simultaneously.

### **Approval Phase (If Applicable)**
- If approvers exist, workflow becomes:
  **Assistant → Approvers → Signers**

---

### **Workflow Logic**
- [ ] Assistant is treated as a **pre‑signing blocking role**.
- [ ] Assistant always receives the document **before** any signers or approvers.
- [ ] Signers do not require a workflow order to ensure the Assistant goes first.
- [ ] Workflow automatically becomes:
  - **Assistant → Signers** (no approvers)
  - **Assistant → Approvers → Signers** (with approvers)
- [ ] Signing order applies **only** to signers, never to the Assistant.

### **Notifications**
- [ ] Assistant receives the initial request immediately.
- [ ] Signers receive signature requests only after the Assistant completes their task.
- [ ] Approvers receive approval requests only after the Assistant completes their task.

### **Status Handling**
- [ ] Document shows **“Pending Assistant Input”** until the Assistant completes their work.
- [ ] After completion, status transitions to:
  - **“Pending Approval”** (if approvers exist), or  
[...]

PR #3294, opened 2026-08-21
fix(api): require sequential signing for ASSISTANT recipients

### LINKED ISSUE #3291: POST /api/v2/envelope/recipient/create-many accepts an ASSISTANT recipient on a parallel-signing envelope
### Issue Description

Two published pages state that the Assistant role is only available when sequential signing is enabled, and one of them lists "Be used in parallel signing mode" under what an assistant cannot do. The v2 API does not enforce it. On an envelope whose `documentMeta.signingOrder` is `PARALLEL`, `POST /api/v2/envelope/recipient/create-many` with `role: ASSISTANT` returns 200 and persists the recipient. There is no refusal of any kind.

The result is not only a contract violation. With `signingOrder` PARALLEL, `send-document.ts` notifies every recipient at once, so signers can sign before the assistant has pre-filled anything, which is the guarantee the restriction exists to protect. And the assistant's own view is broken: `getRecipientsForAssistant` and `getFieldsForToken` filter with `signingOrder: { gte: assistant.signingOrder ?? 0 }`, which is NULL-false in SQL for the null `signingOrder` an API-created parallel recipient carries, so the assistant sees an empty recipient list and none of the fields they are supposed to pre-fill.

#### Documentation

https://docs.documenso.com/docs/concepts/recipient-roles, Assistant tab:

> Assistants can prepare the document by pre-filling fields on behalf of other signers. This role is only available when sequential signing is enabled.

> **What they cannot do:**
>
> * Sign on behalf of other recipients
> * Submit the document as complete
> * Be used in parallel signing mode

> The Assistant role requires sequential signing to be enabled. You cannot use this role when recipients sign in parallel.

https://docs.documenso.com/docs/users/documents/add-recipients:

> The Assistant role is only available when sequential signing is enabled.

> Sequential signing is required to use the Assistant role. Assistants must act before the signers whose fields they pre-fill.

`add-recipients` frames the rule as an editor toggle, but `recipient-roles` states it as a property of the role itself, so it binds the API too.

#### Cause

`createEnvelopeRecipients` runs exactly one per-recipient role assertion: https://github.com/documenso/documenso/blob/3cf2963cd03d8b24770b7490bdb20e596baa5d65/packages/lib/server-only/recipient/create-envelope-recipients.ts#L62-L97. The loop calls `assertCompatibleRecipientRole({ signatureLevel: envelope.signatureLevel, role: recipient.role })`, which keys on signature level alone and only rejects ASSISTANT on AES/QES (TSP) envelopes, returning early for SES, the default. Nothing on the path reads `envelope.documentMeta.signingOrder`, and no PARALLEL or SEQUENTIAL comparison exists anywhere in the file. The request schema accepts `z.nativeEnum(RecipientRole)` with no refinement and the route adds nothing.

The rule lives only in the two editor forms: `packages/ui/primitives/document-flow/add-signers.tsx` and `apps/remix/app/components/general/envelope-editor/envelope-editor-recipient-form.tsx` both intercept a role change to ASSISTANT while signingOrder is PARALLEL, flip the form to SEQUENTIAL and toast "You cannot add assistants when signing order is disabled". Grepping `ASSISTANT` across `packages/lib/server-only` and `packages/trpc/server` finds no signing-order check at all. The role and its client-side-only enforcement arrived together in #1588 (2025-02-01, "feat: assistant role"); the v2 route that exposes the gap is #2105 (2025-11-07, "feat: add envelopes api").

Still present at today's head, `75330166cc` (v2.17.0): https://github.com/documenso/documenso/blob/75330166cc00b29c14399bc2e391e4b4d8080c00/packages/lib/server-only/recipient/create-envelope-recipients.ts#L62-L97. The file and both doc pages are byte-identical to the tested v2.16.0.

The codebase already has the shape of the fix: `packages/lib/server-only/signature-level/` holds `assert-compatible-recipient-role.ts` and `assert-compatible-signing-order.ts`, two asserts called from this same service. A complete fix has to cover `update-envelope-recipients` and the envelope-meta path as well, since flipping an envelope back to PARALLEL while an ASSISTANT recipient exists reaches the same illegal state from the other side.

#### Suggested labels

`type: bug`

#### Related

- #2973, "Assistant (Form Filler) Must Always Act Before Signers Without Requiring Workflow Ordering" (open): argues the restriction should be dropped and the assistant made a role that always acts first. It does not report that the API accepts the combination. There are two acceptable resolutions to this issue and they point in opposite directions: enforce the documented rule server-side, or change the docs and the workflow the way #2973 asks. Either settles it; the current state, where the docs forbid it and the API allows it, does not.
- #2995 (closed unmerged, 2026-06-22): the community PR for #2973, which would have removed the UI restriction and reordered notification and turn logic. Because it never landed, the docs still forbid the combination while the API still allows it.
- #1588 (merged): origin of the role and of the UI-only enforcement.

### Steps to Reproduce

1. Create an envelope and leave its signing order at the default (`documentMeta.signingOrder` is `PARALLEL` for an envelope created this way):

   `POST /api/v2/envelope/use`, `Content-Type: multipart/form-data; boundary=tcguard`, payload part `{"envelopeId":"<template envelope id>","externalId":"tcref-assistant-cb1cebb0ee"}`
   200, `{"id":"envelope_uanynwhkkxelvdtm","recipients":[]}`

   `GET /api/v2/envelope/envelope_uanynwhkkxelvdtm` confirms `"documentMeta":{"signingOrder":"PARALLEL", ...}`.

2. Add an assistant:

   ```
   {"envelopeId":"envelope_uanynwhkkxelvdtm","data":[{"email":"tcref-asst-cb1cebb0ee@documenso.test","name":"TCRef Assistant","role":"ASSISTANT"}]}
   ```

   200, and the recipient is persisted.

Re-tested live on v2.17.0 (75330166cc, 2026-08-19): still reproduces. ASSISTANT was accepted with 200 both on the default envelope, whose documentMeta.signingOrder reads PARALLEL, and on one set to PARALLEL explicitly, and the created recipient carries signingOrder: null, the value that makes the assistant's own document view come back empty.

### Expected Behavior

The call is refused and no recipient is created, because the envelope signs in parallel. No doc page publishes a status code for this refusal; `AppErrorCode.INVALID_BODY` maps to 400 and is what the sibling `assertCompatibleRecipientRole` already throws for the AES/QES case. The point of this report is that there is no refusal at all, not that the status code is wrong.

### Current Behavior

Step 2 returns HTTP 200 with the assistant created:

```json
{"data":[{"envelopeId":"envelope_uanynwhkkxelvdtm","role":"ASSISTANT","readStatus":"NOT_OPENED","signingStatus":"NOT_SIGNED","sendStatus":"NOT_SENT","id":349,"email":"tcref-asst-cb1cebb0ee@documenso.test","name":"TCRef Assistant","token":"<token>","documentDeletedAt":null,"expired":null,"expiresAt":null,"expirationNotifiedAt":null,"signedAt":null,"authOptions":{"accessAuth":[],"actionAuth":[]},"signingOrder":null,"rejectionReason":null}]}
```

The server log records `envelope.recipient.createMany` running normally, with no warning; stderr is empty.

Provenance of the PARALLEL claim: this particular recorded scenario did not read the envelope back between step 1 and step 2, so the `PARALLEL` signing order comes from sibling runs in the same session, where every envelope minted from a template this way carries `documentMeta.signingOrder: "PARALLEL"` (30 of the 31 signing-order readings across the session are PARALLEL, the one exception being an envelope explicitly set to SEQUENTIAL). The GET in step 1 above makes that check part of the reproduction.

Note `"signingOrder": null` on the created row: that is the value the assistant's own queries then filter against with `gte`, which is why the assistant's recipient and field lists come back empty.

n/a (API)

### Version [e.g., 2.13.0]

2.16.0 (tested; tag `v2.16.0` = `3cf2963cd03d8b24770b7490bdb20e596baa5d65`), re-checked in source against 2.17.0 (`75330166cc00b29c14399bc2e391e4b4d8080c00`), where the culprit file and both doc pages are byte-identical.

yzxcj797:
## Summary

Two published pages state the Assistant role is only available with sequential signing, and one lists "Be used in parallel signing mode" under what an assistant cannot do — but the restriction lived only in the two editor forms. `POST /api/v2/envelope/recipient/create-many` with `role: ASSISTANT` on a `PARALLEL` envelope returned 200 and persisted the recipient (#3291).

Beyond the contract violation, the state breaks at runtime: with `PARALLEL`, `sendDocument` notifies every recipient at once, so signers can sign before the assistant pre-fills anything, and the assistant's own lookups (`getRecipientsForAssistant`, `getFieldsForToken`) filter `signingOrder: { gte: assistant.signingOrder ?? 0 }` — NULL-false for the null recipient signingOrder an API-created parallel recipient carries — leaving the assistant an empty recipient list and no fields.

## Implementation

- New `assertAssistantCompatibleSigningOrder`, colocated with the existing signature-level asserts and called alongside `assertCompatibleRecipientRole`. A `null`/`undefined` signing order resolves to `PARALLEL` — the persisted column default — matching `sendDocument`'s distribution-time `|| PARALLEL` fallback.
- Wired into all three entry points that can reach the illegal state:
  1. `createEnvelopeRecipients` — per-recipient check against `envelope.documentMeta.signingOrder` (now included in the query);
  2. `updateEnvelopeRecipients` — same check for role changes;
  3. `updateEnvelope` — the envelope-meta path refuses flipping an envelope that already carries an assistant to parallel, closing the state from the other side.
- Violations answer `400 INVALID_REQUEST`.

Out of scope, per the issue's own framing: #2973 argues the restriction should be dropped entirely — if that lands, this assert is the single place to remove.

## Testing

- 7 new tests: the pure assert (parallel rejection, null/undefined → parallel, sequential acceptance, other roles unaffected) plus source-contract tests pinning each of the three wiring points.
- Differential: with the source changes stashed, the suite fails on unpatched main (module missing, all contracts fail).

Fixes #3291

### DIFF
--- a/packages/lib/server-only/envelope/update-envelope.ts
+++ b/packages/lib/server-only/envelope/update-envelope.ts
@@ -4,7 +4,7 @@ import type { CreateDocumentAuditLogDataResponse } from '@documenso/lib/utils/do
 import { createDocumentAuditLogData } from '@documenso/lib/utils/document-audit-logs';
 import { prisma } from '@documenso/prisma';
 import type { DocumentMeta, DocumentVisibility, Prisma, TemplateType } from '@prisma/client';
-import { DocumentStatus, EnvelopeType, FolderType, WebhookTriggerEvents } from '@prisma/client';
+import { RecipientRole, DocumentStatus, EnvelopeType, FolderType, WebhookTriggerEvents } from '@prisma/client';
 import { isDeepEqual } from 'remeda';
 
 import { TEAM_DOCUMENT_VISIBILITY_MAP } from '../../constants/teams';
@@ -16,6 +16,7 @@ import type { EnvelopeIdOptions } from '../../utils/envelope';
 import { buildTeamWhereQuery, canAccessTeamDocument } from '../../utils/teams';
 import { recomputeNextReminderForEnvelope } from '../recipient/update-recipient-next-reminder';
 import { assertCompatibleDictateNextSigner } from '../signature-level/assert-compatible-dictate-next-signer';
+import { assertAssistantCompatibleSigningOrder } from '../signature-level/assert-assistant-compatible-signing-order';
 import { assertCompatibleSigningOrder } from '../signature-level/assert-compatible-signing-order';
 import { triggerWebhook } from '../webhooks/trigger/trigger-webhook';
 import { assertEnvelopeMutable } from './assert-envelope-mutable';
@@ -60,6 +61,7 @@ export const updateEnvelope = async ({
     where: envelopeWhereInput,
     include: {
       documentMeta: true,
+      recipients: true,
       team: {
         select: {
           organisationId: true,
@@ -86,6 +88,15 @@ export const updateEnvelope = async ({
       signatureLevel: envelope.signatureLevel,
       signingOrder: meta.signingOrder,
     });
+
+    // Flipping an envelope that already carries an assistant to parallel
+    // signing reaches the same illegal state from the other side.
+    if (envelope.recipients.some((recipient) => recipient.role === RecipientRole.ASSISTANT)) {
+      assertAssistantCompatibleSigningOrder({
+        role: RecipientRole.ASSISTANT,
+        signingOrder: meta.signingOrder,
+      });
+    }
   }
 
   if (meta.allowDictateNextSigner !== undefined) {
--- a/packages/lib/server-only/recipient/create-envelope-recipients.ts
+++ b/packages/lib/server-only/recipient/create-envelope-recipients.ts
@@ -12,6 +12,7 @@ import type { EnvelopeIdOptions } from '../../utils/envelope';
 import { mapRecipientToLegacyRecipient } from '../../utils/recipients';
 import { assertEnvelopeMutable } from '../envelope/assert-envelope-mutable';
 import { getEnvelopeWhereInput } from '../envelope/get-envelope-by-id';
+import { assertAssistantCompatibleSigningOrder } from '../signature-level/assert-assistant-compatible-signing-order';
 import { assertCompatibleRecipientRole } from '../signature-level/assert-compatible-recipient-role';
 
 export interface CreateEnvelopeRecipientsOptions {
@@ -47,6 +48,7 @@ export const createEnvelopeRecipients = async ({
     where: envelopeWhereInput,
     include: {
       recipients: true,
+      documentMeta: true,
       team: {
         select: {
           organisation: {
@@ -88,6 +90,11 @@ export const createEnvelopeRecipients = async ({
     assertCompatibleRecipientRole({
       signatureLevel: envelope.signatureLevel,
       role: recipient.role,
+    });
+
+    assertAssistantCompatibleSigningOrder({
+      role: recipient.role,
+      signingOrder: envelope.documentMeta?.signingOrder,
     });
   }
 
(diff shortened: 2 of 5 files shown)


Mythie:

https://documen.so/cla

PR #3295, opened 2026-08-21
fix(health): resolve the signing transport in cert-status the way the signer does

### LINKED ISSUE #3292: /api/health and /api/certificate-status report the signing certificate as available when it is missing
## Summary

On a self-hosted deployment that does **not** set `NEXT_PRIVATE_SIGNING_TRANSPORT`, both `/api/health` and `/api/certificate-status` report the signing certificate as available even when `cert.p12` does not exist and every seal attempt throws `ENOENT`. A monitor pointed at `/api/health` stays green through a total signing outage.

## Cause

`getCertificateStatus()` and the signer disagree on how to resolve the transport when the variable is unset.

`packages/lib/server-only/cert/cert-status.ts`:

```ts
export const getCertificateStatus = () => {
  if (env('NEXT_PRIVATE_SIGNING_TRANSPORT') !== 'local') {
    return { isAvailable: true };   // <-- unset => "not local" => certificate deemed irrelevant
  }
  ...
};
```

`packages/signing/index.ts`:

```ts
const transport = env('NEXT_PRIVATE_SIGNING_TRANSPORT') || 'local';   // <-- unset => "local"

signer = await match(transport)
  .with('local', async () => await createLocalSigner())
  ...
```

So with the variable unset — which is the documented default, and what `.env.example` ships — the signer uses the local P12 transport while the health check concludes the certificate belongs to some other transport and short-circuits to `isAvailable: true`.

`apps/remix/app/routes/api+/health.ts` then reports `checks.certificate.status: "ok"` and an overall `status: "ok"` with HTTP 200.

## Reproduction

On `documenso/documenso:v2.16.0`, with `NEXT_PRIVATE_SIGNING_TRANSPORT` unset, `NEXT_PRIVATE_SIGNING_LOCAL_FILE_PATH=/opt/documenso/cert.p12`, and no file at that path:

```
$ curl -s localhost:3000/api/health
{"status":"ok","timestamp":"...","checks":{"database":{"status":"ok"},"certificate":{"status":"ok"}}}

$ curl -s localhost:3000/api/certificate-status
{"isAvailable":true,"timestamp":"..."}

$ docker exec documenso node -e "require('fs').readFileSync('/opt/documenso/cert.p12')"
ENOENT: no such file or directory, open '/opt/documenso/cert.p12'
```

The container's own entrypoint banner correctly says the opposite in the same boot:

```
🔐 Checking certificate configuration...
⚠️  Certificate not found or not readable
💡 Tip: Documenso will still start, but document signing will be unavailable
```

Setting `NEXT_PRIVATE_SIGNING_TRANSPORT=local` explicitly makes both endpoints report the missing certificate correctly, which confirms the resolution mismatch is the whole of it.

## Suggested fix

Resolve the transport identically in both places:

```ts
export const getCertificateStatus = () => {
  const transport = env('NEXT_PRIVATE_SIGNING_TRANSPORT') || 'local';

  if (transport !== 'local') {
    return { isAvailable: true };
  }
  ...
};
```

Two further points worth considering, both of which also report `ok` today:

1. **The passphrase is never checked.** `getCertificateStatus()` only calls `fs.accessSync` + `statSync` and returns `stats.size > 0`. A P12 that exists but cannot be opened with `NEXT_PRIVATE_SIGNING_PASSPHRASE` reports available, and then fails at seal time with `Failed to parse PKCS#12 file: Integrity for the PKCS#12 data is broken!`.
2. **Certificate expiry is never checked.** An expired signing certificate reports available and breaks sealing just as completely.

A health check that constructed the signer (as `createLocalSigner()` does) rather than stat-ing the file would cover all three cases. That is currently the only reliable way to know from outside the process whether signing works — we ended up having to do exactly that in our own monitoring.

## Environment

- Documenso v2.16.0 (Docker, self-hosted)
- Also present on `main` @ `75330166cc00` (2026-08-19) — `cert-status.ts` still short-circuits on `!== 'local'` while `packages/signing/index.ts` still resolves `|| 'local'`
- Local signing transport, `NEXT_PRIVATE_SIGNING_TRANSPORT` unset

yzxcj797:
## Summary

On a deployment that does not set `NEXT_PRIVATE_SIGNING_TRANSPORT` — the documented default, and what `.env.example` ships — `/api/health` and `/api/certificate-status` reported the signing certificate as available even when `cert.p12` did not exist and every seal attempt threw `ENOENT` (#3292). A monitor pointed at `/api/health` stayed green through a total signing outage, and the container's own entrypoint banner said the opposite in the same boot.

## Cause

`getCertificateStatus()` and the signer disagree on how to resolve the transport when the variable is unset:

- `cert-status.ts`: `env('NEXT_PRIVATE_SIGNING_TRANSPORT') !== 'local'` — unset is "not local", so the certificate is deemed some other transport's problem and the check short-circuits to `{ isAvailable: true }`.
- `packages/signing/index.ts`: `env('NEXT_PRIVATE_SIGNING_TRANSPORT') || 'local'` — unset **is** local, so the local P12 signer is used and the missing file fails every seal.

## Fix

`getCertificateStatus` now applies the same `|| 'local'` resolution before deciding the certificate belongs to another transport. Unset means local, the local file is actually probed, and both endpoints report the truth. Non-local transports (e.g. `gcloud-hsm`) keep their existing behavior.

## Testing

- 6 new tests: the reported case (unset + missing file → unavailable), unset + present file → available, explicit `local` + missing → unavailable, `gcloud-hsm` → skips the file check, inline `FILE_CONTENTS` → available, and a pin that the health check's default matches the signer's default.
- Differential: on unpatched main the reported case fails (`isAvailable: true`).

Fixes #3292

### DIFF
--- a/packages/lib/server-only/cert/cert-status.test.ts
+++ b/packages/lib/server-only/cert/cert-status.test.ts
@@ -1,0 +1,71 @@
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { describe, expect, it, vi } from 'vitest';
+
+const here = dirname(fileURLToPath(import.meta.url));
+
+const envState: Record<string, string | undefined> = {};
+
+vi.mock('@documenso/lib/utils/env', () => ({
+  env: (variable: string) => envState[variable],
+}));
+
+const { getCertificateStatus } = await import('./cert-status');
+
+describe('getCertificateStatus', () => {
+  it('reports the local certificate as unavailable when the transport is unset and the file is missing', () => {
+    envState.NEXT_PRIVATE_SIGNING_TRANSPORT = undefined;
+    envState.NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS = undefined;
+    envState.NEXT_PRIVATE_SIGNING_LOCAL_FILE_PATH = join(os.tmpdir(), 'documenso-cert-missing-test.p12');
+
+    expect(getCertificateStatus()).toEqual({ isAvailable: false });
+  });
+
+  it('reports the local certificate as available when the transport is unset and the file exists', () => {
+    const certPath = join(os.tmpdir(), `documenso-cert-present-${process.pid}.p12`);
+
+    try {
+      fs.writeFileSync(certPath, 'not-a-real-p12-but-non-empty');
+
+      envState.NEXT_PRIVATE_SIGNING_TRANSPORT = undefined;
+      envState.NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS = undefined;
+      envState.NEXT_PRIVATE_SIGNING_LOCAL_FILE_PATH = certPath;
+
+      expect(getCertificateStatus()).toEqual({ isAvailable: true });
+    } finally {
+      fs.rmSync(certPath, { force: true });
+    }
+  });
+
+  it('checks the local file when the transport is explicitly local', () => {
+    envState.NEXT_PRIVATE_SIGNING_TRANSPORT = 'local';
+    envState.NEXT_PRIVATE_SIGNING_LOCAL_FILE_CONTENTS = undefined;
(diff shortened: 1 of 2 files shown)


Mythie:

https://documen.so/cla

PR #3297, opened 2026-08-21
fix(api): reject forbidden field meta configurations on update

### LINKED ISSUE #3296: POST /api/v2/envelope/field/update-many accepts a field that is both required and read-only
### Issue Description

#3290 fixed the create path; the update path has the same hole. `updateEnvelopeFields` (`packages/lib/server-only/field/update-envelope-fields.ts`) validates recipients, item ownership, and the field-meta/type match, but never inspects the `fieldMeta` it persists — and the update replaces the meta wholesale (`fieldMeta: updateData.fieldMeta`), so:

- `POST /api/v2/envelope/field/update-many` with `fieldMeta: {"type": "text", "required": true, "readOnly": true}` returns 200 and persists the forbidden combination;
- a field created legal (`required: true`) can be flipped illegal in one update that supplies `readOnly: true` in its replacement meta.

The documented rules are the same five Field Types pages quoted in #3290, and the runtime consequences are identical: the signer-facing validation error the sender never sees.

### Cause

The per-field loop runs exactly four checks (field exists, recipient exists, `canRecipientFieldsBeModified`, meta/type match) and writes `updateData.fieldMeta` straight to `tx.field.update`. No call to the configuration validators, which after #3293 exist as a reusable `validateFieldMetaConfiguration`.

### Suggested fix

Run `validateFieldMetaConfiguration(fieldType, field.fieldMeta)` in the same loop — the effective type is already computed (`field.type || originalField.type`), and only the *provided* meta needs validating, because the update replaces it wholesale (an omitted meta keeps the stored one, per #3286/#3289). Reject with `AppError(AppErrorCode.INVALID_REQUEST)` so the API answers 400, mirroring #3290's create-path fix.

#### Suggested labels

`type: bug`

Found while reviewing #3293's coverage; the same audit reasoning as #3290 applies.

yzxcj797:
## Summary

Companion to the create-path fix #3293, covering the other direction identified in #3296: `updateEnvelopeFields` replaced `fieldMeta` wholesale without ever inspecting it, so `POST /api/v2/envelope/field/update-many` could persist `{"required": true, "readOnly": true}` — or flip a legal field illegal in one replacement meta — and the signer-facing failure only surfaced at signing time.

## Implementation

The per-field loop now runs the same `validateFieldMetaConfiguration` #3293 introduced, against the effective field type (`field.type || originalField.type`, already computed for the meta/type match directly above), answering `400 INVALID_REQUEST` with the violation list. An omitted `fieldMeta` keeps the stored one untouched (#3286/#3289 semantics) and skips the check — only meta the update *writes* is validated, because the write replaces wholesale.

**Stacked on #3293** (the shared validator lands there) — merge order: #3293 first.

## Testing

- Colocated source-contract tests pin the import, the validation call against the effective type, and the 400 shape. Differential: all 3 fail with the source change stashed.
- `packages/lib/server-only/field/` suite green (6/6).

Fixes #3296

### DIFF
--- a/packages/lib/advanced-fields-validation/validate-field-meta-configuration.test.ts
+++ b/packages/lib/advanced-fields-validation/validate-field-meta-configuration.test.ts
@@ -1,0 +1,69 @@
+import { FieldType } from '@prisma/client';
+import { describe, expect, it } from 'vitest';
+
+import { validateFieldMetaConfiguration } from './validate-field-meta-configuration';
+
+describe('validateFieldMetaConfiguration', () => {
+  it('rejects a field that is both required and read-only', () => {
+    const errors = validateFieldMetaConfiguration(FieldType.TEXT, {
+      required: true,
+      readOnly: true,
+      text: 'pre-filled',
+    });
+
+    expect(errors).toContain('A field cannot be both read-only and required');
+  });
+
+  it('rejects the combination for every documented field type', () => {
+    for (const type of [FieldType.TEXT, FieldType.NUMBER, FieldType.RADIO, FieldType.CHECKBOX, FieldType.DROPDOWN]) {
+      expect(
+        validateFieldMetaConfiguration(type, { required: true, readOnly: true }),
+      ).toContain('A field cannot be both read-only and required');
+    }
+  });
+
+  it('accepts required-only and read-only-with-default configurations', () => {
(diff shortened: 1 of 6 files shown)


Mythie:

https://documen.so/cla

PR #3298, opened 2026-08-21
fix(api): answer setFields validation failures with 400, not 500

yzxcj797:
## Summary

#3290 asked that field-meta violations answer `400 INVALID_REQUEST` "rather than the bare `new Error(errors.join(', '))` the `setFields` path throws today, which surfaces as a 500 even where the rules are enforced." #3293 fixed the create path; this converts the two editor `setFields` paths that already *enforce* the rules, so the same invalid configuration answers 400 everywhere it can be rejected.

## Implementation

Every validator failure and missing-metadata throw in `set-fields-for-document.ts` and `set-fields-for-template.ts` (17 sites) becomes `AppError(AppErrorCode.INVALID_REQUEST)` with the same message — the editor's toast text is unchanged, only the HTTP status and error code become correct. The one internal upsert invariant (`'Not possible'`) stays a plain `Error`: it is unreachable by user input and indicates a programming fault, where a 500 is the honest answer.

## Testing

- Colocated source-contract tests pin: no bare `errors.join`/user-message throws remain, `AppError(INVALID_REQUEST)` count per file, and the internal invariant is untouched.
- Differential: 2 of 3 fail with the source changes stashed; the field-service suite is green (9/9 including #3293's tests).

Companion to #3293 (no merge-order dependency).

### DIFF
--- a/packages/lib/server-only/field/set-fields-for-document.ts
+++ b/packages/lib/server-only/field/set-fields-for-document.ts
@@ -139,7 +139,7 @@ export const setFieldsForDocument = async ({
           const errors = validateTextField(textFieldParsedMeta.text || '', textFieldParsedMeta);
 
           if (errors.length > 0) {
-            throw new Error(errors.join(', '));
+            throw new AppError(AppErrorCode.INVALID_REQUEST, { message: errors.join(', ') });
           }
         }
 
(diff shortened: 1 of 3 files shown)


Mythie:

https://documen.so/cla

PR #3309, opened 2026-08-26
fix: default unset signing transport to local

### LINKED ISSUE #3292: /api/health and /api/certificate-status report the signing certificate as available when it is missing

catalinpit:
`/api/health` and `/api/certificate-status` reported the cert as available when `NEXT_PRIVATE_SIGNING_TRANSPORT` was unset, even though sealing defaults to the local P12 and fails if it is missing, unreadable, or expired.

### DIFF
--- a/apps/remix/app/routes/api+/certificate-status.ts
+++ b/apps/remix/app/routes/api+/certificate-status.ts
@@ -1,8 +1,8 @@
 import { getCertificateStatus } from '@documenso/lib/server-only/cert/cert-status';
 
-export const loader = () => {
+export const loader = async () => {
   try {
-    const certStatus = getCertificateStatus();
+    const certStatus = await getCertificateStatus();
 
     return Response.json({
       isAvailable: certStatus.isAvailable,
--- a/apps/remix/app/routes/api+/health.ts
+++ b/apps/remix/app/routes/api+/health.ts
@@ -22,7 +22,7 @@ export const loader = async () => {
   }
 
   try {
-    const certStatus = getCertificateStatus();
+    const certStatus = await getCertificateStatus();
 
     if (certStatus.isAvailable) {
       checks.certificate = { status: 'ok' };
--- a/packages/lib/constants/app.ts
+++ b/packages/lib/constants/app.ts
@@ -93,6 +93,8 @@ export const NEXT_PRIVATE_USE_PLAYWRIGHT_PDF = () => env('NEXT_PRIVATE_USE_PLAYW
 
 export const NEXT_PRIVATE_SIGNING_TIMESTAMP_AUTHORITY = () => env('NEXT_PRIVATE_SIGNING_TIMESTAMP_AUTHORITY');
 
+export const NEXT_PRIVATE_SIGNING_TRANSPORT = () => env('NEXT_PRIVATE_SIGNING_TRANSPORT') || 'local';
+
 /**
  * Whether this Documenso instance is running in CSC (Cloud Signature Consortium) mode.
  *
--- a/packages/lib/server-only/cert/cert-status.ts
+++ b/packages/lib/server-only/cert/cert-status.ts
@@ -1,26 +1,35 @@
-import * as fs from 'node:fs';
+import { X509Certificate } from 'node:crypto';
 
-import { env } from '@documenso/lib/utils/env';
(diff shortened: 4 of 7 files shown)


Mythie:
prefer `NEXT_PRIVATE_SIGNING_TRANSPORT()` over `getSigningTransport()` and it can sit under a `constants.ts` like others where we do the same kind of normalization.

You can still sign with a cert past its expiration but the changes to handle erroring on that are nice.

Mythie (changes requested):
per above comment

Mythie (approved):

PR #3353, opened 2026-09-09
fix: add terminal-status idempotency guard to the seal handler

### LINKED ISSUE #3187: Concurrent completion can trigger sealing twice, re-signing the PDF over itself
### Issue Description

The decision to seal is a check-then-act with no lock, and the sealing handler has no status guard, so two recipients completing at nearly the same time can both trigger sealing.

**Trigger — `packages/lib/server-only/document/complete-document-with-token.ts`**
Each recipient commits their own transaction, then *separately* queries whether everyone has signed. In `PARALLEL` mode, two recipients finishing a few hundred milliseconds apart both see a complete set and both call `triggerJob('internal.seal-document')`.

Nothing deduplicates: `triggerJob` creates a new `BackgroundJob` row per call, so each gets its own task key, and the job definition sets `optimizeParallelism: true`.

**Handler — `packages/lib/jobs/definitions/internal/seal-document.handler.ts`**
There is no check on `envelope.status` on entry. `isDocumentCompleted` appears only at line 345, inside the `isResealing` branch. The remaining conditions cannot stop a replay: `envelopeItems.length`, `isComplete` (still true), and `fieldsContainUnsignedRequiredField` (still false, since `inserted` stays true).

**Why the failure converges instead of stopping.** The losing job hits P2025 on the `documentDataId` unique constraint, is retried immediately with no backoff, refetches an envelope that is now `COMPLETED` with the already-sealed PDF, and — with no status guard — **seals on top of it**. This time `oldDocumentDataId` matches, so the transaction succeeds.

Result: a second certificate page, a second audit-log page, a signature applied over an existing signature (Acrobat then reports the document as modified since signing), `completedAt` overwritten, a second `DOCUMENT_COMPLETED` webhook, and a second "Signing Complete" email with the signed contract attached to everyone.

### Steps to Reproduce

Two signers on a `PARALLEL` document, both completing within the same window — realistically two people signing side by side, or an API integration completing recipients concurrently.

### Expected Behavior

Sealing happens exactly once per envelope. A duplicate trigger is a no-op.

### Current Behavior

The document can be sealed twice, producing a PDF whose first signature is invalidated, duplicated certificate/audit pages, and duplicate notifications.

### Version

`main` @ 617f8cc (also reproduces against 2.16.0)

### Notes

Found during a code audit. **This one is a concurrency analysis, not an observed failure** — I traced the window (commit → `findFirst` → `triggerWebhook` → `triggerJob` → `findMany`, roughly 50–400 ms) and confirmed the absence of both a status guard and a dedup key on current `main`, but I have not reproduced it against a running instance. Treat the reasoning as the claim, not the outcome.

Two directions, either of which closes it:
- an idempotency guard at the top of the handler — `if (isDocumentCompleted(envelope.status) && !isResealing) return …`;
- an atomic transition at the trigger — `updateMany({ where: { id, status: PENDING }, … })` and only schedule the job when `count === 1`.

breken-ai:
## Root cause

Concurrent completion can trigger sealing twice (check-then-act per recipient, no lock), and the seal handler has **no status guard at entry** - the only `isDocumentCompleted` check is at the bottom, gating the reseal email. So the losing job `P2025`s on the envelopeItem update after the winner swaps `documentDataId`; the framework retry then refetches a COMPLETED envelope holding the already-sealed PDF and seals over it - second certificate page, second audit page, signature over signature, duplicate webhook and completion emails.

## Fix

An idempotency guard at handler entry: a cheap status-only fetch in its own `runTask`, and if the envelope is already COMPLETED or REJECTED (and this isn't an explicit admin reseal, which sets `isResealing`), the handler returns before any writes. A duplicate trigger or a retried loser now no-ops instead of resealing. Verified the other trigger paths: the sweep only selects PENDING envelopes and the admin reseal flow sets `isResealing`, so neither is affected.

## Verification

New vitest file covers the four cases: duplicate COMPLETED no-ops, duplicate REJECTED no-ops, PENDING proceeds, admin reseal proceeds. On current main the first two fail; with the guard all four pass, and the full `packages/lib` suite stays green (216 passed).

One residual worth knowing: two jobs that both pass the guard in the same instant still duplicate the PDF work (the loser's write still fails safe via the P2025, and its retry now no-ops). A conditional status update inside the final transaction would close that window too if you want it.

Fixes #3187

> Built by breken, your AI support engineer - breken.ai - this one's on us.

### DIFF
--- a/packages/lib/jobs/definitions/internal/seal-document.handler.test.ts
+++ b/packages/lib/jobs/definitions/internal/seal-document.handler.test.ts
@@ -1,0 +1,85 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockEnvelopeFindFirst, mockEnvelopeFindFirstOrThrow } = vi.hoisted(() => ({
+  mockEnvelopeFindFirst: vi.fn(),
+  mockEnvelopeFindFirstOrThrow: vi.fn(),
+}));
+
+vi.mock('@documenso/prisma', () => ({
+  prisma: {
+    envelope: {
+      findFirst: mockEnvelopeFindFirst,
+      findFirstOrThrow: mockEnvelopeFindFirstOrThrow,
+    },
+  },
+}));
+
+vi.mock('@documenso/ee/server-only/signing/csc/finalize-tsp-completion', () => ({
+  finalizeTspEnvelopeCompletion: vi.fn(),
+}));
+vi.mock('@documenso/lib/server-only/pdf/add-rejection-stamp-to-pdf', () => ({ addRejectionStampToPdf: vi.fn() }));
+vi.mock('@documenso/lib/server-only/pdf/generate-audit-log-pdf', () => ({ generateAuditLogPdf: vi.fn() }));
+vi.mock('@documenso/lib/server-only/pdf/generate-certificate-pdf', () => ({ generateCertificatePdf: vi.fn() }));
+vi.mock('@documenso/lib/server-only/pdf/get-page-size', () => ({ getLastPageDimensions: vi.fn() }));
+vi.mock('@documenso/signing', () => ({ signPdf: vi.fn() }));
+vi.mock('../../../constants/app', () => ({ NEXT_PRIVATE_USE_PLAYWRIGHT_PDF: false }));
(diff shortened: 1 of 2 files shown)


Mythie:

https://documen.so/cla

